PR #13, opened 2016-06-09
Fixed build on Linux, removed implementation code from header files, fixed compiler warnings.

jl2:
The project now builds for me on Linux.  It *should* still work on OSX, and appears to, but my libffi is too old and it gets a link error for me.  With an up to date libffi I think it will work.

I also split out declarations and definitions from the files in shared/ because they seemed to be causing some link errors for me.


eriksvedang:
Very cool, thank you! 
I'll merge and check if I can still build on OSX.

eriksvedang:
Cmake works fine after running 
```export PKG_CONFIG_PATH=/usr/local/Cellar/libffi/3.0.13/lib/pkgconfig/```
But when i do ```make``` i get an include error for <ffi.h>, is there something missing in the cmake-file or am I doing something wrong?

**Comment by eriksvedang** [comment]
Setting the C_INCLUDE_PATH solved this error but now I get this instead:
```
Undefined symbols for architecture x86_64:
  "_ffi_closure_alloc", referenced from:
      _call_foreign_function in call_ffi.c.o
  "_ffi_prep_closure_loc", referenced from:
      _call_foreign_function in call_ffi.c.o
```

jl2:
Try running "make VERBOSE=1", and see if the homebrew version of libffi is being linked in.  My understanding is the default OSX libffi doesn't include the ffi_closure_alloc functions. 

I'm not very familiar with pkg-config and cmake, so it's entirely possible I used the wrong link variables in the cmake file.

I can look into it more after work this evening.

eriksvedang:
After rebuilding libffi it worked. I noticed that your changes of shared.h produce multiple versions of a lot of functions though, so still can't build. 

Also, have you tried running the tests? (set CARP_DEV=1 before running) The point of having the code in shared.h is to enable the user program to access those functions when compiling. So perhaps you could fix that thing and commit again, or I'll have to revert the patch for now.

eriksvedang:
Could you make a new pull request with the same things, but no changes to the '/shared' directory? Or was there anything in there that needed fixing to make it compile on Linux?

PR #19, opened 2016-06-17
added some fixes so that carp compiles for linux and spin runs

paines:
The lastest version would fail to compile on Linux, but with a few tweaks, as you can see in the pull request, it runs now. We should then add / mention in the Installation manual, that for Linux clang must be installed. This isn't the default compiler as it is for OSX.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,11 @@ set(LIBFFI_LIBRARY_PATH
   PATH
   "libffi library directory")
 
+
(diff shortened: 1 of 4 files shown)


eriksvedang:
Very nice, thank you! I'll merge it soon.

eriksvedang:
Patch worked well, thanks again. I added a note about Clang on Linux and also added you to the list of contributors. Cheers!

PR #22, opened 2016-07-04
Attempting to create clang-format for Carp

mkfifo:
I was wondering what your thoughts are on this clang-format config, I tried to match it as close to your existing style as possible.

This will remove tabs and trailing spaces, which I noticed littered around.

I saw both

```
if(cond) {
} else {
}
```

and

```
if(cond) {
}
else {
}
```

the latter seemed more common so I went with that.

If you are unhappy with this I am happy to change it or revert it, completely up to you.


mkfifo:
Updated to also run via cmake.

mkfifo:
Closing PR to break out, will return shortly.

PR #23, opened 2016-07-04
Attempting to create clang-format for Carp

mkfifo:
I was wondering what your thoughts are on this clang-format config, I tried to match it as close to your existing style as possible.

This will remove tabs and trailing spaces, which I noticed littered around.

I saw both

```
if(cond) {
} else {
}
```

and

```
if(cond) {
}
else {
}
```

the latter seemed more common so I went with that.

If you are unhappy with this I am happy to change it or revert it, completely up to you.

Updated both Makefile and CMakeLists, so both work:

```
make format
```

and

```
cmake .
make format
```


eriksvedang:
Nice! There seems to be one little bug in the command as of now, in your patch it has turned line 13 in shared.h into this:
```#define and&&```
(there should be a space between 'and' and '&&')

Please fix this before I apply the patch.

**Comment by mkfifo** [comment]
Hmmm that is interesting as `and` is a keyword in c++, which may be part of this issue.
I also would recommend against using `and` in this way, especially if you want to export c++ bindings.

I also cannot find any uses of that define

```
git grep - ni 'and' src/ shared/
```

Doesn't seem to show anything of interest.

Interestingly, clang-format doesn't seem to touch most other uses of && outside of define, it leaves alone ```int main(void){ int c = 1 && 2; }```

I will file a bug with clang-format / libformat and get some opinions as I think that handling is a bug, however I also don't think we *should* be using `#define and &&`.

**Comment by mkfifo** [comment]
The other option - which I initially considered - is to leave the files in shared/ untouched.

**Comment by eriksvedang** [comment]
Ah, the 'and' is used by the compiled Carp programs, so that they can emit an "and" which uses the native && operator. I think it would be good to run the script over shared/ also, but maybe leave it out for now so that we can apply the patch...

**Comment by mkfifo** [comment]
clang-format also supports

```
// clang-format off
#define and &&
// clang-format on
```

which I think is a little ugly, but it would allow us to keep this - until I hear back from the clang-format team about the issue.

I will make the change now and you can let me know what you think.

Don't feel pressured to accept this, if you don't think it is best for Carp I am happy to scrap this :)

**Comment by mkfifo** [comment]
Rebased ontop of master, added above clang-format off

see

https://github.com/eriksvedang/Carp/pull/23/files#diff-5e84b84f57bb693656fbaa59cd0298dfR13

**Comment by eriksvedang** [comment]
The ```clang-format-off``` seems like a good workaround to me, let's do that!

I think it's great to have this script so no desire to scrap it, for sure. Thanks for putting it together.

I have one final wish though, regarding casts... To me the extra space (i.e. ```(int *)``` instead of ```(int*)```) looks annoying, is there a way to change that format rule? Or is that such a common way of writing it that I should just learn to appreciate it? :) If you convince me of that I'll merge, otherwise a little fix would be greatly appreciated.

**Comment by mkfifo** [comment]
I prefer right-side `*` for actual pointers declarations, like in say `char *ch`.

I do find it annoying how it is changing casts to also be right-side, although the difference is minimal I do find it a bit ugly in some cases.

So far I haven't figured out how to make it change pointer declarations without also changing casts - I will try have another go.

**Comment by mkfifo** [comment]
Reading over the docs, all mentions of 'pointer' affect both.
and the options for 'cast' don't deal with the space in the middle (only spaces around or after).
I tried them all anyway, to no avail.

Tried:

```
SpaceAfterCStyleCast: true                                                      
SpacesInCStyleCastParentheses: false                                            
DerivePointerAlignment: false                                                   
PointerAlignment: Right        
```

with varying settings (true/fale and Left/Middle/Right).

So it looks like we are stuck with this sorry.

eriksvedang:
OK, no worries! I'll merge it.

mkfifo:
Thanks!

Longer-term I can try submit a patch to clang-format, or maybe even write our own script that runs after clang-format to style things just the way we like.

PR #24, opened 2016-07-04
updating cmake libffi to use pkg-config - now builds under linux cleanly

mkfifo:
The previous cmake included hard-coded paths to libffi which looked like paths on a mac.

This pkg-config version should work on any system as long as 'pkg-config' is installed and working correctly.

What are your thoughts?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,25 @@ set(VERSION "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}")
 
 if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_GNU_SOURCE")
+
+    # linux uses pkg config
+    Include(FindPkgConfig)
+    PKG_SEARCH_MODULE(LIBFFI REQUIRED libffi)
+else()
+    # other systems use manually set paths
+    # set name must match the one set by pkg config
+    set(LIBFFI_INCLUDE_DIRS
+      "${PROJECT_SOURCE_DIR}/../libffi/x86_64-apple-darwin15.3.0/include"
+      CACHE
+      PATH
+      "libffi include directory")
+
+    # set name must match the one set by pkg config
+     set(LIBFFI_LIBRARIES
+      "${PROJECT_SOURCE_DIR}/../libffi/x86_64-apple-darwin15.3.0/.libs"
+      CACHE
+      PATH
+      "libffi library directory")
 endif()
 
 set(SOURCE_DIR src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,32 +7,30 @@ set(VERSION_MINOR "0")
 set(VERSION_PATCH "1")
 set(VERSION "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}")
 
-set(LIBFFI_INCLUDE_PATH
-  "${PROJECT_SOURCE_DIR}/../libffi/x86_64-apple-darwin15.3.0/include"
-  CACHE
-  PATH
-  "libffi include directory")
-
-set(LIBFFI_LIBRARY_PATH
-  "${PROJECT_SOURCE_DIR}/../libffi/x86_64-apple-darwin15.3.0/.libs"
-  CACHE
-  PATH
-  "libffi library directory")
-
-
 if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
-	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_GNU_SOURCE")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_GNU_SOURCE")
 endif()
 
 set(SOURCE_DIR src)
 include(globfiles.cmake)
 
-include_directories(${LIBFFI_INCLUDE_PATH})
-link_directories(${LIBFFI_LIBRARY_PATH})
+include(FindPkgConfig)
+PKG_SEARCH_MODULE(LIBFFI REQUIRED libffi)
+
+include_directories(
+    ${LIBFFI_INCLUDE_DIRS})
+
+link_directories(
+    ${LIBFFI_LIBRARIES})
 
 add_executable(${PROJECT_NAME} ${${PROJECT_NAME}_h} ${${PROJECT_NAME}_c})
 
-target_link_libraries(${PROJECT_NAME} ffi m c pthread dl)
+target_link_libraries(${PROJECT_NAME}
+    ${LIBFFI_LIBRARIES}
+    m
+    c
+    pthread
+    dl)
 
 set_target_properties(${PROJECT_NAME}
     PROPERTIES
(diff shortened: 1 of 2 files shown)


eriksvedang:
I'm a bit worried that this messes with the Windows build which uses Cmake (and a manually entered path). Maybe it's possible to use pkg-config only on Linux?

mkfifo:
Sure thing, I should be able to tweak it to only be used under Linux.

Do we also want to use it on mac, or would you prefer hardcoded mac paths as well as hardcoded windows paths?

Where are the current hardcoded windows paths?

eriksvedang:
The person compiling on Windows would have to enter the paths to his/her local copy of libffi. We have discussed just shipping the dll with the project to not have to set all of that up, perhaps that would be the most usable. So there is no hardcoded path in the project right now. Maybe pkg-config would be the right thing for Mac, I'm not sure how widely used it is by Mac people. Something that "just works" with Homebrew would probably help most people...

daveyarwood:
My 2¢ as a Mac/Homebrew user: I had to do some finagling\* to get it so that my `ffi.h` and `ffitarget.h` could be found, so this is already not working out-of-the-box. It makes me wonder if pkg-config might help for OS X users, too? Not everyone is as willing to finagle as I am :)

\*A few places ([example](https://pyspotify.mopidy.com/en/latest/installation/#id4)) suggested running `xcode-select --install`, but that didn't work for me. What ended up working was manually copying the above 2 files into `/usr/include`.

eriksvedang:
Yeah, good points. I'll try to sort this out in a nice way ASAP.

mkfifo:
It would be great to get to the point of "install these deps, run cmake, run make" and get to a working carp binary.

I think pkg-config is probably the best bet for getting this experience working.

Is anyone here familiar with the state of pkg-config under mac?
I don't have a mac to test, would be good to test my changes.

eriksvedang:
Yes, that would be great indeed! I have pkg-config on Mac and it worked fairly well, I'll try out your patch again and try to write up the instructions for how to make it work. I think the trick is that you need to set the PKG_CONFIG_PATH which is not super clear if you're not used to compiling C projects.

mkfifo:
I would also be happy for us to write a small script that ran to find out the appropriate PKG_CONFIG_PATH on a mac - if need be

Although I lack a mac for actual testing.

eriksvedang:
Interesting idea! How would it find that out?

mkfifo:
does homebrew track where it has installed the contents of each package? if so, can we consult homebrew for an arbitrary package? e.g. "homebrew, please tell me all the files you created when you installed <dep x>'

If so, we can then make sure the `.pc` files for each dep are in the PKG_CONFIG_PATH

eriksvedang:
Yes, this seems like a good way to do it. I'll read up on Homebrew.

mkfifo:
Awesome, I'm very interested in lowering the install barrier for Carp.

mkfifo:
I have rebased ontop of current master and added second patch to make pkgconfig usage only happen on linux.

I also changed the mac hardcoded paths to have the same names as those used by pkgconfig, to make the include lines later on simpler.

Please test on mac.

daveyarwood:
@mkfifo Just gave this a try on OS X 10.9.5 with `pkg-config` previously installed via Homebrew. The `cmake .` step failed because it couldn't find `libffi`. Here's the output:

```
$ brew install pkg-config
Warning: pkg-config-0.29.1 already installed

$ brew info pkg-config
pkg-config: stable 0.29.1 (bottled)
Manage compile and link flags for libraries
https://freedesktop.org/wiki/Software/pkg-config/
/usr/local/Cellar/pkg-config/0.29.1 (10 files, 631.7K) *
  Poured from bottle on 2016-05-11 at 20:37:43
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pkg-config.rb

$ cmake .
-- The C compiler identification is AppleClang 6.0.0.6000057
-- The CXX compiler identification is AppleClang 6.0.0.6000057
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PkgConfig: /usr/local/bin/pkg-config (found version "0.29.1")
-- Checking for one of the modules 'libffi'
CMake Error at /usr/local/Cellar/cmake/3.5.2/share/cmake/Modules/FindPkgConfig.cmake:574 (message):
  None of the required 'libffi' found
Call Stack (most recent call first):
  CMakeLists.txt:37 (PKG_SEARCH_MODULE)
```

I get the same result even if I run `cmake . -DLIBFFI_INCLUDE_PATH=/usr/local/opt/libffi/lib/libffi-3.0.13/include/ -DLIBFFI_LIBRARY_PATH=/usr/local/opt/libffi/lib/
make` which worked for me on origin/master.

mkfifo:
@daveyarwood thank you very much for trying this.

Which patch did you try?

I have the first patch on here which should use pkg-config on all platforms.
I then have a second patch that will only use pkg-config under linux, although you would have to use `cmake . -DLIBFFI_INCLUDE_DIRS=/usr/local/opt/libffi/lib/libffi-3.0.13/include/ -DLIBFFI_LIBRARIES=/usr/local/opt/libffi/lib/ make` to match the pkg-config syntax - I will add a follow up to allow both syntaxes.

daveyarwood:
What I tried yesterday was the second patch.

Just now I tried both `cmake .` and `cmake . -DLIBFFI_INCLUDE_DIRS... (etc.)` on origin/master, and got the same error about not finding libffi.

eriksvedang:
My best suggestion right now is to build libffi from source and install it on the system level. We're working on a better solution, sorry for the trouble.

mkfifo:
@daveyarwood was Carp building for your correctly before this patch?

daveyarwood:
@mkfifo Yes, it was working when I did `cmake . -DLIBFFI_INCLUDE_PATH=/usr/local/opt/libffi/lib/libffi-3.0.13/include/ -DLIBFFI_LIBRARY_PATH=/usr/local/opt/libffi/lib/`

Now it doesn't work anymore, even if I do the above or the above with `DIRS` instead of `PATH`.

mkfifo:
@daveyarwood thanks for getting back to me.

can you please try

```
cmake . -DLIBFFI_INCLUDE_DIRS=/usr/local/opt/libffi/lib/libffi-3.0.13/include/ -DLIBFFI_LIBRARIES=/usr/local/opt/libffi/lib/
```

so INCLUDE_PATH changes to INCLUDE_DIRS and LIBRARY_PATH changes to LIBRARIES.

daveyarwood:
Just tried that... same result. I did double-check that those directories are where my libffi headers, etc. are kept, and they are.

mkfifo:
@daveyarwood thanks for reporting this.

Currently master builds cleanly for @eriksvedang 

I'm trying to setup travis-ci to build osx so we can prevent breakages like this agian.

@daveyarwood can you please detail how you installed the deps and the full cmake line you are running? - I lack a mac so I want them to replicate them on travis-ci.

mkfifo:
@daveyarwood moving discussion to #54 as this is an outstanding issue and I don't want it being lost.

PR #28, opened 2016-07-06
macports: avoid dylib/link confusion

rurban:
use true bundle/dylib extensions, avoid mixup, avoid hardcoded .so,
add -L/opt/local/lib -I/opt/local/include search paths for macports,
use libglfw.dylib, not libglfw3.dylib

not necesserily to be merged, but this is what I needed

### DIFF
--- a/bin/carp
+++ b/bin/carp
@@ -7,4 +7,4 @@ WRAPPER=rlwrap
 #WRAPPER=lldb
 #WRAPPER=valgrind
 
-CARP_DIR=$DIR/../ $WRAPPER $DIR/carp-repl $1
+CARP_DIR=$DIR/../ $WRAPPER $DIR/carp-repl $@
--- a/lisp/gl.carp
+++ b/lisp/gl.carp
@@ -1,6 +1,7 @@
 
-(if (osx?) (def *LIBNAME* "libglfw3.dylib") "")
-(if (linux?) (def *LIBNAME* "libglfw.so") "")
+;(if (osx?) (def *LIBNAME* "libglfw3.dylib") "")
+;(if (linux?) (def *LIBNAME* "libglfw.so") "")
+(def *LIBNAME* (str (if (windows?) "" "lib") "glfw" (:dylib-extension platform-specifics))
 
 (def glfw (load-dylib *LIBNAME*))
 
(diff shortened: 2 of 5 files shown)


eriksvedang:
Nice, making it work well with macports would be great. I get a lot of errors when running the tests after this patch though. I think it has to do with the .bundle suffix. Run the tests with ```CARP_DEV=1 carp```.

daveyarwood:
@rurban Your instructions in `INSTALL.md` pointed me in the right direction to build Carp successfully! 🎉 

I use Homebrew, not Macports, so my libffi is in a slightly different directory. What I ran was:

```
git clone https://github.com/eriksvedang/Carp.git
cd Carp
cmake . -DLIBFFI_INCLUDE_PATH=/usr/local/opt/libffi/lib/libffi-3.0.13/include/ -DLIBFFI_LIBRARY_PATH=/usr/local/opt/libffi/lib/
make
```

lisp/gl.carp
```diff
@@ -1,6 +1,7 @@
 
-(if (osx?) (def *LIBNAME* "libglfw3.dylib") "")
-(if (linux?) (def *LIBNAME* "libglfw.so") "")
+;(if (osx?) (def *LIBNAME* "libglfw3.dylib") "")
+;(if (linux?) (def *LIBNAME* "libglfw.so") "")
+(def *LIBNAME* (str (if (windows?) "" "lib") "glfw" (:dylib-extension platform-specifics))
```
mkfifo: I find this a lot harder to reason about.

Where is the '3' needed for 'libglfw3.dylib' now coming from?

I would much prefer an if for each os, as previously done.

bin/carp
```diff
@@ -7,4 +7,4 @@ WRAPPER=rlwrap
 #WRAPPER=lldb
 #WRAPPER=valgrind
 
-CARP_DIR=$DIR/../ $WRAPPER $DIR/carp-repl $1
+CARP_DIR=$DIR/../ $WRAPPER $DIR/carp-repl $@
```
mkfifo: this is solid and we should merge this.

mkfifo:
since @daveyarwood was helped by install instructions I think it is worth including those too.

eriksvedang:
I (manually) added the changes suggested by @mkfifo and will now close this. Added you as a contributor to README.md, @rurban . Thanks for the help!

PR #35, opened 2016-07-07
adding blank lines in RESEARCH.md for better markdown formatting

mkfifo:
Markdown will squash adjacent lines into a single sentence/paragraph, blank lines between them will prevent this.

See the current formatting of
https://github.com/eriksvedang/Carp/blob/master/RESEARCH.md
which often has multiple links on the same line, and with word-wrapping.

and compare it to
https://github.com/mkfifo/Carp/blob/research-markdown-formatting/RESEARCH.md
which has a single link per line.

We could alternatively add in some html `<br />`, or format it as a mark-down list by prefixing each line with ` -`.
But this proposed solution has the advantage of the file still being easy to treat as plain-text.

### DIFF
--- a/RESEARCH.md
+++ b/RESEARCH.md
@@ -1,70 +1,82 @@
 https://news.ycombinator.com/item?id=12039268
 
 # C PROGRAMMING, LLVM, AND DEBUGGING
-http://clang.llvm.org/docs/AddressSanitizer.html
-https://www.iar.com/support/resources/articles/advanced-preprocessor-tips-and-tricks/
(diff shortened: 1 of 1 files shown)


eriksvedang:
Oh, I usually just read the file in my text editor so that's why I haven't cared about this. Maybe making them bullet points would work better for both situations?

mkfifo:
okay, will make that change now

mkfifo:
Done, please see https://github.com/mkfifo/Carp/blob/research-markdown-formatting/RESEARCH.md

eriksvedang:
Excellent!

mkfifo:
Thanks :)

PR #36, opened 2016-07-07
enable -std=c99 on cmake allowing for GCC linux builds

shyguy:
This fixes building Carp on linux with GCC for me, untested if this breaks other platform's builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,3 +43,4 @@ add_custom_target(
     COMMAND find . -iname '*.[ch]' | xargs clang-format -style=file -i
 )
 
+set_property(TARGET carp-repl PROPERTY C_STANDARD 99)


eriksvedang:
I'll try this on Mac before merging

mntmn:
Had to make the same change on my system (Debian Jessie w/ GCC 4.9.2-10) and almost created the same pull request.

shyguy:
Hmm, through more testing it appears that there are still problems with the build. After building by using cmake (3.2.2, built from source) on Ubuntu 14.04.4 LTS the resulting `carp-repl` binaries are not properly linked causing the error:

```
ERROR: Failed to open file: "lisp/boot.carp"
 ----------------------------------------------------------------
  0 load-lisp                      main.c 1:0
 ----------------------------------------------------------------
��> 
```

I am not sure if this is a Carp problem, or something I need to fix in the CMake builds.

shyguy:
Ignore the previous comment. I was running it incorrectly when I should be running just `carp`, forgive me for that mess up :smile:. Although it is worth noting that the characters before the REPL do not display properly, probably should be put in a separate issue though. 
```
��>
```

doublec:
I hit the same problem and also did a fix. Unfortunately the fix in this pull request seems to require cmake 3.1 or greater. The distro I'm on is at cmake 3.0. Based on [this post](https://stackoverflow.com/questions/24840030/forcing-c99-in-cmake-to-use-for-loop-initial-declaration) I did [this commit](https://github.com/doublec/Carp/commit/734a0bc7d15444925675ca000a7526986f98351a) which works for me.

shyguy:
Another option is just to make another PR which changes all the forloops to be pre-c99 complaint. Or we could just bump up the required cmake version.

eriksvedang:
I added @doublec 's commit (https://github.com/doublec/Carp/commit/734a0bc7d15444925675ca000a7526986f98351a) and it still builds fine on OSX so I'll close this for now. Does this work for you too, @shyguy ?

shyguy:
Yup, it does the trick, thanks!

Issue #33, opened 2016-07-06
Mkdir, touch unsafety


tomsmeding:
The mkdir and touch implementations in core.karp just concatenate the given name into a shell expression. Things like `(mkdir "dir; malicious_program")` will therefore not do the expected thing; this is a problem if user input is directly given to mkdir and/or touch.
I'd expect not to have to quote input when giving it to those functions.
Fix would be to use fork/execlp not sure how feasible that is from lisp?

eriksvedang:
Yeah, those functions are not very good overall... I'll try to make safer versions

mkfifo:
So far Carp is a reasonably thin wrapper around the c functions for those - so I would say that at this point you should be validating all user input before handing to those functions.

In general, you should be validating *all* user input before doing anything with it.

Although I also agree that Carp should do it's best to make this easy and prevent obvious attack vectors.

tomsmeding:
Well, in C, the corresponding functions would *not* have the attack vector. That's pretty much my point.
A different example would maybe be in order:

`(mkdir "a b")`

Should this make two directories `a` and `b`, or should it make one directory `a b`? I would strongly expect the second, while the first is what happens as far as I can see.

I like the idea of Carp very much so far, and although I haven't used it yet, I'm wanting to do that in the near future. And even though I probably won't be using these functions, I wouldn't like a vulnerability à la SQL injection to be present in the very language -- think php before they understood escaping and prepared statements were a good idea.

eriksvedang:
Yeah, agreed. Right now these commands just use system() which is also used to call the external c compiler. I just made these as handy little things to use when in the repl, to not have to turn it off to check for file output, etc. I will fix this eventually, but for now they will remain dangerous. If you wanna write a safe version of one or more of them, feel free to do so!

mkfifo:
@tomsmeding ahh very good point about this exposing an additional set of attack vectors.

TY for the example.

My general point was that all input should be validated anyway, regardless of the safety of the underlying implementation.

However I also think we should try make them safe, if I have time I will try have a peek at this.

mkfifo:
I have written a first attempt at addressing this, please see #37

tomsmeding:
That looks great for safety (y)
Then the only point is that one might want to make a directory named "a b", but that's a functionality issue rather than a safety one. :)

mkfifo:
Yeah, I was unsure what set of characters to allow exactly, so I went for a simple set that we know work safely.


[...]

PR #37, opened 2016-07-09
Safe(r) mkdir and touch

mkfifo:
Attempting to address #33

I have defined a set of characters allowed within mkdir and touch, both will refuse any string which isn't made up entirely of those characters.

I have also changed boot.lisp to load builtins before core, so that core can use builtins in the implementation - I think this makes sense.

### DIFF
--- a/lisp/boot.carp
+++ b/lisp/boot.carp
@@ -16,8 +16,8 @@
 (def carp-dev (env-variable-true? (getenv "CARP_DEV")))
 
 ;; ~~~ CORE ~~~
-(load-lisp (str carp-dir "lisp/core.carp"))
 (load-lisp (str carp-dir "lisp/builtins.carp"))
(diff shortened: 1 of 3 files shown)


eriksvedang:
Cool, I'll merge this when I'm back at my computer tomorrow or Monday.

shyguy:
What about spaces and special characters (as pointed out by @lieuwex)?

mkfifo:
@shyguy as mentioned on #33 I was planning to add those later, happy to drop in favour of #43

mkfifo:
I'm breaking out the other parts of this commit in preparation for this being closed.
#44 - changing boot ordering (builtins before core)
#45 - string contains

mkfifo:
rebased and squashed down.

depends on #44 and #45 landing first.

mkfifo:
Closing in favour of #43

PR #40, opened 2016-07-09
Core.any short circuit

mkfifo:
Also including commit to add some more documentation to lisp/core.carp

### DIFF
--- a/lisp/core.carp
+++ b/lisp/core.carp
@@ -103,10 +103,23 @@
   (array-contains? (chars str) char))
 
 (defn any (pred xs)
-  (match xs
-    () false
-    (x) (pred x)
-    (x ... rest) (if (pred x) true (any pred rest))))
+  ;; short circuit if any (pred item) is true
+  (let [cont true
+        res  false]
+    (do
+      (while cont
+        (match xs
+          ()  (do
+                (reset! cont false))
+          (x) (do
+                (reset! cont false)
+                (reset! res (pred x)))
+          (x ... rest) (if (pred x)
+                           (do
+                             (reset! cont false)
+                             (reset! res  true))
+                           (reset! xs rest))))
+      res)))
 
 (defn true? (x) (if x true false))
 (defn false? (x) (if x false true))
--- a/lisp/core.carp
+++ b/lisp/core.carp
@@ -103,7 +103,10 @@
   (array-contains? (chars str) char))
 
 (defn any (pred xs)
-  (contains? (map pred xs) true))
+  (match xs
+    () false
+    (x) (pred x)
+    (x ... rest) (if (pred x) true (any pred rest))))
 
 (defn true? (x) (if x true false))
 (defn false? (x) (if x false true))
--- a/lisp/core.carp
+++ b/lisp/core.carp
@@ -176,8 +176,26 @@
 (defn true? (x) (= true x))
 (defn false? (x) (= false x))
 
-(defn all? (pred xs)
-  (= (count xs) (count (filter pred xs))))
+(defn all? [pred xs]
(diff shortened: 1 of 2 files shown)


eriksvedang:
Nice, I think the problem with this version is that Carp isn't tail recursive at the moment :/ I'll make it possible to implement a better version using while/return (which will compile nicely to C also)

mkfifo:
ahh good point, I'm used to working in scheme implementations with TCO.

Are while/return working now?
I am happy to convert this to them - I also want to do 'all?' as it can also short-circuit.

mkfifo:
okay so I can see that `return` doesn't yet exist.

I will try have a poke around and see if it is easy to add, my current issue is I cannot easily see how long the previous function call was - so I am unsure how many items to stack_pop.

mkfifo:
I have moved the documentation commit back to it's own PR #46

mkfifo:
I have written a version of `all?` which short circuits, it is a bit ugly though as I lack a real return.

```
  (defn all? [pred xs]                                                               
+   ;; short circuit if any (pred item) is false                                     
    (let [cont true                                                                  
          res  false]                                                                
      (do                                                                            
        (while cont                                                                  
          (match xs                                                                  
            ()  (do                                                                  
                  (reset! cont false))                                               
            (x) (do                                                                  
                  (reset! cont false)                                                
                  (reset! res (pred x)))                                             
            (x ... rest) (if (not (pred x))                                          
                             (do                                                     
                               (reset! cont false)                                   
                               (reset! res false))                                   
                             (do                                                     
                               (reset! xs rest)                                      
                               (reset! res true)))))                                 
        res)))  
```
What do you think?

testing
```
λ> (defn is-a? (x) (= x \a))
(fn (x) (= x \a))

λ> 
()

λ> (defn is-a? (x) (= x \a))
(fn (x) (= x \a))

λ> (all? is-a? (list \a \b \a))
false

λ> (all? is-a? (list \a \a \a))
true

λ> (all? is-a? (list))
false

λ> (all? is-a? (list \a))
true

λ> (all? is-a? (list \b))
false

λ> (all? is-a? (list \b \b \a))
false

λ> (all? is-a? (list \a \b \a))
false

λ> (all? is-a? (list \a \a \a))
true
```

mkfifo:
likewise I have a matching version of `any`

```
(defn any (pred xs)                                                             
  ;; short circuit if any (pred item) is true                                   
  (let [cont true                                                               
        res  false]                                                             
    (do                                                                         
      (while cont                                                               
        (match xs                                                               
          ()  (do                                                               
                (reset! cont false))                                            
          (x) (do                                                               
                (reset! cont false)                                             
                (reset! res (pred x)))                                          
          (x ... rest) (if (pred x)                                             
                           (do                                                  
                             (reset! cont false)                                
                             (reset! res  true))                                
                           (reset! xs rest))))                                  
      res)))   
```

and some testing

```
λ> (defn is-a? (x) (= x \a))
(fn (x) (= x \a))

λ> (any is-a? (list \a \b \a))
true

λ> (any is-a? (list \a \b \c))
true

λ> (any is-a? (list \d \b \c))
false

λ> (any is-a? (list))
false
```

mkfifo:
both new versions are now part of this PR, they are a bit ugly though.

Please let me know what you think.

mkfifo:
So two interesting points:e

1) my version of `core.all?` will return false on the empty-list, I think the old version returns true. Which is desirable?

2) I am unsure how match works - in both my examples I have 2 cases that may be able to be squashed down into one - can `(x ... rest)` capture the empty list as `rest`? - if so my `(x)` case(s) may be removed.

PR #42, opened 2016-07-09
Fix backslashes in strings in the repl

tomsmeding:
Something like `(println "test \" test")` compiled fine when the code was in a file, but in the repl, it mistakenly thought that a string was not terminated because it didn't interpret the `\`.

This pull request fixes that issue, and thereby enables one to have a `"` inside a string in the repl.


eriksvedang:
Very nice, will merge tomorrow!

PR #43, opened 2016-07-09
Make mkdir and touch safe and without surprises

tomsmeding:
This 1. makes mkdir and touch safe against shell injection vulnerabilities,
and 2. still allows the user to create any kind of file or directory.

This would fix my #33 I think.

### DIFF
--- a/lisp/core.carp
+++ b/lisp/core.carp
@@ -286,11 +286,19 @@
           text
           (map2 list (keys substitutions) (values substitutions))))
 
+(defn shell-quote [arg]
+  (str
+    "\"" ;; surround with quotes
+    (str-replace
+      (str-replace arg "\\" "\\\\") ;; replace backslashes
+      "\"" "\\\"") ;; replace double-quotes
+    "\""))
+
 (defn ls () (system "ls"))
 (defn pwd () (system "pwd"))
 (defn user () (getenv "USER"))
-(defn mkdir (dir-name) (system (str "mkdir " dir-name)))
-(defn touch (file-name) (system (str "touch " file-name)))
+(defn mkdir (dir-name) (system (str "mkdir " (shell-quote dir-name))))
+(defn touch (file-name) (system (str "touch " (shell-quote file-name))))
 
 (register-builtin "platform" '() :int)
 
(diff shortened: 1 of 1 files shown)


tomsmeding:
Sorry, this doesn't make sense at all. Will fix, hang on...

tomsmeding:
This should work, and still allow any name to be specified (including slashes etc.)

shyguy:
Possibly conflicting with #37, although #37 looks a little cleaner.

lieuwex:
@shyguy This PR seems to allow spaces and other special characters in the name of files. #37 seems not to.

shyguy:
Ah yes, I didn't pay enough attention to that. It's up to @eriksvedang to decide which he likes better. I personally think that the shell-quote function would use useful as part of the standard library even, or for other internal functions that use the shell.

eriksvedang:
Looks great, I'll compare the solutions and merge when I'm back at my work computer after the weekend.

mkfifo:
As author of #37 I think this (#43) is better, assuming it is safe.

I think it captures all the cases that are obviously dangerous.

tomsmeding:
@mkfifo: Thanks, and I definitely agree that calling a shell routine should be preceded by correct escaping/quoting/etc. My point, which I'm going to repeat here again (at the risk of becoming annoying...), is that I feel like *these routines should not be like shell routines to the programmer*. They might use the shell, but to the user they must be normal functions like any other.
</annoy>

I've finally gotten around to cloning the repo and playing around with it, as you've noticed, and I like this idea very much. Please continue developing this!

mkfifo:
@tomsmeding I wasn't making that comment to you in particular, I just wanted to make sure people understood the risks.
I also agree these should appear like normal functions, and their implementation should be hidden behind the interface.

Therefore this function should deal with any escaping specific to the implementation (in this case - the shell), and the caller should be responsible for general user input validation/scrubbing.

I also think that my iffyness around special characters is not a programming-level concern, and as mentioned in the earlier thread I was hoping to eventually add quoting in, so I think this (#43) is a better approach.

So I think we should merge this (#43) instead - and close my earlier work #37.

mkfifo:
@tomsmeding thanks!

tomsmeding:
@mkfifo No problem :)

PR #46, opened 2016-07-10
adding some more documentation to core.carp functions

mkfifo:


mkfifo:
rebased to fix collision, should apply clean now.

PR #50, opened 2016-07-11
load dylib libGL.so under linux and use for gl functions, otherwise continue using glfw

mkfifo:
this gets `./bin/carp examples/spin.carp` working for me under linux


paines:
Hi,

sorry I must have missed #50.Was it merged ? I just saw that the approach is almost the same as mine.... If #50 is going to be merged, i will just ditch my patch. Anyways I am happy to got it fixed :).
Thanks for chiming in!

mkfifo:
@paines hi there, #50 has been merged - but your change (although the same functionally) is cleaner code.

So I think yours should still be merged, I have rebased it in #57 
Does #57 look okay for you? it should be your changes updated to current master.

PR #56, opened 2016-07-13
Small fix for GL stuff

paines:
So I updated from Ubuntu 15.10 to 16.04 and realized that spin and other GL examples would not run, but they compiled fine. On startup you would get a complain that glClearColor etc could not be found. The patch now loads glClearColor etc... (all GL functions you would register in gl.carp from  line 37 on) from the library libGL.so.1. This is very linux specific I cross tested with my mac, and the code there still compiles and runs fine.

### DIFF
--- a/lisp/gl.carp
+++ b/lisp/gl.carp
@@ -1,9 +1,15 @@
+(if (linux?)
+    (do
+       (def *LIBNAME* "libglfw.so.3")
+       (def gl (load-dylib "libGL.so.1"))) "")
 
-(if (osx?) (def *LIBNAME* "libglfw3.dylib") "")
-(if (linux?) (def *LIBNAME* "libglfw.so.3") "")
+(if (osx?)
+    (do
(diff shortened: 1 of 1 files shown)


mkfifo:
I can confirm this change works for me under linux (arch linux, current).

Github seems to think there are conflicts, you should try rebasing on top of current master and then force push.

You should also be able to squash these 3 commits down into 1, since none of them works without the other - and they all have the same title.

If you would like for me to handle the rebase and squash for you let me know :)

mkfifo:
@paines notice that your checkout is a bit behind current carp master "This branch is 3 commits ahead, 147 commits behind carp-lang:master"

Opengl under Linux was initially fixed in #50 

I have rebased your commits ontop of current master, squashed it down to a single commit, and added you to contributors file.

@paines please review #57

paines:
Looks fine to me!

mkfifo:
Awesome, thanks for the work @paines :D

PR #69, opened 2017-07-01
Minimal readline support in REPL

kmate:
Allows to move in the input line with cursors and also keeps an in-process history. Not perfect, but it helps a lot, especially typing up arrow to repeat the last command.

### DIFF
--- a/CarpHask.cabal
+++ b/CarpHask.cabal
@@ -46,6 +46,7 @@ executable carp
                      , CarpHask
                      , containers
                      , process
+                     , readline
   default-language:    Haskell2010
 
 test-suite CarpHask-test
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,6 +1,7 @@
 module Main where
 
 import Control.Monad
+import System.Console.Readline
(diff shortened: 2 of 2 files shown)


eriksvedang:
Cool! I've been using rlwrap for this but a more integrated solution is obviously much better in the long run. I'll merge after the weekend when I'm back at my big computer. Thanks!!!

eriksvedang:
Seems like this package adds a dependency for the "readline" command line tool. Do you know if this utility exists on Windows?

kmate:
Well, there is something among the [GNUWin packages](http://gnuwin32.sourceforge.net/packages/readline.htm).

hellerve:
I’d suggest using [Haskeline](https://hackage.haskell.org/package/haskeline) instead, which is a Haskell implementation of editline. That makes it a little more portable, but its interface is a little different. I’ve used it before and could aid in the process of shifting this over to Haskeline if desired.

eriksvedang:
That's an excellent idea, let's do that instead!

PR #75, opened 2017-10-19
Use haskeline for REPL

hellerve:
This PR introduces Haskeline, a pure Haskell implementation fo editline, into Carp’s REPL. This means we have regular editline capabilities now, including history. Should we want to tweak the settings, we will have to move from `defaultSettings` to something more advanced, but I suggest we stick to this for now, as it is already better than what we used to have.

Cheers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 /out/main.c
 /.DS_Store
 /Carp.prof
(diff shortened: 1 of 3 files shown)


eriksvedang:
Awesome, thanks!

PR #79, opened 2017-10-23
Added Test module

hellerve:
This PR introduces a whole lot of stuff, most notably a test module. This is a WIP, and I’d advise against merging it at this time. It is, however, in a state in which we should be starting a conversation about it.

Introduced by this PR:
- `=` and `/=` for `Bool`
- `/=` for `Int`
- `/=` for `String`
- `from-chars` for `String`: makes a string from an array of characters (can be removed, not necessary for this PR)
- `repeat` for `String`: repeats a given string `n` times
- `string-join`: append multiple strings (TODO: naming)
- introduce the testing library

What do y’all think?

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -155,6 +155,7 @@ preludeModules carpDir = map (\s -> carpDir ++ "/core/" ++ s ++ ".carp") [ "Macr
                                                                          , "Bool"
                                                                          , "IO"
                                                                          , "System"
+                                                                         , "Test"
                                                                          ]
 
 main :: IO ()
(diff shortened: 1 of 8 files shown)


eriksvedang:
Looks very nice! I'll play around with it a bit and see if I have any comments.

eriksvedang:
The functions on line 50 and 51 in 'prelude.h' must return bool to work on my system:

```
bool Int__EQ_(x, y)     { return x == y; }
bool Int__DIV__EQ_(x, y) { return x != y; }
```

hellerve:
Makes sense. Seems your compiler is pickier than mine!

hellerve:
I fixed the return type for the comparison functions. They were all wrong, I suppose that was just historical baggage. It only turns up when using it in a higher-order fashion, though.

PR #80, opened 2017-10-24
Vector module

hellerve:
This PR adds a Vector module. Vectors come in the flavors `Vector2`, `Vector3` and `VectorN`. 2- and 3-dimensional vectors are structs with special fields, whereas n-dimensional arrays are basically just arrays on which we can perform vector operations.

## The API of Vectors

The following functions are included:
- Arithmetic operations: i.e. `+`, `-`, `*`, and `/`
- Logic operations: i.e. `=`, `/=`, and `approx` (which checks whether the vectors are identical except for negligible rounding errors)
- `mag` to get the magnitude of a vector
- `mag-sq` to get the squared magnitude of a vector
- `normalize` to normalize a vector
- `dist` to get the distance between two vectors (only 2D)
- `heading` to get the direction of the vector (only 2D)
- `rotate` to rotate a vector (only 2D)
- `dot` to get the dot product of two vectors
- `cross` to get the cross product of two vectors (only 3D)
- `angle-between` to get the angle between two vectors (only 2D and 3D, although ND would be possible)
- `lerp` to linearly interpolate between to vectors

### Open questions
- Did I miss anything?

## Other miscellanous additions

The following is included to make this work:
- A Geometry module (currently only handles conversion between radians and degrees)
- Logical operators on doubles: `=`, `>`, `<`, and `approx` (see above)
- Mathematical functions on doubles: `acos`, `atan2`, `clamp`, and `sqrt`
- prelude: converted macros to functions to enable higher-order handling
- haskell: fix bug in `multiLookupQualified` where a case wasn’t treated.
- test: changed API so that datatypes have to provide custom `str` function to enable genericism
- added test suites for all of them

### Open questions
- Do we need any more geometry functions for now?
- Should all operators be implemented for all numerical datatypes?
- Why do the test suites for Vector3 and VectorN not work? **IMPORTANT; FIXME BEFORE MERGE**

I think that’s it!

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -157,6 +157,7 @@ preludeModules carpDir = map (\s -> carpDir ++ "/core/" ++ s ++ ".carp") [ "Macr
                                                                          , "IO"
                                                                          , "System"
                                                                          , "Test"
+                                                                         , "Vector"
                                                                          ]
 
 main :: IO ()
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -158,6 +158,7 @@ preludeModules carpDir = map (\s -> carpDir ++ "/core/" ++ s ++ ".carp") [ "Macr
                                                                          , "System"
                                                                          , "Test"
                                                                          , "Vector"
+                                                                         , "Geometry"
                                                                          ]
 
 main :: IO ()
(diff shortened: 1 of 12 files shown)


eriksvedang:
It looks very nice, can't wait to merge this!
I'll take a look at the failing tests and see if I can figure it out.

eriksvedang:
Also – I've now added `(add-hook 'before-save-hook 'whitespace-cleanup)` to my emacs config to avoid all the diffs regarding trailing whitespace :)

hellerve:
I fixed the tests. It required me to fix `Array.copy` as well, which up until now produced bogus values. I hope the fix is adequate!

PR #83, opened 2017-10-26
Statistics: add module

hellerve:
This PR adds a statistics module and a few required changes to the core to make this work. All statistical functions operate on doubles, because we need functions like `sqrt` to work. If you want to use this library with any other numerical type, you will have to cast it first.

The API is fairly minimal, because I’m not sure I’d be the right person to implement fancier statistical functions.

## The API of Statistics

The following functions are included:
- `sum`, a helper function to sum an Array
- `mean` calculates the mean of an Array
- `min` gets the minimum element in an Array
- `max` gets the maximum element in an Array
- `median`, `low-median`, and `high-median` calculate various flavors of the median
- `grouped-median` calculates another flavor of the median. You need to specify the interval size for it to work.
- `variance` calculates the variance
- `pvariance` calculates the population variance
- `stdev` calculates the standard deviation
- `pstdev` calculates the population standard deviation

## Other miscellaneous additions

To enable this, I had to include the following items:
- `Array.sort` sorts an Array given a sorting function (based on the C function `qsort`)
- `Array.index-of` finds the first index of an element in an Array
- `Array.element-count` counts the occurrence of an element in an Array

### Open questions
- `index-of` and `element-count` are implemented as templates. That’s not necessary, but simplified the implementation, because no custom comparison function had to be injected. It means that the comparison uses `==`, however, which will only work on simple types. This should probably be fixed in the future.

A minimal test suite is included.

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -125,6 +125,9 @@ arrayModule = Env { envBindings = bindings, envParent = Nothing, envModuleName =
                                 , templateDeleteArray
                                 , templateCopyArray
                                 , templateStrArray
+                                , templateSort
+                                , templateIndexOf
+                                , templateElemCount
                                 ]
 
 startingGlobalEnv :: Env
@@ -160,6 +163,7 @@ preludeModules carpDir = map (\s -> carpDir ++ "/core/" ++ s ++ ".carp") [ "Macr
                                                                          , "Vector"
                                                                          , "Geometry"
                                                                          , "Test"
+                                                                         , "Statistics"
                                                                          ]
 
 main :: IO ()
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -6,8 +6,8 @@
         (for [i 0 (count xs)]
           (set! &total (f &total (nth xs i))))
         total)))
-  
-  )
+
+)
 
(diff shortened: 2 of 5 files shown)


eriksvedang:
Looks good, will merge soon. I have some ideas for how to solve the problem where you have to inject '=' and similar cases, so let's hope we don't need many more template functions (and can remove some of them eventually). I think the templates added in this PR could potentially find the correct equality function also, using the 'findFunctionForMemberIncludePrimitives' function?

hellerve:
Quite possibly. I’m not quite sure how I’d do that, sadly. I’ll look into it.

Of course as Carp moves along I’ll retrofit all changes to the modules that I introduce.

eriksvedang:
Hello, 
Sorry for taking so long to merge, caught the flu this weekend so haven't been able to work for a couple of days. I'll fix the thing regarding '=' so you can see how I think it should be done.
Also – I noticed there are quite a few nested if-statements in this code, perhaps using the 'cond' macro could look nicer?

hellerve:
No problem at all! Glad you’re better!

The statement regarding the `if` is certainly true, I’ll fix that right away!

PR #88, opened 2017-11-07
Statistics, revisited

hellerve:
This PR introduces the missing tests from #87, where the statistics module was extended but never got the love it needed. In the course of that, I also added the function `Double.abs` for taking the absolute value of a `Double`, and fixed a bug in `median-abs-dev`, which failed to adhere to its promise of taking said absolute value.

Cheers


eriksvedang:
Excellent!

Issue #93, opened 2017-11-14
Add all the math functions


eriksvedang:
@hellerve has added a whole lot of math functions to the Double module. The Float module is way behind though, and there are also probably a few useful functions left to add for all the different math modules (Int / Long / Float / Double).

eriksvedang:
I guess we can close this now?! (yay!!!)

hellerve:
I’d like to add some more integral math functions, but in general we’re pretty ready. Give me a day to think this over.

PR #95, opened 2017-11-15
Math functions, Vol. I: Math for Doubles

hellerve:
This PR introduces all math functions found in `math.h` into the `Double` module, in reference to #93 . It also introduces a simple smoke test suite.

The list of functions introduced can be found in [the reference](http://www.cplusplus.com/reference/cmath/); use cases are attached in the test suite.

This is the first installment of a few to introduce useful mathematical functions to our numerical types.

Cheers


eriksvedang:
Wow, awesome!

PR #96, opened 2017-11-16
Math functions, Vol. II: Math for Floats

hellerve:
This PR introduces all math functions found in math.h into the Float module, in reference to #93. It is congruent to #95, and also introduces a simple smoke test suite.

The list of functions introduced can be found in the reference; use cases are attached in the test suite.

Cheers


eriksvedang:
Very nice.

PR #97, opened 2017-11-16
RFC: safe arithmetic

hellerve:
This PR introduces the functions `safe_add`, `safe_sub`, and `safe_mul` to the `Int`/`Long` types. It uses `__builtin_<type-shorthand><op>_overflow()` internally, which is a compiler builtin that has been supported by major compilers—at least Clang and GCC—for quite some time. I’m not sure whether it is perfectly portable, but it is hugely beneficial to correct software. The interface is always:

```
(safe-fun op1 op2 &res); => returns boolean saying whether op overflowed
```

You can thus write software like this:

```
(let [res 0]
  (if (safe-add possibly-huge-number another-huge-number &res)
     (IO.println "It overflowed! Oh no")
     (IO.println &(String.append @"Result: " (Int.str res)))))
```

which is obviously good for error handling.

I personally consider it to be a huge win, but compatibility might stop it from happening.

A test suite is included.

Cheers

### DIFF
--- a/core/Int.carp
+++ b/core/Int.carp
@@ -17,5 +17,9 @@
   (register inc (λ [Int] Int))
   (register dec (λ [Int] Int))
   (register copy (λ [&Int] Int)) ;; TODO: Should not be needed when refs to value types are auto-converted to non-refs.
+
+  (register safe-add (λ [Int Int (Ref Int)] Bool))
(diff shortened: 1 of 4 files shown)


hellerve:
I consider this to be part of #93, though admittedly differently than the last two installments and thus not named as another installment.

PR #98, opened 2017-11-17
Math functions, Vol. III: Math for Ints and Longs (Closer)

hellerve:
This PR is the third and last installment in our beloved series “Math functions”. It adds a few functions for Ints and Longs, namely `min`, `max`, and `abs`. I feel like math on integers is and should be restricted, but if you’re missing any functions, please let me know!

As always, a test suite is included.

Cheers


hellerve:
Also, for housekeeping purposes, closes #93.

PR #100, opened 2017-11-19
Apply hlint fixes

hellerve:
This PR introduces a few hundred suggestions given by hlint.

After rebasing, there is one hlint fix still open (complaining about code duplication in `moduleForDeftype` and `bindingsForRegisteredType`). I didn’t fix that one because I consider the duplication there to be okay actually, because the duplicated sections use so many closured variables that it would be a pain to introduce this function. If you disagree, I can work on a patch to reduce duplication there as well.

Cheers


eriksvedang:
Awesome, thanks for the hard work! Really nice improvements, there's a lot to learn from hlint :)

I have a few disagreements with it's choices though, so if it's OK with you I'd like to omit the following changes:

* AssignTypes.hs:10, Deftype.hs:51, Obj.hs:334, Obj.hs:429. I enjoy the point-free style when it's easy to mentally parse, but in these cases I feel that the argument names are there as documentation and to help you read the code quicker.

* GenerateConstraints.hs (line 45). I don't really understand how to use Arrows and as a principle avoid putting things I don't understand into the code base. Another solution to this problem would be for me to read up on arrows, so maybe I should do that :)

* Parsing.hs:99,254,259. The arrays of chars is a stylistic choice to remind the reader of the code of which parameter is which.

hellerve:
Yeah, I anticipated some of these complaints.

As for the arrows, in this case they are fairly simple: all we do is combine two functions that each take one argument—in this case even the same function—to a function that takes two arguments and return a tuple. In general, all that arrows do is define sort of function-combinatorial pipelines, where you combine and split arguments and return values of functions. This is all a bit hand-wavy, but should give you an intuition into how it works.

As for the others, I completely agree with all of your complaints. I can instead rely on annotations to locally disable hlint bickering at these locations; would that suit you better?

eriksvedang:
Sure, sounds good!
Let's go for the arrow then, I admit it's pretty cool =)

hellerve:
This should now be fixed.

eriksvedang:
Epic!

PR #121, opened 2017-12-03
Length-prefixed strings

### LINKED ISSUE #119: Suggestion: Use Pascal-strings instead of only C-strings
Pascal strings are length-prefixed strings in contrast to Zero-terminated C strings.

https://en.wikipedia.org/wiki/String_(computer_science)#Length-prefixed

They have many advantages over C strings and no drawbacks [1] when you have something like [std::string_view](http://en.cppreference.com/w/cpp/string/basic_string_view) or [Rust's slices](https://doc.rust-lang.org/book/second-edition/ch04-03-slices.html).

[1] One might still say the extra memory overhead for the length field is a problem.

philix:
Closes #119

### DIFF
--- a/core/SDLHelper.h
+++ b/core/SDLHelper.h
@@ -1,16 +1,17 @@
 #include <SDL.h>
+#include "core.h"
 
 typedef struct {
     SDL_Window *window;
     SDL_Renderer *renderer;
 } App;
 
-App app_MINUS_init(const char *title, int width, int height) {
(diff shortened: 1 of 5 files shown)


philix:
@eriksvedang what you think? `./run_carp_tests.sh` works completely.

eriksvedang:
Looks great! (Thanks!) Will merge soon.

eriksvedang:
Hmm, got some problems with this patch now when I tried it out more. Will explain in #119.

PR #122, opened 2017-12-04
Miscellaneous fixes

hellerve:
This PR fixes a few problems that I’ve encountered while getting Snake to work, specifically:

- add a `copy` function for `Bool` types
- add a `copy` and `str` function for `SDL_Keycode` types
- fix the compiler so that it looks for a `str` function that matches even if the type is external
- remove `copy` functions from the vector modules, because they conflict with the types

Cheers


eriksvedang:
Great!

PR #124, opened 2017-12-12
Break

hellerve:
This PR adds `break` as a function-like statement. It’s functionally equivalent to `break` in C, allowing us to exit loops early.

Test cases are included.

Cheers

### DIFF
--- a/out/keep.txt
+++ b/out/keep.txt
@@ -1,1 +1,0 @@
-Keep this directory.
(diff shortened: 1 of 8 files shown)


eriksvedang:
Tiny, tiny nitpick: line 76 in Emit.hs should follow the pattern of the functions around it (if it gets a match some other part of the emitter messed up). 

Everything else looks ace!

hellerve:
Done. This was a remnant of a prior commit. Good catch!

PR #126, opened 2017-12-13
Add when and unless macros

hellerve:
This PR adds the macros `when` and `unless`. Those are single-branch alternatives to `if` that are of type `()` and thus only make sense when they’re side-effectual.

```clojure
(let [x 0]
  (when true
      (set! &x 1))) ; x will be 1

(let [x 0]
   (unless false
      (set! &x 1))) ; same
```

Test cases are included.

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -99,3 +99,6 @@
 
 (defmacro when [condition form]
   (list 'if condition form (list)))
+
+(defmacro unless [condition form]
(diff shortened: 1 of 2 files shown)


eriksvedang:
Huh, gets a conflict?

hellerve:
A rebase fixed it. Probably sale commits.

eriksvedang:
merci beaucoup

PR #129, opened 2017-12-14
Add useful macros to the standard library

hellerve:
This PR introduces a few useful macros to the standard library. It also changes the behavior of `cond`.

## Summary of macros added

- `do-let`: combines `let` and `do`.
- `defn+`: combines `defn` and `do`.
- `comment`: comments a block of code by ignoring the forms passed into it.
- `forever`: runs a sequence of forms forever.
- `case`: simple equality matching, similar to `case` in Scheme.

## Changes to other parts of Carp

I took the liberty of changing the definition of `cond` to expand into a last `if` statement if the number of forms  is even. That way it might still fail when type-checking, but not silently remove the last conditional clause. This is a net positive, but maybe using `macro-error` instead would be even better.

Some test cases are attached, although many of these macros are hard to test.

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -103,7 +103,7 @@
 (defmacro unless [condition form]
   (list 'if condition (list) form))
 
-(defmacro do-let [bindings :rest forms]
+(defmacro let-do [bindings :rest forms]
   (list 'let bindings
     (cons 'do forms)))
 
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -2,7 +2,7 @@
   (if (= (count xs) 0)
     (list)
     (if (= (count xs) 2)
-      (macro-error "case has even number of branches; add an else branch")
+      (macro-error "cond has even number of branches; add an else branch")
       (if (= (count xs) 1)
         (car xs)
         (list
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -2,7 +2,7 @@
   (if (= (count xs) 0)
     (list)
     (if (= (count xs) 2)
-      (list)
+      (list 'if (car xs) (car (cdr xs)) ())
       (if (= (count xs) 1)
         (car xs)
         (list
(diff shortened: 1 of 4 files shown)


eriksvedang:
Cool! Here are some thoughts before merging:

How do you feel about changing the name of `defn+` to `defn-do` instead? I have a hunch that there are other macros that enhances `defn` that could be worthwhile later on (like one that takes a type signature perhaps), and `+` seems a bit easy to forget what it's about. It also nicely documents that this is a side-effecting function.

If you agree that change is a good idea, I would also propose that all the other forms that add `do` inside them are named in the same way. `let-do` and perhaps even `forever-do`. It has been an (undocumented, I'm afraid) philosophy of Carp to not use implicit do, so I think it's nice if it's spelled out in the macros. Do you agree?

The case macro is nice. I believe it will eventually be subsumed by the pattern-matching facilities that are needed for ADT:s. But that transition should be very smooth since it will work exactly the same, if I'm not mistaken.

Thanks for finding and fixing the error with `cond`, I didn't think it would silently discard things. I think an error message could be better, since debugging macros is unnecessarily hard (I'm sure you're aware...)

hellerve:
Hey!

That’s true. I opted for `+` because I was a bit lost with the naming, but I guess `-do` is a better suffix. I also added it to the other names, although it does make them a little clunky. `forever` especially has to be a side-effecting function (otherwise it would make no sense to run something again and again), so I think it’s fairly explicit about that. But you are right.

I’m not as clear as you on the vision of ADTs, so I don’t quite know how congruent the API will be in the end, but `case` has always been useful for me in Scheme, so I thought that adding it here couldn’t hurt. It implements a linear search through the cases, though, and should at one point be optimized, for instance with the insights from [this paper](http://scheme2006.cs.uchicago.edu/07-clinger.pdf), but for now I think the implementation is just fine.

I also made `cond` and `case` throw errors on an even number of clauses.

eriksvedang:
Great! Now I get an error in test/statistics.carp when running all the tests though. Seems to be a `cond`?

hellerve:
Alright, this is also fixed. Turns out that `cond` was actually masking an error already!

eriksvedang:
Very nice.

Issue #94, opened 2017-11-14
Useful functions missing in the String module?


eriksvedang:
The String module is probably lacking a lot of useful functions commonly found in other standard libraries. Let's add them!

hellerve:
I have a list of useful functions (this is a verbatim copy of what’s present in zepto):

- `(String.join strs delim)`: joins a list of strings with a delimiter
- `(String.split str delim)`: splits a string on a delimiter, returns a list of substrings
- `(String.in? str sub)`: checks whether a substring is within a string
- `(String.count str c)`: counts the number of occurrences of character c in a string
- `(String.starts-with? str sub)`: checks whether a string starts with a substring
- `(String.ends-with? str sub)`: checks whether a string ends with a substring
- `(String.trim str)`: trims whitespace from a string
- `(String.upper? str)`: checks whether string is upper case
- `(String.lower? str)`: checks whether string is lower case
- `(String.num? str)`: checks whether string is numerical
- `(String.alpha? str)`: checks whether string is alphabetical
- `(String.alphanum? str)`: checks whether string is alphanumerical

There are more useful functions, but that’s a whole lot and a good start, I’d say.

Cheers

eriksvedang:
Great, yes this looks like a very decent start indeed.

Kungsgeten:
Maybe [s.el](https://github.com/magnars/s.el) (the long lost Emacs string manipulation library) can be used as inspiration?

hellerve:
Looks good! It also reminds me that we still have to wrap a regex library!

bjorn3:
* string to int
* int to string

hellerve:
Those are already provided as `Int.str` and `Int.from-string`! The floating point data types do not have `from-string` functions, but should probably have them...

bjorn3:
Should have looked at the source

hellerve:
Don’t worry! We have to document the standard modules sooner or later, to make it a bit easier for people to know what’s there.

eriksvedang:
I feel like this is OK to close now?

hellerve:
I haven’t ported everything from the long lost Emacs string library yet, nor do I intend to. SOme of the functions do not need to b included in the standard library, in my opinion.

That being said, I do think that a handful of functions from the library should be included, like `truncate` and `split-up-to`.

eriksvedang:
OK, let us leave it open until then, then :)

mkfifo:
[...]

PR #134, opened 2017-12-18
Strings I: A few functions

hellerve:
This PR introduces a few useful String functions, as discussed in #94. It is not quite the whole list yet, but it’s already a good seed.

## Functions added

- `(char-at s i)`: returns the character at index i in string s
- `(head s)`: returns the first character of a string s (will error if the string is empty)
- `(tail s)`: returns the string s, except for the first character (will error if the string is empty)
- `(pad-left len c s)`: will pad the string s on the left with the character c until it is of length len
- `(pad-right len c s)`:  same as above, but padding on the right
- `(count-char s c)`: count occurrences of character c in string s
- `(reverse s)`: reverses string s
- `(empty? s)`: checks whether the string s is empty (i.e. its a length zero string)

## Other functions

I also took the liberty of adding an ArrayTemplate for `reverse`, because I built `String.reverse` on top of that.

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -116,7 +116,6 @@ arrayModule :: Env
 arrayModule = Env { envBindings = bindings, envParent = Nothing, envModuleName = Just "Array", envUseModules = [], envMode = ExternalEnv }
   where bindings = Map.fromList [ templateNth
                                 , templateReplicate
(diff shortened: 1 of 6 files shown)


eriksvedang:
These are great!!!

PR #138, opened 2017-12-19
Fixed segfault in read-line when file doesn't exist.

Nananas:
When reading a non-existing file using `read-file`, currently a segfault happens.
This is more of a hack though. Maybe this function could return a `Maybe` type?

### DIFF
--- a/core/core.h
+++ b/core/core.h
@@ -273,6 +273,12 @@ string String_tail(string* s) {
   return news;
 }
 
+string String_empty() {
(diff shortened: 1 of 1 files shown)


hellerve:
Maybe we could also `register` `empty` in the String module?

Nananas:
`empty` would create a heap allocated empty string? Not sure if `""` vs `empty` would be a clear way to distinguish stack vs heap.

PR #140, opened 2017-12-19
Moved a few examples to test

hellerve:
This PR moves a few examples to tests and deletes some more. Before merging make sure that all the examples that are deleted are actually fine to be deleted.

Also, thie PR introduces equality functions for non-ref Strings, and fixes `/=` for Strings, which should’ve never been a C function anyway.

Cheers


eriksvedang:
I think I wan to keep the following files:

* interfaces.carp (or is this covered by some other example? I guess implicitly..? I think it could be worthwhile to write some tests for Interfaces before deleting the file)
* memory.carp, I'll convert this to proper tests now!
* mutual_recursion.carp (good to show how it can be done, should create a macro for making it look nicer though...)
* selection.carp (it should be converted to some test cases that ensure that the correct implementation is chosen)
* temp.carp (this is my scratch file that lets me share WIP on multiple computers, which admittedly is kinda lazy, but very convenient..)

hellerve:
Alright, we can do that. One thing: both selection and interfaces are needed for he test framework itself, so that’s why I didn’t add test cases for them. If they don’t work, none of our tests will even compile.

eriksvedang:
True, let’s nuke them then :) I guess it’s good to keep the number of files down.

PR #143, opened 2017-12-21
Add logic and arithmetic macros

hellerve:
This PR introduces `>`, `<`, `and`, `or`, and `not` to the macro system. It also generalizes `=` to work on all numeric types. This was discussed in #142.

A test suite is attached.

Cheers


eriksvedang:
Nice! I realised after merge:ing that the '<' and '>' functions probably should be put into Commands.hs since they are normal functions. '=' could also be put there

hellerve:
I’ll do that, then.

PR #145, opened 2017-12-22
Arithmetic/logic macro rework

hellerve:
This PR is an addendum to #143. It moves the arithmetic macros `=`, `>`, and `<` to Commands. It also reworks `and` and `or` to short-circuit.

Tests for short-circuiting are included.

Cheers


eriksvedang:
Excellent!

PR #147, opened 2017-12-23
Make example a bit more friendly

devn:
- Seeds based on time so the random number isn't the same every time
- Splits out a couple of functions and variables to demonstrate def and defn
- Shows how to format with string-join

If you want this, cool. If not, no hard feelings.

Thanks for building Carp. I'm enjoying tinkering with it so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,33 +47,40 @@ The Carp REPL has built-in documentation, run ```(help)``` to access it!
       (println "~ The Number Guessing Game ~")
       (println "(Enter q to quit.)\n\n")
 
+      (set! &guessing true)
       (set! &answer (random-between 1 100))
+
       (print "Please enter a number between 1 - 99: ")))
 
 (defn exit! []
   (do (println "Good bye...")
       (set! &guessing false)))
 
+(defn play-again? [user-input]
+  (if (= user-input "y\n") true false))
+
 (defn correct! []
   (do (println "Correct!")
-      (exit!)))
+      (print "Play again? (y/n): ")
+      (let [user-input (get-line)]
+        (if (play-again? &user-input)
+          (init!)
+          (exit!)))))
 
 (defn guess-again [low-or-high]
   (do (println &(string-join @"->Too " @low-or-high @"."))
       (print "\nPlease guess again: ")))
 
 (defn main []
-  (do
-    (init!)
-
-    (while guessing
-      (let [user-input (get-line)
-            guessed-num (from-string &user-input)]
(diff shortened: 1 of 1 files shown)


eriksvedang:
Cheers, glad to hear that! I think this might be a little bit too long to put on the front page, it's already kind of crowded over there. Your code is cleaner than the existing one though, so I'd be happy to put this new code into the existing file `examples/guessing.carp`.

I think the example on the front page should change to something more splashy anyways, probably a graphical thingy to show off a bit more.

devn:
Yeah, I agree it's a bit long for a README.

I was initially pretty confused about needing to explicitly seed, which is why I even began messing with it. I don't know enough about C to know whether seeding by default with an option to override should be a sensible default in Carp or not, but it didn't initially confuse me enough to delete `a.out`, restart the REPL a few times, etc.

Tangent ahead: I know that it is not an explicit design goal of Carp to match Clojure, but there are enough similarities that I sort of expected to do something like `(rand-nth (range 1 100))` instead of `(random-between 1 100)`. The name of `random-between` is also a little misleading as it suggests that it is exclusive on both sides. My 2¢.

Anyhow, I'll change it up and add my code to `examples/guessing.carp`.

Thanks!

eriksvedang:
Accidentally merged because I thought you fixed it so I swapped the code myself, now your solution is in `examples/guessing.carp`.

Yeah the need to seed is pretty weird, perhaps Carp should just do that for you. Coming from the C world it's just another annoying thing you have to learn early on :)

Yeah you could definitely write something like `(rand-nth (range 1 100))` but Carp has no iterator abstraction and would have to allocate a full Array of ints, which isn't very effective. There could definitely be some value in making a kind of iterator for generating values on-demand but since Carp:s main focus is effective Array processing it isn't anything we've focused on. Feel free to try to write something like that!

What would you call the (random-between) function if its behaviour wouldn't change?

devn:
That's a good question! :)

I don't have a decidedly better name, I don't think, but here goes:

- `(rand-in 1 100)` (still has the problem of suggesting exclusivity on both ends, and frankly `range` as a name says nothing about exclusivity either, that's in the docs)
- `(rand-from 1 100)` would suggest inclusivity on the left.

:man_shrugging: It would be cool if there were a way to express `random-between` without dropping to `c`. Like if a `range` could be expressed not as a sequence but as a floor and ceiling. Though, that may be confusing too. Thinking something like:

```clojure
(defn random-between [start end]
  (rand-from (range 1 100)))

PR #149, opened 2017-12-23
Add Float.from-int

hellerve:
This PR adds `Float.from-int`, which was notably absent from conversions between numeric types. It also fixes the signature of `Float.to-int`, which claimed it took a `Double` as argument.

Cheers


eriksvedang:
Nice catch!

PR #151, opened 2017-12-24
Fix for --core-flag

NOP0:
Hi Erik, please do check this work a bit extra if you dont mind. I checked with the repl that core.h is not included anymore, but sadly I'm not a Haskell wizard (yet) :-) 

Merry Christmas.


eriksvedang:
Sorry for being slow at merge:ing this, looks good to me!

PR #152, opened 2017-12-28
Better Vector API & Random

hellerve:
This PR began as a humble fix for some problems in Vector, but it has grown to be a bit more.

## The new Vector functions

I’m adding a few functions to the Vector API, namely:

- `zero`, which will create a zero’d vector (`zero-sized` for VectorN for dimensionality)
- `random`, which will create a random unit vector (`random-sized` for VectorN for dimensionality)
- `parallel?` to check whether two Vectors are parallel
- `anti-parallel?` to check whether two Vectors are anti-parallel
- `perpendicular?` to check whether two Vectors are perpendicular

Test cases are included.

## Other additions, most notably random

I took the liberty of adding a few interfaces, namely `mod`, `random`, and `random-between`.

I also implemented the following methods:
- `random` and `random-between` for Chars
- `random-sized` for Strings (there is no pretty way of defining `random-between` without closures)
-  `random` and `random-between` for Doubles
- `random` for Floats

Cheers

### DIFF
--- a/core/Char.carp
+++ b/core/Char.carp
@@ -6,10 +6,10 @@
   (register copy (Fn [&Char] Char))
 
   (defn random []
-    (from-int (Int.random) 255))
+    (from-int (Int.random)))
 
   (defn random-between [a b]
-    (from-int (Int.random-between (to-int a) (to-int b)) 255))
+    (from-int (Int.random-between (to-int a) (to-int b))))
 
   (defn meaning [char-ref]
     (cond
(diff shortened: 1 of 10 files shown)


eriksvedang:
What a nice bag of Christmas candy! Will merge when back at the laptop.

eriksvedang:
Seems like the Char random functions are wrong and don't compile.

hellerve:
Fixed. This was a remnant of me experimenting with manually limiting the range of the input. But I think overflows should also result in a uniform distribution (?).

PR #153, opened 2017-12-29
Add varargs macro

hellerve:
This PR (re-)introduces the varargs macros `and*`, `or*`, `str*`, and `println*`. `str*`/`println*` produces pretty ugly strings for now, because it calls `str` on Strings (you can observe that behavior in the last included test case, "str* macro works as expected".

Test cases are included.

Cheers


eriksvedang:
Cool! Yes, we need a ‘prn’ function for that, it’s in the cards for sure.

PR #154, opened 2017-12-30
Add formatted string output

### LINKED ISSUE #108: Consider adding formatted string output
Gitter discussion suggests that possible solution could be printf macro.

hellerve:
This PR fixes #108 and adds formatted strings. It is fairly big and introduces a breaking change, so review with caution!

It adds the dynamic functions `+`, `-`, `*`, and `/` and the dynamic module `String` containing the functions `char-at`, `index-of`, `substring`, and `count`. I refrained from making `count` generic, but that could be discussed as well.

I also made registering new commands more convenient, with an arity that then gets checked at dispatch time, which is a little more convenient than writing an error branch for every function. Multi-arity commands are TBD.

I also added the interface `format`, which takes a string and a thing and returns the string representation of that thing (so its `sprintf` with one argument). This is used in the `fmt` macro, which splits `sprintf`-like strings at compile time and transforms it into a concatenation. *NOTE:* This is possibly expensive!

There are also a few minor things, like making `=` for macros work on Chars and such.

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -137,48 +137,48 @@ arrayModule = Env { envBindings = bindings, envParent = Nothing, envModuleName =
 
 dynamicStringModule :: Env
 dynamicStringModule = Env { envBindings = bindings, envParent = Nothing, envModuleName = Just "String", envUseModules = [], envMode = ExternalEnv }
-  where bindings = Map.fromList [ addCommand "char-at" (CommandFunction commandCharAt)
-                                , addCommand "index-of" (CommandFunction commandIndexOf)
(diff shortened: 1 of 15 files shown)


eriksvedang:
This is so cool! From reading the code briefly it seems like the `fmt` functions implemented in core.h will leak, since they take a non-ref String and don't do anything with it, right? They should probably take refs if there is no good reason not to.

I like the changes to `addCommand`, much cleaner!

hellerve:
Ah yes, that’s true. I hadn’t realized that that’s what’s going to happen, but it makes sense. I’ll fix it tomorrow.

PR #158, opened 2018-01-08
Small project cflag/libflag fix

Nananas:
- `AddIfNotPresent` should append to list (or at least in this case), so `cflag` & `libflag` are in the right order.
- Fix: `libflag` should fill `projectLibFlags`, instead of `projectCFlags`.


eriksvedang:
Thanks!

PR #159, opened 2018-01-08
Miscellaneous fixes

hellerve:
This PR fixes multiple minor issues Namely, the following problems are addressed:

- `String.ends-with?` did not actually work. It did happen on a small range of inputs (namely if the pattern string was exactly half of the length of the string to match), and thus the test worked. I fixed that and added a regression test for the future.
- `String.from-chars` should use `memmove` instead of `snprintf`. Added a test case for `from-chars` as well.
- The arithmetic macros `+`, `*`, `-`, and `/` should accept all numerical types, not just integers.

Cheers

### DIFF
--- a/core/String.carp
+++ b/core/String.carp
@@ -93,7 +93,7 @@
     (= sub &(prefix-string s (count sub))))
 
   (defn ends-with? [s sub]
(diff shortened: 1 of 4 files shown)


eriksvedang:
Great!

PR #160, opened 2018-01-09
[DO NOT MERGE] Make range a regular function

hellerve:
This PR rewrites range to be a regular function. This solves a todo in `docs/Todo.md`.

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -120,7 +120,6 @@ arrayModule = Env { envBindings = bindings, envParent = Nothing, envModuleName =
                                 , templateCopyingMap
                                 , templateEMap
                                 , templateFilter
-                                , templateRange
                                 , templateRaw
                                 , templateAset
                                 , templateAsetBang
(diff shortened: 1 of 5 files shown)


eriksvedang:
Great, so nice to get rid of these things in the compiler!
I think the new implementation is a bit too prone to infinite loops though, for example this:
```(Array.range 1 100 2)``` since it "misses" the end condition.

hellerve:
Right. We’ll have to check whether `start` or `end` are smaller and then use the appropriate check.

hellerve:
`range`, too, could benefit from a way of statically allocating an array, by the way.

hellerve:
I added a commit that gets the correct function. However, that exposed a compiler bug. The compiler emits the type `Fn__int_int_bool`, but does not emit a definition of that type.

eriksvedang:
Pushed a fix now, hope it helps.

hellerve:
I rebased and ran the tests; it seems to work now! I added two regression test cases as well.

PR #161, opened 2018-01-10
Add test case for sort

hellerve:
This PR adds a test case for `sort`.

I also want to start a conversation around whether we should have a generic function `cmp` or make it an interface. A problem with the current implementation of `sort` is that `cmp` takes refs, and that might require us to copy things. There are a few different strategies with which we could avoid this; for instance, we could change `sort` not to work with refs. That would only defer the problem of copying, though. If we made `cmp` an interface, we would have to copy a lot of code, except in the places where we wouldn’t, with strings for example.

I’m not entirely sure how to go about this.

Cheers

### DIFF
--- a/test/array.carp
+++ b/test/array.carp
@@ -12,6 +12,11 @@
 
 (defn inc-ref [x] (+ @x 1))
 
+(defn cmp [a b]
+  (if (= @a @b)
+    0
+    (if (> @a @b) 1 -1)))
+
 (defn main []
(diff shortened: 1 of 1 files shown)


eriksvedang:
Yes, that's a very interesting discussion! Should we do that here or on gitter?

hellerve:
Let’s move it over to Gitter, shall we?

PR #169, opened 2018-01-17
Print macro error (warning) when using more then one expression in body

Nananas:
Running this:
```
(use IO)

(defn main []
	(for [x 0 10]
		(println "test")
		(println "other")))
```
Prints lots of (10) `test`s, but **no** `other`: the rest of the for "body" is ignored in the macro.

I feel that `for` should be a `do` by default, though.

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -44,21 +44,23 @@
   (cond-internal xs))
 
 (defmacro for [settings body] ;; settings = variable, from, to, <step>
-  (list
-   'let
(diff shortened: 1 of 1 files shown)


eriksvedang:
Why did you close it?

Nananas:
Still fixing some corner-case bugs

Nananas:
I found a bug when having an empty for body.

eriksvedang:
Very nice!

PR #172, opened 2018-01-22
Fix size of stringification buffers

hellerve:
This PR introduces some logic to calculate the buffer length needed for the stringification of everything. This is most impactful on Arrays and Deftypes, but has an impact on all primitives. This should both reduce the brittleness of Array and Deftype stringification, and reduce the memory footprint of Carp as a whole, since no wasteful bytes are allocated anymore.

Cheers

### DIFF
--- a/core/core.h
+++ b/core/core.h
@@ -162,7 +162,7 @@ void IO_print(string *s) { printf("%s", *s); }
 
 string IO_get_MINUS_line() {
     size_t size = 1024;
-    char *buffer = CARP_MALLOC(size);
+    string buffer = CARP_MALLOC(size);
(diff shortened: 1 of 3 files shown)


eriksvedang:
First of all – this PR is awesome, I was scared of fixing the string allocations so thanks a lot for working through that!

Some minor feedback:

* The formatting of the template generated code could be a little nicer, with proper indentation and some newlines removed.
* There is an assertion lingering in the array string function (line 444) that can be removed now.

I'll make sure to add memory tests for the string functions as soon as I can, for now it seems like you got everything right – fingers crossed :)

PR #173, opened 2018-01-23
Int.clamp/pow, random fix, Array.initialize and symbol?

Nananas:
Added clamp to Int too. Fixed random-between floating point exception when lower = upper.
Added Array.initialize, symbol? commands.

Also added pow for integers. This does not use math's `pow` function, as it can be **very** slow in certain cases.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -117,6 +117,7 @@ arrayModule = Env { envBindings = bindings, envParent = Nothing, envModuleName =
   where bindings = Map.fromList [ templateNth
                                 , templateReplicate
                                 , templateRepeat
+                                , templateInitialize
                                 , templateCopyingMap
                                 , templateEMap
                                 , templateFilter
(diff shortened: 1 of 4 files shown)


eriksvedang:
Neat, what exactly does Array.initialize do?

Nananas:
like `repeat` but with the index as argument.

eriksvedang:
Ah, thank you!

hellerve:
I don’t particularly like the name `initialize`. I know this is a stone/glasshouse situation, because I suck at naming, but `initialize` isn’t completely intuitive for me. Any reasons it’s called that? Are there other languages that do that?

hellerve:
Also, it’s very similar to a constrained `range`, i.e. `(range 0 n)`.

Nananas:
@hellerve as far as I know, OCaml with `init` and Elm `initialize`. 
Any suggestions?

hellerve:
Ah, I though of `replicate`, not `repeat`. This makes more sense!

Nananas:
I renamed `initialize` to `repeat-indexed`. Naming is not necessarily final though.

I also made `cons` work for lisp arrays.

PR #174, opened 2018-01-24
Miscellaneous fixes

hellerve:
This PR fixes an issue in the code generated by Deftypes that would occasionally lead to compiler warnings because of unused variables and adds an import of the Statistics module to the Bench module.

Cheers


hellerve:
I also snuck in `Bench.set-min-runs!`.

Issue #125, opened 2017-12-12
Rewrite more of the template functions using plain Carp.


eriksvedang:
Now when we have interfaces it should be possible to rewrite some (most) of the Array functions using plain carp. This will clean up the compiler and make it less likely that a memory error slips in through the template generated C code.

hellerve:
Which functions are left to be done? I guess `filter` (but it does a `realloc`)  and `sort` (but it uses `sizeof`). Thus I think we can close this for now?

eriksvedang:
Yes!

PR #175, opened 2018-01-24
Array allocation

hellerve:
This PR addresses #125 by rewriting `repeat`, `replicate`, and both flavors of `map` in Carp. It also introduces `allocate` to obtain a fixed size array.

Test cases are included.

Cheers


hellerve:
#174 should be fixed prior to this. If #173 is merged prior to this I will rewrite the array template contained therein into Carp as well.

PR #181, opened 2018-01-29
Add defcompose macro

hellerve:
This PR adds the `defcompose` macro. It enables us to define functions in terms of compositions.

A test case is included.

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -197,3 +197,11 @@
 
 (defmacro ignore [form]
   (let [_ form] ()))
+
+(defdynamic defcompose-internal [forms]
+  (if (= (count forms) 0)
+    'arg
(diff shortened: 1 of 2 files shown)


eriksvedang:
I would much rather see people use `defn` + `=>` than this. A proliferation of "def" forms seems like it will confuse newcomers and be an annoyance to anyone reading Carp code. Another good thing with `=>` is that it helps you remember the order the functions are applied in, something that becomes very confusing with this (and inline `comp` too). So I'll hold off on merge:ing this unless someone can convince me I'm wrong.

Regarding discussion in gitter channel – a curry macro seems like it would fill a more actual need. Not sure how hard it would be to implement though..?

hellerve:
I tend to agree, though I think there is a place for using syntax to simplify things. Maybe that place is not the standard library and rather an external package, though.

PR #189, opened 2018-02-23
Change globals to vars in test

hellerve:
This PR fixes a code smell introduced into Test before complex global variables were a thing. We used to need to create functions to work with complex data structures, even if they’re static. Now we can use a global for that.

Cheers


eriksvedang:
Nice! I'll wait with merge:ing this until the initialization order of globals takes interdependencies into account, right now it's pure luck that this code works, heh.

hellerve:
Oh, hah! Alright, I think this can wait for a while. 😸

PR #192, opened 2018-02-27
Add regex library

hellerve:
This PR introduces an initial API for regexes in Carp. It is an adaptation of the Lua regex system (I should probably think about proper attribution before we merge this; thoughts?).

Tests are included.

The following functions were added:
- `find : (Fn [&String &String] Int)`: finds the index of a regex inside another string
- `match : (Fn [&String &String] (Array String))`: returns the match groups in a match
- `global-match : (Fn [&String &String] (Array (Array String)))`:  returns the match groups of all matches
- `substitute : (Fn [&String &String &String Int] String)`: substitutes a pattern `n` times. Passing `-1` will result in substitution of every occurrence of the pattern.
- `matches? : (Fn [&String &String] Bool)`: will check whether there are any valid matches of the regex in the string
- `match-str : (Fn [&String &String] String)`: will return the matched string

All in a few hours worth of work! :tada: (this probably means that it’s buggy)

Cheers


RyanSquared:
please note that if you're porting mostly from the Lua version, while this one is _extremely_ lighter than what is a fully working regex implementation, this is a _pattern matching_ system, not regex.

if you'd instead like something more powerful, I'd suggest porting over [LPeg][1] - a commonly used library in Lua that has more advanced features and is, IMO, much more powerful and better than regex (while also having a [re][2] mode that lets you use regex-like syntax).

[1]: http://www.inf.puc-rio.br/~roberto/lpeg/
[2]: http://www.inf.puc-rio.br/~roberto/lpeg/re.html

hellerve:
@RyanSquared thanks for your input! Ideally I would like to have a PCRE-compatible library in the long run, but I believe that this is quite good for now. It is small and maintainable, and this is more important than completeness at the moment, since we’re a very small team who maintain and develop Carp nights and weekends.

So, while I appreciate that Lua’s pattern system is more simplistic than what you would like to have in the long run, this is by design.

RyanSquared:
> So, while I appreciate that Lua’s pattern system is more simplistic than what you would like to have in the long run, this is by design.

And I'm perfectly fine with that, but what you're saying is that this is a _regex_ library, and the Lua team has made very strong efforts in the past to point out that _this is not regex_. It's pattern matching. It's faster. It's more useful. But most importantly, even if it looks like regex, it's not.

hellerve:
You’re right, I should probably rename them to patterns at the moment. Thanks again for your input!

eriksvedang:
Very exciting! Here's some feedback:

* Should there be `delete` & `copy` functions for Pattern? Perhaps even `init` and `str`?
* Naming of the internal functions in `carp_pattern.h` might clash with user defined functions (i.e. "match"). Either we need to put them into a .c file and make them static, or (easier) prefix them with `Pattern_`, they still don't need to be registered in Carp.
* I noticed at least one use of malloc, make sure all calls use CARP_MALLOC / CARP_STRDUP / CARP_FREE (unless there's a very good reason not to, of course?)
* How should the function `isManaged` in Lookup.hs handle the PatternTy?

hellerve:
Thanks for the review! I think all thos points should now be fixed :)

eriksvedang:
Epic!

Issue #195, opened 2018-03-09
Naming clashes


hellerve:
When defining variables that clash with typenames, we get spurious error from the C compiler, most notably when naming things `string` or `pattern`—which might be desirable names.

Minimum viable example:
```
(def string "hi")

(defn main []
  (IO.println string))
```

Please note that while `string` and `pattern` are most certainly the most notable examples, this also works with any other C type, e.g. `uint8_t`.

I guess this could be fixed by mangling variable names as well?

Cheers

eriksvedang:
Right, this is annoying. Changing the C names to `String` and `Pattern` seems like a good and simple start. Perhaps preventing identifiers from starting with a capital letter could be worthwhile. Or at least having a list of restricted words for symbols.

hellerve:
That sounds sensible enough for the moment!

hellerve:
Can this be closed as per #200?

PR #200, opened 2018-03-18
Rename string and pattern types

hellerve:
This PR addresses #195 by renaming `string` to `String` in C land and `pattern` to `Pattern`.

We should probably still change the parsing to require a lower-case character at the start, but I’m not totally sure how to best do that.

Cheers


eriksvedang:
Excellent! Yeah, not sure what to do about forcing lower case in variable names. It becomes a bit tricky when interop:ing ... but perhaps it's a good constraint on user defined variables in a carp program.

src/Parsing.hs
```diff
@@ -4,6 +4,7 @@ module Parsing (parse, validCharacters, balance) where
 
 import Text.Parsec ((<|>))
 import qualified Text.Parsec as Parsec
+import qualified Data.Char as Char
```
hellerve: I just realized I introduced this spurious import here, @eriksvedang! Sorry about that!

eriksvedang: Its not needed?

hellerve: No, I added it when trying to patch the parser for non-capitalized names (did not work out, though, for obvious reasons).

eriksvedang: OK, I’ll remove it then, cheers.

Issue #208, opened 2018-03-29
use of undeclared identifier 'CLOCK_REALTIME'


JoelSanchez:
I did this:
```
$ carp
Welcome to Carp 0.2.0
This is free software with ABSOLUTELY NO WARRANTY.
Evaluate (help) for more information.
> (load "Bench.carp")
> (use Bench)
> :cb
```
This happened (output ellided):
```
In file included from ./out//main.c:1:
.//core/carp_bench.h:5:3: warning: implicit declaration of function 'clock_gettime' is invalid in C99 [-Wimplicit-function-declaration]
  clock_gettime(CLOCK_REALTIME, &tv);
  ^
.//core/carp_bench.h:5:17: error: use of undeclared identifier 'CLOCK_REALTIME'
  clock_gettime(CLOCK_REALTIME, &tv);
                ^
1 warning and 1 error generated.
carp: callCommand: clang -fPIC -lm ./out//main.c -shared -o "./out//Untitled"   -I.//core/  -g  (exit 1): failed
```
I'm on ArchLinux, kernel version 4.15.12-1-ARCH, clang 6.0.0 (target: x86_64-pc-linux-gnu)

Tests pass, btw.
Same issue when running `benchmarks.sh`:

```
joel (LinerunnerGPM) ~/Desarrollo/self/carp git:master o % ./benchmarks.sh 
Copying from /home/joel/Desarrollo/self/carp/.stack-work/install/x86_64-linux-tinfo6-nopie/lts-8.12/8.0.2/bin/carp to /home/joel/.local/bin/carp
Copying from /home/joel/Desarrollo/self/carp/.stack-work/install/x86_64-linux-tinfo6-nopie/lts-8.12/8.0.2/bin/carp-header-parse to /home/joel/.local/bin/carp-header-parse

Copied executables to /home/joel/.local/bin:
- carp
- carp-header-parse
./bench/arithmetic.carp
In file included from ./out//main.c:1:
.//core/carp_bench.h:5:3: warning: implicit declaration of function 'clock_gettime' is invalid in C99 [-Wimplicit-function-declaration]
  clock_gettime(CLOCK_REALTIME, &tv);
  ^
.//core/carp_bench.h:5:17: error: use of undeclared identifier 'CLOCK_REALTIME'
  clock_gettime(CLOCK_REALTIME, &tv);
                ^
1 warning and 1 error generated.
[...]

PR #210, opened 2018-03-30
Adding time.h header to carp_bench.h to resolve bug #208

mkfifo:
Under POSIX clock_gettime and timespec are defined in time.h
I don't think we need sys/time.h

Tested under Linux, not tested under OSX


mkfifo:
Closing this in favour of #211

PR #211, opened 2018-03-30
Adding time.h header to carp_bench.h to resolve bug #208

mkfifo:
Under POSIX clock_gettime and timespec are defined in time.h
I don't think we need sys/time.h

### DIFF
--- a/core/carp_bench.h
+++ b/core/carp_bench.h
@@ -1,4 +1,4 @@
-#include <sys/time.h>
+#include <time.h>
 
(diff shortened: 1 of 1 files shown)


hellerve:
This seems fine to me on OS X, where I initially wrote the library, so :+1: from me!

mkfifo:
@hellerve Thanks for confirming!

I also just tested on Travis-CI against OSX and Linux and is looks good \o/
https://travis-ci.org/mkfifo/Carp/builds/361019149

PR #215, opened 2018-04-02
Travis will now also run ./run_carp_tests.sh

mkfifo:


mkfifo:
DO NOT MERGE

SDL is not currently enabled on Travis, so this will cause that test to fail
https://travis-ci.org/mkfifo/Carp/jobs/361022320
https://travis-ci.org/mkfifo/Carp/builds/361022317

PR #218, opened 2018-04-03
Add System.args

hellerve:
**This should be considered a WIP/RFC.**

This PR adds `System.args`, which is a Carp version of C’s `argc`/`argv` pair.

The system with which these are injected are pretty shoddy and rely on handrolled C, because I wasn’t entirely sure how to reuse the array templates here. It also relies on `Array_delete__String`, but doesn’t require it (again because I’m not sure how to do that when we’re emitting C).

Another open question for now: this requires us to readd a `malloc` dependency in `main`. We explicitly removed this a while ago to make it possible to rewrite a bunch of `core` for embedded systems if desired. How can we make sure this doesn’t break then? We could wrap it in an `ifdef` or such...

Cheers

### DIFF
--- a/core/System.carp
+++ b/core/System.carp
@@ -7,6 +7,7 @@
   (register sleep-seconds (Fn [Int] ()))
   (register sleep-micros (Fn [Int] ()))
   (register system (Fn [&String] ()))
+  (register args (Array String))
 )
 
 (defmodule Int
--- a/core/System.carp
+++ b/core/System.carp
@@ -7,7 +7,8 @@
   (register sleep-seconds (Fn [Int] ()))
   (register sleep-micros (Fn [Int] ()))
   (register system (Fn [&String] ()))
-  (register args (Array String))
+  (register get-arg (Fn [Int] (Ref String)))
+  (register get-args-len (Fn [] Int))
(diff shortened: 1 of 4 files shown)


eriksvedang:
Right, I didn’t think about the Array depending on bringing the full Core libs in. Perhaps a nicer solution is to create a function – more in the vein of the solution you first suggested on Gitter – that can access an arg by position. And store argc/argv in two globals. On top of this it’s easy to implement a helper function that fetches the whole thing as an Array, for programs that imports Array. This will clean up the deleter deps also. What do you think?

hellerve:
The problem is that we still need to store the arguments somewhere, which also means that we have to `malloc` space for them. Alternatively, we could also just reuse the space allocated for them already and skip deleting them, but I’m not sure whether this has any weird implications concerning the borrow checker?

eriksvedang:
Yes, I meant storing the argv pointer in a global and never handle it with the memory management system. The function accessing an arg at a specific position would return a ref.

hellerve:
This sounds good to me!

hellerve:
Updated to new API. This in fact made it simpler.

Issue #216, opened 2018-04-02
Tests should also check output


mkfifo:
Ideally we should have a suite of tests where each is a pair of `foo.carp` (input program) and `foo.out.expected` (expected stdout output) to be compared against

we can then have these run within each Travis test, so we can detect unexpected ouput changes

hellerve:
How would this complement the unit tests? Would those be for testing the basic language features with `examples`?

The unit tests themselves do check the output of various things (like stringification) already. I’m not saying that this is thus invalid, I just want to get a feel of what the use case is.

mkfifo:
Generally unit tests test a unit in isolation, this is often a single function/feature, testing input against output over a set of cases. For this I was thinking more integration or system tests.

I was thinking of including a set of decent sized programs and testing these against their expected output as a whole, a good start would be examples/*.carp and test/*.carp.

The idea of these is to form a set of larger tests which try to capture a large cunk of the Carp sytem and runtime in order to capture regressions.

These also begin to form some kind of documentation about the system.



For my toy language Icarus I had:

* 'unit tests' [1] written in C (the host language) to test individual units of the language implementation
* 'milestone' tests [2] each covering one new feature more than the previous run and trying to test this feature end-to-end in isolation
* 'custom' tests [3] [4] [5] [6] which try to capture a wide array of small programs

both 'milestones' and my 'custom' tests are testing the output, some of my custom tests also test the various intermediate forms and compilation targets (AST, IR [7], Bytecode [8], C [9], ...).

These tests meant that whenever I had a regression I was able to quickly narrow it down to the exact phase it happened in, as well as giving me a large set of examples of where it fails/pases.

These kinds of tests only make sense if your language syntax and behaviour is reasonably fixed, adding new features is possible, but large-scale changes means you have lots of tests to also update.

[1] https://github.com/mkfifo/icarus/tree/master/t/unit 
[2] https://github.com/mkfifo/icarus/tree/master/t/milestones
[3] https://github.com/mkfifo/icarus/tree/master/t/custom
[4] https://github.com/mkfifo/icarus/blob/master/t/custom/test_generics_progress.pl
[5] https://github.com/mkfifo/icarus/blob/master/t/custom/test_backend_pancake_end_to_end_output_only.pl
[6] https://github.com/mkfifo/icarus/blob/master/t/custom/test_backend_2c_end_to_end_output_only.pl
[...]

PR #220, opened 2018-04-10
Add test script recording and checking

hellerve:
This PR introduces the system described in #216, pretty much as @mkfifo described.

*It also fixes a bug in `System.args` and `noCore`, in that the variable `System.args` is not available when we do not use `core`. This is reasonable I think(?). The tests were broken otherwise.*

The script `test/record.sh` takes in a file path, compiles and runs the file, and records the output in an appropriate file. The other script `test/check.sh` will do the same and diff the outputs. If it fails, it prints an error message and returns 1.

Cheers


eriksvedang:
Looks good, though it's a bit hard for me to judge since it fails immediately with:
```
./examples/basics.carp failed.
```
I think the error reporting must be improved.

hellerve:
Whoops. This is bad. Usually it should print the diff between the two before exiting.

eriksvedang:
One more thing, the .actual files should be ignored by git, right?

hellerve:
Actually, if the test doesn’t fail, it should remove the `.actual` files, so them remaining in the repo is a smell that something went wrong. Failing tests don’t remove their output, because you might want to inspect them for debugging. But once all tests run fine, you shouldn’t have any of those files anymore.

I’ll probably not be able to fix your comments very soon (as tomorrow I’m 💍 )! If anyone wants to go ahead and take this up, that’d be great. Otherwise I’ll be available again in a while.

eriksvedang:
Ah, I see – that sounds good.

PR #223, opened 2018-04-14
fix benchmarks on MacOS

wbbradley:


hellerve:
This seems to be a problem before MacOS 10.12? What is your OS version?

(I’m on High Sierra myself)

eriksvedang:
Will close this for now, thanks for the patch though.

PR #225, opened 2018-05-07
Add neg protocol

hellerve:
This PR adds the `neg` protocol. It provides implementations for `Int` and `Float`.

Test cases are included.

Cheers


mkfifo:
Note: I see it adds implementations and tests for `Int` and `Long`, but I don't see any for `Float`.

mkfifo:
@hellerve I was commenting how your first comment above says "This PR adds the neg protocol. It provides implementations for Int and Float."

hellerve:
Oh right, sorry. I meant `Int` and `Long`, of course

PR #226, opened 2018-05-12
Add docs

hellerve:
This PR adds documentation to some functions. Not to all of them, sadly, but quite a few.

It also marks some functions as private.

Cheers


hellerve:
This seems to break the test `init_globals` with undefined symbols. This is weird because none of the functions were changed.

eriksvedang:
Yes, that's strange. I'll try to solve it – thanks a lot for adding the docstrings!

core/Bench.carp
```diff
@@ -5,38 +5,51 @@
 
 (defmodule Bench
   (def min-runs 50)
+  (private min-runs)
```
mkfifo: nitpick: move these private/hidden before the def to be consistent

core/Debug.carp
```diff
@@ -28,16 +29,17 @@
...
     (do
       (IO.println &(str &x))
       x))
 
-  ;; Leak some memory. Useful for testing tools that detect leaks.
+  (doc leak-array "Leak some memory. Useful for testing tools that detect leaks.")
   (register leak-array (Fn [a] ()) "Debug_leak_MINUS_array")
 
   )
+)
```
mkfifo: did you mean to add this bracket ?

core/Bench.carp
hellerve: I cannot, since those are variables and the compiler will error otherwise. I’ll gladly do that when the compiler is fixed.

core/Debug.carp
hellerve: No, I didn’t. Wow, this was the error that made the compiler throw an opaque error message. Thank you for reading through this diff!

hellerve:
@eriksvedang thanks to @mkfifo’s help, the error is fixed. It occurred due to a stray closing parenthesis, though, so it might be worth investigating throwing a better error message.

core/Bench.carp
mkfifo: Fair enough :)

core/Debug.carp
mkfifo: No problem, I saw your reported issue above so decided it was worthwhile just in case - I got lucky :)

eriksvedang:
Thanks a lot for figuring it out! Any ideas for how to give a better error message in this case?

hellerve:
The problem is that it doesn’t complain about unconsumed closing parentheses at the end, which might be a Parsec configuration thing (forcing it to consume everything). I remember having the same problem in the zepto parser, but cannot remember if and how I resolved it.

Issue #217, opened 2018-04-02
Travis-ci should invoke run_carp_tests


mkfifo:
First attempted at
https://github.com/carp-lang/Carp/pull/215

The problem is that our tests use SDL, which I haven't yet setup on Travis

'''
Already visited f__void
/bin/sh: sdl2-config: command not found
./out//main.c:1:10: fatal error: 'SDL2/SDL_image.h' file not found
#include <SDL2/SDL_image.h>
         ^
1 error generated.
carp: callCommand: clang -fPIC -lm ./out//main.c -o "./out//Ant"  `sdl2-config --cflags --libs` -D_THREAD_SAFE -fsanitize=address  -lSDL2_image -I.//core/  -g  (exit 1): failed
''''

https://travis-ci.org/mkfifo/Carp/jobs/361022320
https://travis-ci.org/mkfifo/Carp/builds/361022317

mkfifo:
PR #230 includes the work to resolve this

PR #229, opened 2018-05-19
Changing run_carp_tests.sh and test/check.sh to use `stack exec carp`

mkfifo:
rather than relying on stack install and ~/.local/bin being in path

This is a step towards issue #217


mkfifo:
This has been merged into #230

PR #230, opened 2018-05-19
Travis running run_carp_tests.sh

mkfifo:
This merges in two other PRs #229 and 228


mkfifo:
This is currently passing on Travis OSX and failing on Travis Linux due to differences in trailing newlines.

It is working on my Linux workstation though.

https://travis-ci.org/mkfifo/Carp/jobs/380981853#L568

mkfifo:
Just testing why this is, do not land yet

mkfifo:
Okay this is now passing on
https://travis-ci.org/mkfifo/Carp/builds/380985710

The underlying issue is due to a change in travis runtime environment which they are not prioritising a fix for
https://github.com/travis-ci/travis-ci/issues/9033

I had to switch our linux environments to use the older travis infrastructure, so our linux test times have gone from 1 minute to 12 minutes.

The only other option I can see is to disable our use of leak sanitizer, but I think 12 minutes is 'okay'.

This is now ready to land if you are happy with it,.

eriksvedang:
What was needed from the older travis infrastructure?

mkfifo:
@eriksvedang the older linux travis infrastructure allows us to use leak sanitizer, the newer linux travis infrastructure under linux won't let it run at all (due to the recent travis-ci/travis-ci#9033)

eriksvedang:
I'll wait with merge:ing this until we have the new Random module so that we can keep those examples.

mkfifo:
@eriksvedang we can merge this in so we get better travis testing now, and then add those tests back in easily enough

eriksvedang:
Fair enough, I'll add a comment in `updating.carp` about brining back the old code later.

PR #232, opened 2018-05-19
String library improvements

mkfifo:
I have some more I want to do, but this is probably a decent chunk to land in one go.

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -166,13 +166,28 @@
         (aset-uninitialized! &na i (f (nth a i))))
       na))
 
+  (doc sum-count "Returns the sum of counts from an Array of Arrays.")
+  (defn sum-count [xs]
+    (let-do [sum 0
+             lxs (Array.count xs)]
+      (for [i 0 lxs]
+        (set! sum (+ sum (Array.count (Array.nth xs i)))))
+      sum))
+
+  (doc concat "Returns a new Array which is the concatenation of the provided `xs`.")
   (defn concat [xs]
-    (let-do [result []
-             len (Array.count xs)]
-      (for [i 0 len]
-        (let [ys (nth xs i)
-              inner-len (Array.count ys)]
-          (for [j 0 inner-len]
-            (set! result (Array.push-back result @(nth ys j))))))
+    ;; This is using a StringBuilder pattern to only perform one allocation and
+    ;; to only copy each of the incoming Array(s) once.
+    ;; This currently performs wasted Array.count calls, as we call it for each
+    ;; Array once here and once in sum-count.
+    (let-do [j 0
+             lxs (Array.count xs)
+             result (Array.allocate (sum-count xs))]
+      (for [i 0 lxs]
+        (let-do [arr (Array.nth xs i)
+                 len (Array.count arr)]
+          (for [k 0 len]
+            (aset-uninitialized! &result (+ j k) @(Array.nth arr k)))
+          (set! j (+ j len))))
       result))
 )
--- a/core/carp_pattern.h
+++ b/core/carp_pattern.h
@@ -584,11 +584,11 @@ String Pattern_internal_add_value(PatternMatchState *ms, String res, String src,
         }
         res = Pattern_internal_add_char(res, tr[i]);
       }
-      else if (tr[i] == '0') res = String_append(res, src);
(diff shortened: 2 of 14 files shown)


mkfifo:
I have added a change which renames String.join -> String.concat, and String.join-with -> String.join

As this better matches other languages and libraries (python, perl, s-el,)
Please take a look and let me know if you are unhappy.

mkfifo:
The tests are passing on my workstation, Linux travis, and Mac OSX travis

I merged this with my travis branch for testing and it runs cleanly \o/

https://travis-ci.org/mkfifo/Carp/builds/381241262

mkfifo:
String.concat, String.join, String.join-with-char, and Array.concat are now all using a 'StringBuilder' pattern to minimise copies and allocate calls

mkfifo:
I have tried to keep this somewhat conservative, if you aren't happy with any part please let me know and I can remove it from this PR.

core/carp_string.h
```diff
@@ -195,3 +265,24 @@ String Long_format(String* str, long x) {
 long Long_from_MINUS_string(String *s) {
     return atol(*s);
 }
+
+int String_index_MINUS_of_MINUS_from(String *s, char c, int i) {
```
hellerve: Couldn’t this be expressed in Carp using `String.chars` and then iterating over the array? It would have worse runtime characteristics, though...

mkfifo: I initially wrote it in Carp, but AFAIK Carp lacks an explicit 'return' statement.
This meant exiting from a for/while loop early in Carp (early return) wasn't as neat.

eriksvedang:
This looks great! One thought – do we really need `StringCopy.append`? Since `String.append` will cut down on the number of unnecessary copies it should always be preferred, I think?

core/carp_string.h
eriksvedang: There is `break` if that helps. We have some functions in Core that use that in similar cases.

core/Array.carp
```diff
@@ -166,13 +166,28 @@
         (aset-uninitialized! &na i (f (nth a i))))
       na))
 
+  (doc sum-count "Returns the sum of counts from an Array of Arrays.")
+  (defn sum-count [xs]
```
eriksvedang: This could also be expressed in a more succinct way using `copy-map` and `Array.sum`.

core/carp_pattern.h
```diff
@@ -584,11 +584,11 @@ String Pattern_internal_add_value(PatternMatchState *ms, String res, String src,
         }
         res = Pattern_internal_add_char(res, tr[i]);
       }
-      else if (tr[i] == '0') res = String_append(res, src);
+      else if (tr[i] == '0') res = StringCopy_append(res, src);
```
eriksvedang: @hellerve Should probably start using `String_append` here and free `res` and `src` explicitly?

mkfifo:
@eriksvedang I was thinking that StringCopy.append is useful as it matches the current String.append behaviour, but I think you are right in that we can just remove it if you want

core/carp_string.h
mkfifo: ahh okay, I was looking for 'return'
I want to avoid String.chars as IIRC it is a copy, but I can always use String.chat-at.

Can we land this C version and I will try follow up later with a Carp version ?

core/Array.carp
mkfifo: @eriksvedang I wanted to avoid the cost of the intermediate allocation of an array for the results

eriksvedang:
Yes, I think it will lead to confusion down the road, better to just have one. This change will require some removal of copy and addition of refs at the call sites of course. Let me know if you’re up for fixing it.

core/carp_pattern.h
hellerve: Probably a good idea.

mkfifo:
@eriksvedang I'm up for giving it a go, I'll have a chat with you on Gitter to make sure I understand your ideas.

Is it okay if we land this as is, or at least parts of it? As it has the potential to conflict a lot, so I don't want this PR staying open too long.

core/carp_string.h
eriksvedang: Sounds good!

mkfifo:
Thanks!

PR #237, opened 2018-05-20
Trivial: Making all test/*.carp assert bracket styling consistent

mkfifo:


eriksvedang:
Why use this unusual Lisp style here?

mkfifo:
@eriksvedang sure thing, I will go the other way then :)

Issue #236, opened 2018-05-20
String and Array library `count` should be called `length`


mkfifo:
Is it okay if I make this change?

PR #239, opened 2018-05-20
Renaming: String.count -> String.length, Array.count -> Array.length

mkfifo:
Issue #236

The next steps would be to change count-char to just count


eriksvedang:
Yes, this is a good idea! I'll merge it when it has no conflicts.

mkfifo:
@eriksvedang thanks! rebased now to remove conflicts, and all tests are passing \o/

It's so nice having all those tests.

eriksvedang:
Will merge when the new conflicts are solved!

mkfifo:
@eriksvedang thanks, all conflicts fixed. There isn't much we can do here, since we are touching the same areas in the same files across multiple PRs, it isn't a big problem.
I also included documentation changes as a separate commit.

PR #240, opened 2018-05-20
Split Char.prn and Char.str

hellerve:
This PR splits up `Char.prn` and `Char.str`. Formerly, `str` and `prn` had the same behaviour, like most stdlib types. This is not correct, since `prn` should prepend a slash where `str` shouldn’t.

Cheers


eriksvedang:
Yes, good fix! There's a conflict because of the docs now.

hellerve:
I rebased it!

eriksvedang:
Thanks!

Issue #235, opened 2018-05-20
Want testing functions for checking errors occur


mkfifo:
It would be nice to have some assert-fails which runs an expression and expects it to fail (e.g. it will run an assert and exit with non zero)

A use case for this would be for testing aset! where we want to check that it deals with bounds behaviour correctly

For string-set! I would like to be able to write

```
(assert-fails
    &(let-do [str @"abcd"]
        (string-set! &str 4 \q)) ;; this should fail and throw an assert
    "string-set! correctly detects out of bounds writes"
)
````

**Comment by hellerve** [comment]
The problem is that C `assert`s abort the process on failure, which means we cannot really do anything with them. Maybe if we provided our own definition of `assert` that overrides the one provided by the system we could hack something up?

**Comment by mkfifo** [comment]
The libcheck testing library runs each test in a new process so that it can test for various forms of exit (assert, exit, segfault, ...).

If we only want to test assert, we could try swapping it out, although it might be cleaner to use CARP_assert and leave the stdlib assert untouched.

**Comment by hellerve** [comment]
Right. I might be able to work on this sometime this week; it would be nice if we could tackle concurrency at some point anyway.

**Comment by hellerve** [comment]
I created a function inside `Test` that looks like this:

```
(load "Test.carp")

(use Test)

(defn die []
  (System.exit 1))

(defn main []
  (with-test test
    (assert-fails test
                  1
                  die
                  "can exit with return code 1"
    )
    (print-test-results test)))
```

So if you wanted to test for a failed `assert`, you’d have to check for `SIGABRT`. Would that be to your liking?

hellerve:
Ecch, actually this is a signal, and as such has nothing to do with the exit code. We will have to catch the signals as well...

eriksvedang:
I think it's OK to close it. There are probably a hole bunch of other things that will require porting to Windows, so this case isn't that special.

mkfifo:
@hellerve thank you!

PR #244, opened 2018-05-23
Test for signals and exit stati

hellerve:
This PR introduces two functions into the test suite, called `Test.assert-exit` and `Test.assert-signal` that assert that a given function exits with a given status code or creates a certain signal, respectively.

This is useful for testing error conditions.

To make this work a few functions had to be added to the `System` module—but none to C—, namely `fork`, `wait`, and `signal`. A macro for getting the exit status of a child process called `WEXITSTATUS` had to be wrapped (we call it `get-exit-status`). So did the various signals; in Carp they were renamed from `SIG*` to `signal-*` to make the interface a little more idiomatic.

Cheers

### DIFF
--- a/core/System.carp
+++ b/core/System.carp
@@ -15,5 +15,8 @@
   (register get-arg (Fn [Int] (Ref String)))
   (doc get-arg "Gets the number of command line arguments.")
   (register get-args-len (Fn [] Int))
+  (register fork (Fn [] Int) "fork")
+  (register wait (Fn [(Ptr Int)] Int) "wait")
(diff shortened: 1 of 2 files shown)


hellerve:
This PR fixes #235.

eriksvedang:
Awesome. Since a not too far off goal is to make Carp run on Windows, some (most?) of this will have to be wrapped and/or conditionally compiled I believe. But until then we're all in happy Unix land, so this is really useful! And the new assert macros should stay the same even on Windows, so I guess it's fine to start using them in the test suite.

mkfifo:
@hellerve thank you!

PR #245, opened 2018-06-07
Core: add to-bytes

hellerve:
This PR adds a function `to-bytes` that returns the bytes for floating point values. This is useful for hashing, and maybe for other kinds of applications that need to fiddle with the bits within a float or double. It’s different from `to-int` et al. because it actually `memcpy`s the value into an integral value, preserving binary representation.

Cheers


core/carp_double.h
```diff
@@ -22,6 +22,12 @@ double Double_from_MINUS_int(int x) {
     return (double)x;
 }
 
+long Double_to_MINUS_bytes(double x) {
+    long y;
```
mkfifo: I think a `long` is only guaranteed to be at least as large as an integer, so it could be 32 bits (although it happens to be 8 bytes on my sysem).

A double is usually 64 bits, so here you should use a `long long` which is also at least 64 bits.

You also probably want this to be unsigned, so `long long unsigned y`.

The memcpy is a bit danerous too, as it is using the size of the from data type and not the to data type, but that should be fine once you convert to `long long`.

It would be even safer to use the fixed size integer types `uint6_4t` for `Double` and `uint32_t` for `Float`.

core/carp_float.h
```diff
@@ -21,6 +21,12 @@ float Float_from_MINUS_int(int x) {
     return (float)x;
 }
 
+int Float_to_MINUS_bytes(float x) {
+    int y;
```
mkfifo: this should also probably be unsigned

hellerve: I don’t think that’s necessary, and we don’t actually have unsigned types in the core library yet.

core/carp_double.h
hellerve: You’re right, of course. The standard C types are woefully underspecified, but most platforms use IEEE754 types, which are 32 and 64 bits. We should also rewrite converting floats to ints, because they’re not even guaranteed to be 32 bits (only 16, in fact).

I wrote a wrapper for `stdint`, but it’s not part of core, so I don’t think we want to use these types.

As for unsigned types, we don’t currently support them, and all of the functions expect signed types. We would thus lose precision passing it around. In the long run, having unsigned versions of all types is probably a good idea.

PR #251, opened 2018-06-15
Document string, remove string-join

hellerve:
This PR documents some more of the string module. I also removed `string-join` and replaced all of its uses with `str*`.

Cheers


eriksvedang:
Very nice!

PR #254, opened 2018-06-23
Adding missing includes headers for POSIX wait and signals

mkfifo:
Fixing build & test on Linux.

These headers are required as part of POSIX, so it is surprising it ever worked on Mac.

This built and passed all tests on my Travis for Linux and Mac.


mkfifo:
https://travis-ci.org/mkfifo/Carp

Issue #256, opened 2018-06-23
Integer division incorrect in interpreter, correct in compiler


mkfifo:
The interpreter gives an incorrect answer of `2`
```
$ stack exec carp
> (/ 3 2)
2
```

The compiler gives a correct answer of `1`
```
$ cat t.carp 
(use IO)
(defn main [] (IO.println &(format "%d" (/ 3 2))))

$ stack exec carp -- t.carp -x
1
```

eriksvedang:
Nice catch!

PR #258, opened 2018-06-23
Issue #256 The interpreter should use integer division 

mkfifo:


mkfifo:
The test is for compiled code, yes, it was more to serve as documentation - eventually we should also run those same tests over the interpreter.

PR #260, opened 2018-06-23
adding Array.copy

mkfifo:


eriksvedang:
This function is already generated by the compiler. It's nicer to just have it in Carp code though. Let me investigate if we can indeed remove the compiler version...

mkfifo:
It was unclear from the Array.carp module if I could just use copy or not, since it didn't have a `register copy` in there like the other modules.

hellerve:
What’s the status of this?

hellerve:
I’m not sure I follow. Shouldn’t this work fine as long as `copy` is defined on the encased type?

eriksvedang:
Sorry for being unclear. The implementation here works and is good, it's just that I don't know if we should replace the current (built-in) implementation of Array.copy with this code (because of optimization concerns). Does that make any more sense?

hellerve:
Totally! I guess it’s time to benchmark this then :)

hellerve:
Generic array functions are generated by the compiler and thus a little different. [They even have their own module](https://github.com/carp-lang/Carp/blob/master/src/ArrayTemplates.hs). This is because arrays are generic and thus all the specialized functions have to be generated rather than just imported from C.

They are generated from templates on demand (after the specializer has figured out what’s needed).

All of this is because it’s the only primitive type that is generic. 

Aside: it also means that if you want to work with generic types and the C FFI together, you’re sadly out of luck at the moment.

eriksvedang:
Actually there is a fun little loophole with generic types in the C ffi where you can use a generic carp type i.e. `a` to call C functions expecting `void*`.

hellerve:
Oh is that so? Will it not try to specialize then? That seems handy (if a little hacky)!

PR #261, opened 2018-06-24
Adding Array.push-back! and Array.pop-back!, along with tests for all push/pop backs

mkfifo:


mkfifo:
This has been subsumed into PR #262

Issue #252, opened 2018-06-19
Implement `sort` in Carp


eriksvedang:
The current implementation of `sort` relies on qsort and doesn't play well with lambdas (which are currently being implemented). If anyone wants to write a native sorting function in Carp that'd be great!

mkfifo:
I might give this a go if I have time this weekend

mkfifo:
Heapsort added in PR #262

PR #262, opened 2018-06-24
Issue #252 Heapsort

mkfifo:
This adds
  Array.push-back!
  Array.pop-back!
  Heap modules including Heap (generic), MaxHeap, MinHeap, and HeapSort
  Sort module including in place sort and external sort

This adds tests for all of the above and
  Array.push-back
  Array.pop-back

This subumes the previous PR #261


eriksvedang:
Very nice! Code looks great. My one piece of feedback before merge:ing is to move the `(use Heap)` statement inside the `Sort` module to not effect global scope. Cheers

mkfifo:
@eriksvedang I don't think that use is needed, I was unsure, but testing shows it isn't needed - I presume because Heap is within Core

fixed.

mkfifo:
Please hold off on merging this I just want to tidy up the interface a little bit to more closely match the current sort

I'll let you know when I think it is ready to be merged

mkfifo:
so now we have 

sort! &arr -> unit (in place mutable)
sort arr -> arr (in place ownership)
sorted &aa -> arr (makes copy and then sorted in copy)

I still want to add some -with-cmp or -with-ord versions for custom sorts

mkfifo:
The -with-cmp and -with-ord versions will be much cleaner once we have closures, so I might leave them until then.

The only place this matters is if you want to sort an array with a non-default ordering, but that is awkward to do without closures anyway.

This is ready to merge

mkfifo:
Now testing sorting of arrays for strings/ints/floats/chars

eriksvedang:
Sorry for being obnoxious but do we really need the Sort module now? It won't help when we add `sort` functions for other types, right? I'm definitely happy to merge this as-is but now there will be code that uses Array.sort and some code that uses Sort.sort which seems unnecessarily messy. 

If you want to add  `sort`/`sorted` interfaces in Interfaces.carp already, that's also fine by me.

mkfifo:
@eriksvedang not obnoxious at all :)

PR #267, opened 2018-07-17
Fix multiple bugs: Array and Char

hellerve:
This PR fixes two issues, namely:

- Reduce overly generic version of `Array./=` that works for all types.
- Cast to `unsigned char` first when casting a character to an integer to prevent sign spillover.

Cheers


chrisosaurus:
Thanks for fixing my Array bug!

PR #268, opened 2018-07-17
Change docs rendering to show whole form

hellerve:
This changes the HTML rendering of `defn`, `defmacro`, and `defdynamic` forms to include the argument names. This makes documenting the arguments easier.

See for instance this before and after comparison. This is what we had before:

bildschirmfoto 2018-07-17 um 19 05 46

And this is what we have now:

bildschirmfoto 2018-07-17 um 19 05 30

You can immediately see how the references to the variable names make more sense that way.

Cheers


hellerve:
Please note that I also made sure that the links still work.

bjorn3:
Could you please put some newlines (for example after `</div>`) in the generated docs, so that doc diffs are a bit smaller when code is changed?

hellerve:
I looked at this, but didn’t see a way in `Lucid` to pretty-print HTML ([see also](https://github.com/chrisdone/lucid/issues/28)). I’d love this, though!

Do you know of a good, simple tool for pretty-printing HTML that we could just pipe into or something?

bjorn3:
`html5-tidy` seems to work:

```
tidy -i -output doc_out.html doc.html 2> /dev/null
```

Or modifying in place:

```
tidy -i -m doc.html 2> /dev/null
```

hellerve:
This looks promising, thanks so much!

I’m going to try to find a library that does this (so devs don’t have to install another binary), but if I don’t find anything, this looks like a great way to do this! :)

I will take care of this in another PR, though, if that’s alright.

bjorn3:
> I’m going to try to find a library that does this (so devs don’t have to install another binary)

I understand

> I will take care of this in another PR, though, if that’s alright.

No problem

eriksvedang:
The problem with arguments name has been bugging me too, so this is a great initiative to fix that!!! I think the current solution is a bit noisy though, especially since some functions (external ones) don't get the names and are thus rendered without parens. I understand that those are even harder to do correctly since we don't have the names for external arguments, but still, the whole doc seems quite a bit less readable now, doesn't it?

I would like to experiment a bit more with different ways of rendering this. One thing could be to put the full info (with argument names) below the title in the spot were it just says `defn` now, and leave the function names as headings. 

Another thing that might be enough is to add fake argument names to external functions (a, b, c...) and make it possible to set it as meta data.

All in all I think the names make it much easier to understand the functions and what they do, I just want to tweak the information design a tiny bit before merge:in. I'm up for doing this, but if you have the time @hellerve that's even better!

hellerve:
I can happily do this. The design that I chose is pretty much a carbon-copy of what I used in zepto. We could do something like Clojuredocs, maybe ([example](https://clojuredocs.org/clojure.core/count))?

eriksvedang:
Cool, please go ahead. Clojure way of doing the heading + parenthesised form with args looks good.

hellerve:
How’s that?

bildschirmfoto 2018-07-19 um 13 03 05

The design can be changed easily, of course.

hellerve:
Also note that this will only be shown for `defn`, `defmacro`, and `defdynamic`.

PR #269, opened 2018-07-18
docs: Change to pretty rendering

hellerve:
This PR replaces `lucid` by `blaze-html` to leverage its pretty printer. This leads to pretty HTML, which is a bit nicer to deal with. This is per request from @bjorn3 in #268.

It also means that this diff is super huge, sorry about that!

Cheers

### DIFF
--- a/src/RenderDocs.hs
+++ b/src/RenderDocs.hs
@@ -2,7 +2,10 @@
 
 module RenderDocs where
 
-import Lucid
+import Text.Blaze.Html5 as H
+import Text.Blaze.Html5.Attributes as A
+import Text.Blaze.Html.Renderer.Pretty (renderHtml)
+import Text.Blaze.Internal (stringValue)
 import Data.Text.Lazy as T
 import Data.Text.Lazy.Encoding as E
 import Data.Text as Text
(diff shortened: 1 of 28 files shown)


eriksvedang:
It writes a bunch of function names to stdout now..?

src/RenderDocs.hs
```diff
@@ -95,13 +102,13 @@ binderToHtml (Binder meta xobj) =
...
+  in  do H.div ! A.class_ "binder" $
+           do H.a ! A.class_ "anchor" ! A.href (stringValue ("#" ++ name)) $
+                do H.h3 ! A.id (stringValue name) $ (toHtml name)
+              H.div ! A.class_ "description" $ (toHtml description)
+              H.p  ! A.class_ "sig" $ (toHtml typeSignature)
               case maybeNameAndArgs of
-                Just nameAndArgs -> pre_ [class_ "args"] (toHtml nameAndArgs)
-                Nothing -> span_ [] (toHtml (""::String))
-              p_ [class_ "doc"] (toHtml docString)
+                Just nameAndArgs -> H.pre ! A.class_ "args" $ toHtml (trace nameAndArgs nameAndArgs)
```
bjorn3: > It writes a bunch of function names to stdout now..?

Could this be the source? (`trace`)

PR #270, opened 2018-07-27
Add tests for args

cammellos:
Thanks for the project!

This PR just adds a couple of tests for `args.carp`, it assumes they are run with:

`carp -x test/args.carp`

or through the provided script:

`./run_carp_tests.sh`

### DIFF
--- a/test/args.carp
+++ b/test/args.carp
@@ -1,3 +1,19 @@
+(load "Test.carp")
+(use Test)
+
+(defdynamic _make-exe-path [pth]
+  (String.join (array (Project.get-config "output-directory") pth)))
+
+(defmacro make-exe-path [pth]
+  (_make-exe-path pth))
+
 (defn main []
-  (for [i 0 (System.get-args-len)]
-    (IO.println (System.get-arg i))))
+  (with-test test
+    (assert-equal test
+                 1
+                 (System.get-args-len)
+                 "System.get-args-len works as expected")
+    (assert-equal test
+                  (make-exe-path "/Untitled")
+                  (System.get-arg 0)
+                  "System.get-arg works as expected")))


hellerve:
Thanks for this!

test/args.carp
```diff
@@ -1,3 +1,11 @@
...
+(use Test)
+
 (defn main []
-  (for [i 0 (System.get-args-len)]
-    (IO.println (System.get-arg i))))
+  (with-test test
+    (assert-true test
+                 (= 1 (System.get-args-len))
+                 "System.get-args-len works as expected")
+    (assert-true test
```
eriksvedang: This test fails for me... Also, it's better to use `assert-equal` in that case to get a better error message.

eriksvedang:
@cammellos Thanks!

hellerve:
@cammellos are you up to work on this still or should I take over from here? I’d totally make this PR mergeable if you’re okay with it; if you’d rather work on it yourself that’s great as well!

cammellos:
@hellerve sorry I missed the previous messages as I was on holiday, I can give it another try.

@eriksvedang what's the error message?

Thanks

eriksvedang:
```
Test 'System.get-arg works as expected' failed:
	Expected value: 'true', actual value: 'false'
```

hellerve:
Heya, I just submitted #292, which will make this simple and configurable. My working version of this would then be:

```
(load "Test.carp")
(use Test)


(defdynamic _make-exe-path [pth]
  (String.join (array (Project.get-config "output-directory") pth)))

(defmacro make-exe-path [pth]
  (_make-exe-path pth))

(defn main []
  (with-test test
    (assert-equal test
                 1
                 (System.get-args-len)
                 "System.get-args-len works as expected")
    (assert-equal test
                  (make-exe-path "/Untitled")
                  (System.get-arg 0)
                  "System.get-arg works as expected")))
```

Sorry for stealing your thunder, @cammellos, just wanted to help you get this show on the road!

cammellos:
@hellerve thanks! sorry for not being responsive but don't have much bandwidth at the moment (I also don't own a mac so can't test).
Thanks!

cammellos:
@eriksvedang I have incorporated @hellerve suggestion and should be working now, could you try again?
thanks

eriksvedang:
Oh excellent, thank you!

PR #271, opened 2018-08-06
Refactor colorization

hellerve:
ANSI colorization has historically been a mess. This is an attempt at cleaning things up a little bit by adding another file, another module, and separating string creation and IO action to increase reuseability.

Cheers


eriksvedang:
Nice! (it could use the Map now also, right..?)

PR #272, opened 2018-08-06
Miscellaneous fixes in core

hellerve:
This is a set of bugfixes that I found while developing the [diff](/hellerve/diff) package. Specifically, it:

- adds a `zero` function to `Array` that returns an empty array.
- Makes `Map.put` drop ownership of the map.
- Fixes the docstring for `Map.get`.
- Adds a custom, prettier `str` function to `Map`.
- Makes `Set.put` drop ownership of the set.
- Fixes the docstring for `Set.get`.
- Adds a custom, prettier `str` function to `Set`.
- Adds tests for stringification

Cheers


test/map.carp
```diff
@@ -78,12 +83,17 @@
...
                   "empty? works"
     )
     (assert-equal test
                   true
-                  (Set.empty? &(Set.remove (Set.put (Set.create) "1") "1"))
+                  (Set.empty? &(Set.remove (Set.put &(Set.create) "1") "1"))
                   "remove works"
     )
+    (assert-equal test
+                  "{ hi bye }"
```
eriksvedang: Shouldn't the strings have quotes here? (like with `str` for Array)

hellerve: Depends. `str` of `String` doesn’t do that, `prn` does IIRC.

eriksvedang: Yeah, String's the reason there are two different functions. It's certainly cleaner looking to have it like this but also confusing since you can't really see that those are strings and not variable names. Since structs and arrays do the other thing I think map should behave the same way.

hellerve: Makes sense!

PR #275, opened 2018-08-07
Headerparser: Parse function-like macros and multiline macros

hellerve:
This PR extends the parts of the header parser that deal with macros. It allows for function-like macros (i.e. macros that take arguments), and multi-line macros.

“Test” cases are included in `examples/parse_me.h`.

Cheers


eriksvedang:
Awesome!

PR #276, opened 2018-08-08
Escape chars in macro stringification

hellerve:
This PR introduces escaping of `\` in macro stringification, because otherwise the C compiler will complain about invalid escape sequences. For an example, try: `(assert (= \. \+))`.

Cheers


eriksvedang:
Thanks for fixing this!

PR #280, opened 2018-08-26
Simplify map

hellerve:
This PR adds a few changes to map that make it simpler and more performant:

- remove the `Entry` type, as it is equivalent to `Pair`
- add a `SetBucket` type to avoid carrying around useless values in hashsets
- Fix a few docstrings
- Add benchmarks

Cheers


eriksvedang:
Very nice!

PR #284, opened 2018-08-28
Fix discoloration on reset

### LINKED ISSUE #283: Starting carp from the terminal messes up the terminal colours
I tried invoking `carp` from my terminal (iTerm2 to be precise), and this is what happened:

screen shot 2018-08-28 at 6 38 52 pm

First of all, with my default White on Black profile, the content is almost illegible, and secondly, even after exiting the program, the colour profile still remained till some random commands such as `ls -la` cleared that up.

hellerve:
This PR fixes #283. The original reset in `strWithColors` was ANSI code `37`, which is `white`. It should’ve been `0`, which is `reset`.

Cheers

### DIFF
--- a/src/ColorText.hs
+++ b/src/ColorText.hs
@@ -5,9 +5,9 @@ import Util
 data TextColor = Blue | Red | Yellow | Green | White
 
 strWithColor :: TextColor -> String -> String
-strWithColor color str = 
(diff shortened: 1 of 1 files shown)


eriksvedang:
That makes a lot of sense!

PR #290, opened 2018-09-08
Print short-circuiting test results

croyzor:
When running the carp test suite, I noticed that the results were not being printed at the end of this test. I Implemented this change and the results printed upon re-running the tests.

### DIFF
--- a/test/short_circuiting.carp
+++ b/test/short_circuiting.carp
@@ -67,4 +67,5 @@
                   3
                   (vararg-or)
                   "'or*' works")
+    (print-test-results test)
     ))


hellerve:
Oh, yes, that’s an oversight! I should probably generalize the macro `with-test` to print the test results at the end anyway, but I’m not sure?

croyzor:
I think since `with-test` prints anyway, it can't hurt to add printing the results to it. I'm happy to take the reigns on that change. It should be a good issue to get my feet wet!

eriksvedang:
Thx!

PR #291, opened 2018-09-12
Print results as part of test macro

croyzor:
As discussed in #290, this change moves the printing of test results to the `with-test` macro. Additionally, I renamed the misspelled `test/safe_arithmetic` file.

Tested using the `run_carp_tests.sh` script and observing that the results are the same as before the change.

### DIFF
--- a/core/Test.carp
+++ b/core/Test.carp
@@ -104,9 +104,7 @@
         (IO.color "red")
         (IO.print "\tFailed: ")
         (IO.println &(Int.str failed))
-        (IO.color "reset")
-        (State.copy state))))
-)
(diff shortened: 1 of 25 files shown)


hellerve:
Thanks for this! It seems to work fine and definitely improves the signal-to-noise ratio.

You have changed a function’s signature (`print-test-results`), so when I run the test suite I get a little documentation change. I would add that to the PR (if there is no reason not to that I haven’t thought about?)!

croyzor:
Thanks for pointing that out! There's no reason not to update the documentation, it just hadn't occurred to me. I've updated the PR now

PR #292, opened 2018-09-13
add Project.get-config

hellerve:
This PR adds `Project.get-config`, the getter alternative to `Project.config`. This is there to simplify use cases like in #270 where you might want to fing out values like `output-directory` from within the program.

Test cases are included.

Cheers

PS: I noticed that a few properties of `Project` are missing in `Project.config`. WIll add those in a separate PR.


eriksvedang:
Handy!

PR #295, opened 2018-09-19
Add file, line, and column commands

hellerve:
This PR adds commands to get the `line`, `file`, and `column` info from a given `XObj`. This could be useful from within macros. I have something specific in mind, namely debug printing:

```clojure
(defmacro prn [x]
  (list 'let-do (array 'tmp x)
    (list 'IO.println (list 'ref (list 'fmt "%s:%d:%d: %s" (file x) (line x) (column x) '&(str tmp))))
    'tmp)
)
```

This would allow me to print an arbitrary object and its location automatically, so that I could do things like this:

```clojure
(if (= 1 2)
  ; do something
  ; do something else
)

(if (= (prn 1) 2) ; will work the same, but also print <file>:<line>:<column>: 1
  ; do something
  ; do something else
)
```

I hope this serves as a good motivating example why this might be useful!

Cheers

### DIFF
--- a/core/Debug.carp
+++ b/core/Debug.carp
@@ -30,10 +30,11 @@
                 ())))
 
   (doc trace "Print the value of an expression to stdout, then return its value.")
-  (defn trace [x]
-    (do
-      (IO.println &(str &x))
-      x))
+  (defmacro trace [x]
+    (list 'let-do (array 'tmp x)
+      (list 'IO.println (list 'ref (list 'fmt "%s:%d:%d: %s" (file x) (line x) (column x) '&(str tmp))))
+      'tmp)
+  )
 
   (doc leak-array "Leak some memory. Useful for testing tools that detect leaks.")
   (register leak-array (Fn [a] ()) "Debug_leak_MINUS_array")
--- a/core/Debug.carp
+++ b/core/Debug.carp
@@ -41,15 +41,10 @@
 
 )
 
-;; HACK! This silences compiler errors about 'source-location' and 'source-path' being undefined.
-;; Actual calls to these special forms will be handled directly in the dynamic evaluator, see 'Eval.hs'.
-(defdynamic source-location [] "")
-(defdynamic source-path [] "")
-
 ;; Crash the program with an error message unless the expression evaluates to 'true'.
 (defmacro assert [expr]
   (list 'if (list '= true expr)
         ()
         (list 'do
-              (list 'println* "Assertion '" (str expr) "' failed at " (list 'source-location))
+              (list 'println* (list 'fmt "Assertion '%s' failed at line %d, column %d in file %s" (str expr) (line) (column) (file)))
               '(System.exit 1))))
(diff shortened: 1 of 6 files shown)


eriksvedang:
This was an interesting PR! 

1. There are two dynamic functions named `source-location` and `source-path` (found in `Debug.carp`) that contains the same info, but in a less granular (and thereby worse) format than yours. To make those functions report the correct location I had to do a bunch of hacks, the reason being that they don't take any arguments which complicated things a great deal. So in that way I like your functions more. Having both seems a bit too much though, and I think the non-argument ones are more generally useful?! (not sure about that though) 

If you agree, we could use the existing functions but make them more granular (one for line, one for column) and make a compound macro that puts them together like `source-location` does now. If you don't agree – well then I'm up for having both if the motivation is good!

2. The example function above (`prn`) does basically what `Debug.trace` tries to do, but in a better way, perhaps you could replace it with your code and that could become the official tracing macro? Maybe put it at top level to have to avoid qualifying the name also...

hellerve:
Thanks for your thorough input! I hate that I don’t know all of the API here, and I’d like to avoid any duplication of functionality of course.

I looked at all of the functions you pointed me at.

I do think that if we end up putting these macros in the standard, we should replace `Debug.trace`, because my version is—in my opinion, anyway—superior.

On the issue of `source-location` and `source-path`. My issue with these functions is that they do not return the data in a usable format and instead just return strings, which might or might not be what I want in my macros generally. I think the more granular versions are thus superior. On the issue of zero vs one argument, I’d say that it would make sense if we add both, in other words make them variadic. Zero arguments would mean: `give me the location of the current statement`. One argument would mean: `give me the location of the argument I gave you`. Implementation-wise, his wouldn’t be much of an issue since they are currently part of the `eval` function anyway. I can do that fix.

Are you okay with both of these? If so, I’ll adjust the PR accordingly.

eriksvedang:
Lovely, that sounds perfect.

hellerve:
@eriksvedang Would you take another look and tell me whether this is more to your liking as well?

hellerve:
I just realized there was still a stray use of the old API, fixing...

hellerve:
Now everything should be fine!

eriksvedang:
Awesome, this is much better.

eriksvedang:
There's one problem with the trace macro still though (which was also a problem before) – it doesn't handle types that requires ref:ing very well, for that it should look like this:

```
(defmacro trace [x]
    (list 'let-do (array 'tmp x)
      (list 'IO.println (list 'ref (list 'fmt "%s:%d:%d: %s" (file x) (line x) (column x) '&(str &tmp))))
      'tmp)
```

Notice the '&' before `tmp`. Any ideas for how to solve this in a neat way?

eriksvedang:
Hmm, perhaps the best solution for now will be to have two trace-functions, one that expects refs and one that doesn't, since it varies what you "have" and sometimes you don't want to change it (since that's the purpose of just letting it pass through the macro unchanged. Unfortunately the current rule of not returning refs from `let` forms also gets in the way, but that's unsolvable until we have better lifetimes so that will just have to stay that way for now.

PR #296, opened 2018-09-19
Fix mode resolution for multisymbols

### LINKED ISSUE #288: Ambiguous module resolution causes erroneous name mangling
If the code uses two modules which have similarly named functions, it may be able to infer the correct one but the name mangling doesn't keep up.
This code has "init"-functions from two modules (Vector2, GLFW) and Carp infers that GLFW is the correct one but produces incorrect C-code:

```
(load "GLFW.carp")
(use GLFW)

(load "OpenGL.carp")
(load "Vector.carp")

(use Vector2)

(defn-do main []
  (let-do [_ (init)
           window (create-window 640 480 (cstr "GLFW <3 CARP") NULL NULL)]
    (make-context-current window))
  0)
```

The error message from C-compiler is 

```
./out//main.c:4591:18: warning: implicit declaration of
      function 'GLFW_init' is invalid in C99
      [-Wimplicit-function-declaration]
        int _7 = GLFW_init();
```

hellerve:
This PR fixes mode resolution for multisymbols during concretization, by which I mean it fixes #288.

A test case is included. It’s in a file called `text/regression.carp`, in which I intend to test for any regressions for fixes like this one.

Cheers

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -216,22 +216,24 @@ concretizeXObj allowAmbiguityRoot typeEnv rootEnv visitedDefinitions root =
     visitMultiSym allowAmbig env xobj@(XObj (MultiSym originalSymbolName paths) i t) =
       let Just actualType = t
           tys = map (typeFromPath env) paths
+          modes = map (modeFromPath env) paths
           tysToPathsDict = zip tys paths
-      in  case filter (matchingSignature actualType) tysToPathsDict of
+          tysPathsModes = zip3 tys paths modes
+      in  case filter (matchingSignature3 actualType) tysPathsModes of
             [] ->
               --if allowAmbiguity
               --then return (Right xobj)
               --else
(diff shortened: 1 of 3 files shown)


src/Concretize.hs
```diff
@@ -386,6 +392,25 @@ typeFromPath env p =
       | otherwise -> error "Local bindings shouldn't be ambiguous."
     Nothing -> error ("Couldn't find " ++ show p ++ " in env:\n" ++ prettyEnvironmentChain env)
 
+-- | Get the mode of a symbol at a given path.
```
eriksvedang: Perhaps add a comment about that this duplicates some functionality from the Qualify module, as a reminder in the future?

eriksvedang:
Looks good! I added a small comment, will merge asap.

src/Concretize.hs
hellerve: You’re right, I added a small comment pointing eager refactorers to the right spot!

eriksvedang: :D

PR #297, opened 2018-09-23
Proposal: make and and or macros

hellerve:
This PR removes the special case of `or` and `and` in the compiler and instead makes them macros. The code generated is more or less equivalent.

An annoying side-effect of this is that `and` and `or` within `Eval.hs` have to be renamed. At the moment I renamed them to `mand` and `mor`, but that’s not particularly great naming. I’m open for any suggestions!

Cheers

### DIFF
--- a/core/Interfaces.carp
+++ b/core/Interfaces.carp
@@ -1,3 +1,5 @@
+(load "Macros.carp")
+
 ;; The 'copy' and 'str' interfaces are defined internally:
 ;;(definterface copy (λ [&a] a))
 ;;(definterface str (λ [a] String))
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -242,6 +242,12 @@
 (defmacro case [name :rest forms]
   (case-internal name forms))
 
+(defmacro and [x y]
+  (list 'if x y false))
+
+(defmacro or [x y]
+  (list 'if x true y))
+
 (defdynamic build-vararg [func forms]
   (if (= (length forms) 0)
     (macro-error "vararg macro needs at least one argument")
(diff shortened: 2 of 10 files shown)


bjorn3:
+13 −103 :tada:

eriksvedang:
I like this a great deal better than what we had before implementation wise, but the ergonomics of having another name for 'and' and 'or' seems pretty bad. I was hoping that removing the cases from the evaluator and relying on the new macros for dynamic code would work, but for some reason it doesn't... We could leave this as-is (with the weird names) and hope that it's fixable in the future but it also seems like people will cut their fingers on it before then (plus its annoying to have to update your code for such a minor thing) so perhaps I need to ponder this a bit more before merge:ing.

hellerve:
While I don’t have anything substantial to add, I feel the same way.

In the end, what matters most to me is the ergonomics. If we can’t have that, then it might be worth dropping this allotgether, but I think this needs to be possible. We just don’t know how yet.

eriksvedang:
I haven't come up with a good solution for this, but one observation is that the dynamic `and` and `or` are used very little (actually not at all?). So having slightly bad names for them isn't a big deal maybe. I was thinking that maybe they could be named `Dynamic.and` and `Dynamic.or` (even though they are special forms). That would make it super obvious what they are for. With better macro support in dynamic code they could then be moved into the Dynamic module and no code that depend on them will have to change (and they can be removed from the evaluator).

hellerve:
Clarification: by the dynamic versions, you mean the ones in `Eval`, right?

eriksvedang:
Yeah. Sorry for being a bit uncoherent!

hellerve:
No problem at all. We don’t have a vocabulary yet, and we should probably think about the concrete terms associated with some of the concepts in Carp soonish.I usually make them up as I go, but that’s not super productive!

eriksvedang:
What do you think of the name change idea? I admit it won’t really matter much, and the real solution is better macro system, but I think I would like it a tiny bit more at least...

hellerve:
I can do that! Sorry, i wasn’t avalable the last few days because of work; I will either do it tomorrow or during the weekend!

hellerve:
Done!

PR #298, opened 2018-09-25
Use closures in core

hellerve:
This PR changes the core/stdlib to use closures instead of helper functions for `map` et al.

This uncovered a bug in emitting closures: there was a stray `.` in the name that was passed into the emitter (e.g. `_Lambda_VectorN.mag-sq_17`). This is probably due to `Emit.hs:188`, where we mangle any old name as if it was a single name. The simple fix for this was to add a new mangle rule that substitutes `.` by `_`. This is maybe a little hacky, but it works and is unintrusive. Maybe someone else has a better suggestion?

Cheers

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -57,7 +57,7 @@
 
   (doc minimum "Sum an array (elements must support + and zero).")
   (defn sum [xs]
-    (Array.reduce add-ref (zero) xs))
+    (Array.reduce (fn [x y] (+ @x @y)) (zero) xs))
 
   (doc subarray "Get subarray from start-index to end-index.")
   (defn subarray [xs start-index end-index]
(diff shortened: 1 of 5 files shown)


eriksvedang:
Right... I guess the Fn should carry with it a `Maybe Path` instead of a `Maybe String`... let me fix that before merge:ing this.

hellerve:
It also seems as if `typedef`s required by lambdas aren’t emitted. This bug is present elsewhere, and should probably be tackled elsewhere. Will open an issue as soon as I have a minimal example.

eriksvedang:
I pushed a potential fix for this now, can you confirm that it works even without your name mangling trick?

eriksvedang:
Looks like the mangling is still in there..?

hellerve:
No, that’s just me reformatting the code with a newline! I can remove that if you prefer to have it on one line. :)

eriksvedang:
Oh, I was fooled by the highlighting sorry!

PR #300, opened 2018-09-25
Add Char.upper-case and Char.lower-case

hellerve:
This PR adds the functions `Char.upper-case` and `Char.lower-case` to test whether a character is upper or lower case. It will only return `true` for characters that are actually alphabetical, which means that e.g. `!` is neither upper nor lower case.

Test cases are included.

Cheers

### DIFF
--- a/core/Char.carp
+++ b/core/Char.carp
@@ -25,6 +25,14 @@
 
   (defn /= [a b]
     (not (= (the Char a) b)))
+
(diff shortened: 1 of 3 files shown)


eriksvedang:
Cool. How about adding `?`:s to the end of the names, to differentiate them from functions that turn a char into upper/lower case?

hellerve:
Fair enough!

hellerve:
Done!

PR #308, opened 2018-10-17
Add more file IO functions

hellerve:
This PR adds a few more IO functions around files to IO. Maybe we should think about adding a `File` module to the standard library?

Cheers


hellerve:
Nota bene: I added more standard C functions around file descriptors to this!

eriksvedang:
Very nice!

PR #310, opened 2018-10-19
Compiler: only use colored errors on ttys

hellerve:
This PR adds a check for output coloring that restricts colors to TTY devices (i.e. terminals). This prevents files from being clobeered with ANSI escape codes.

One downside of this is that it depends on the `unix` package, which is kind of a bummer, but seems to be the only way.

Cheers

### DIFF
--- a/CarpHask.cabal
+++ b/CarpHask.cabal
@@ -55,6 +55,7 @@ library
                      , blaze-html
                      , blaze-markup
                      , text
+                     , unix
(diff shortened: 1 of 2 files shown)


eriksvedang:
The list of things to try/fix when I'm on a Windows machine grows yet again..!

eriksvedang:
Nice patch though :)

PR #312, opened 2018-10-22
Add suport for markdown in docstrings

hellerve:
This PR adds support for Markdown (more specifically, CommonMark) to docstrings in Carp. This should simplify generating rich, good documentation.

This is a proposal and should be treated as such. If people have strong feeling for or against this, I’d like to hear it and why!

Cheers


hellerve:
The conflict is resolved!

PR #319, opened 2018-10-31
Add performance-oriented patches

hellerve:
This PR adds `Map.put!`, `Map.remove!`, `Set.put!`, and `Set.remove!`. They avoid copying the map/set, but also take ownership over the data.

It also changes the type signature of `Array.reduce` to take ownership of the accumulator to reduce the need for copying there as well.

Test are included, examples are fixed.

Cheers

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -5,7 +5,7 @@
     (let [total x]
       (do
         (for [i 0 (length xs)]
-          (set! total (f &total (nth xs i))))
+          (set! total (f total (nth xs i))))
         total)))
 
   (doc first "Take the first element of an array.")
@@ -57,7 +57,7 @@
 
   (doc minimum "Sum an array (elements must support + and zero).")
   (defn sum [xs]
-    (Array.reduce (fn [x y] (+ @x @y)) (zero) xs))
+    (Array.reduce (fn [x y] (+ x @y)) (zero) xs))
 
   (doc subarray "Get subarray from start-index to end-index.")
   (defn subarray [xs start-index end-index]
--- a/core/Map.carp
+++ b/core/Map.carp
@@ -99,13 +99,19 @@
   (defn create-with-len [len]
     (init len (Array.repeat len Bucket.empty)))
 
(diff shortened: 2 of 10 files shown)


eriksvedang:
This is great, nice to see that it was fairly straight forward to optimise!

I'm confused by the naming though. The pattern for the Array functions should be followed as closely as possible. It has the following rules (perhaps those should actually be stated somewhere... sorry about that):

* Functions that behave "functionally" (returns a "new" version of their argument, usually by taking ownership and doing internal mutation) should be named without `!`
* Functions that mutate by taking a ref to the object should have a `!` at the end
* Functions that take refs and return sub-parts of their argument don't need `!` either

Both `put!` and `remove!` fall into the first category, I believe? (correct me if I'm missing something)

hellerve:
Yes, that’s correct. What names would you propose?

eriksvedang:
I also need to think a bit more about this new signature of `reduce` since I thought the old one was the most reasonable, but perhaps this is actually better...

eriksvedang:
Maybe just `put` and `remove`?

hellerve:
So should we throw out the versions that take refs?

hellerve:
The accumulator gets overriden in every iteration and invalidated. You return a new accumulator. The normal use case would be to work on the accumulator and return the new version. If you only have a ref, you always need to copy for no reason.

eriksvedang:
Are the versions that take refs doing more mutation and could thereby be faster? If so I think they should be called `put!` and `remove!`.

If they just do a copy of the Map/Sent passed to them and work on that copy I don't think we need them to be special versions.

hellerve:
It’s the other way around: the ones that take ownership are faster! I’ll remove the old versions, then.

hellerve:
Done!

eriksvedang:
Thank you!!!

eriksvedang:
Looks like there's still a `remove!` in there?

hellerve:
Not anymore!

eriksvedang:
I'm convinced, this optimisation of reduce is actually really important – thanks for finding it!

PR #321, opened 2018-11-01
Add Pattern.find-all

hellerve:
This PR adds `Pattern.find-all`, which behaves similarly to `Pattern.find`, but instead of only finding the index of the first occurrence, it will return an array of all occurrences within the provided source string. It also updates to `String.lines` to use this algorithm, which speeds up `String.split-by` by almost an order of magnitude because it reduces the number of Array allocations to one.

Of note is that this introduces the function `Pattern_internal_update_int_array`, which is extremely similar to what `Array_push_back__int` would generate (but with C in-place updates).

Test cases are included!

Cheers


eriksvedang:
Good stuff. The small duplication is fine, I think.

PR #322, opened 2018-11-07
test: use deftest everywhere

hellerve:
This PR simplifies the test harness by using `deftest` everyhwere instead of the `main` + `with-test` combo, because that’s basically equivalent, but the former is a little more declarative and shorter.

Cheers


eriksvedang:
Nice! Needs some rebasing now.

hellerve:
Done!

PR #328, opened 2018-11-08
Fixed SDL import capitalization in Life example

jburell:
Found a typo in the game of life example causing it to not compile under Linux (due to case sensitivity).

### DIFF
--- a/examples/life.carp
+++ b/examples/life.carp
@@ -4,8 +4,8 @@
 (use Double)
 (use Array)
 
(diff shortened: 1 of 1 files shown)


jburell:
Yes. Adding a regression test suite and perhaps a CI-server with VM:s for the supported targets might be helpful down the road.

hellerve:
Some work was done with Travis CI (see #230), but it’s currently not enabled, though I can’t remember why.

PR #331, opened 2018-11-13
Add sort flavors with custom comparators

hellerve:
This PR adds `Array.sort-by`, `Array.sorted-by`, and `Array.sort-by!`, which are generalized versions of the `sort` versions that take custom comparator functions.

I think @chrisosaurus wante to add this anyway.

Test cases are included!

Cheers

### DIFF
--- a/core/Heap.carp
+++ b/core/Heap.carp
@@ -33,7 +33,7 @@
 
     (defn push-down-until! [heap i len ord]
         (while true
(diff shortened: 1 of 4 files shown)


eriksvedang:
These are awesome!

chrisosaurus:
@hellerve thanks! I never got around to adding it, glad it wasn't left behind :)

PR #333, opened 2018-11-15
core: more sdl functions

hellerve:
This PR introduces a few more SDL functions. Those are important for picture saving, which I implemented for Anima ([this is what it looks like](https://github.com/hellerve/anima.carp/blob/master/examples/whispering_drink_picture.carp), [this is how it is implemented](https://github.com/hellerve/anima.carp/blob/master/anima.carp#L81)).

Cheers


eriksvedang:
Very nice!

Issue #332, opened 2018-11-14
Getting Started: fatal error: "stdbool.h" file not found


theronic:
Hello! I'm trying to get Carp going on OS X. I followed all the installation steps, but trying to eval anything throws a "stdbool.h" not found error. I assume Carp is trying to use clang, which I have installed. Running `clang -v` shows:

clang version 6.0.1  (emscripten 1.38.12 : 1.38.12)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Users/petrus/Projects/wallfly/emsdk/clang/e1.38.12_64bit

## The Error:

```
$ carp
Welcome to Carp 0.2.0
This is free software with ABSOLUTELY NO WARRANTY.
Evaluate (help) for more information.
鲮 (eval (my-function 12345 "w00t"))
In file included from ./out//main.c:1:
In file included from /Users/petrus/Projects/Carp/core/carp_debug.h:1:
/Users/petrus/Projects/Carp/core/core.h:5:10: fatal error: 'stdbool.h' file not found
#include <stdbool.h>
         ^~~~~~~~~~~
1 error generated.
carp: callCommand: clang -fPIC -lm ./out//main.c -shared -o "./out//Untitled"   -I/Users/petrus/Projects/Carp/core/  -g  (exit 1): failed
```

I'm sure this is not a Carp-specific issue, but I have no recourse to solve it.

theronic:
I managed to get things going by setting compiler to gcc, but I don't know how to make it stick?
```
(Project.config "compiler" "gcc")
```

 --important-flag didn't work

eriksvedang:
Hi!

You can make it stick by putting a file at ‘~/.carp/profile.carp’

Also, Clang should work if it knows it should use C99. Perhaps we need to send that flag as default..?

hellerve:
We could alternatively also just define the macros `bool`, `true`, and `false` ourselves to get rid of the header (that’s basically all it does anyway).

This would make the program more portable; the header is prefixed with `std`, but it isn’t actually super portable.

<details>
<summary>Extended info about the header</summary>

My assertions are mostly based [on the ISO standard](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdbool.h.html). YMMV.

For reference, this is what it looks like on my machine:
```C
/* Don't define bool, true, and false in C++, except as a GNU extension. */
#ifndef __cplusplus
#define bool _Bool
#define true 1
#define false 0
[...]

PR #334, opened 2018-11-17
Remove stdbool dependency

hellerve:
This PR adds our own version of `stdbool.h` to remove the C99 dependency. This should solve #332.

Cheers

### DIFF
--- a/core/carp_bool.h
+++ b/core/carp_bool.h
@@ -1,4 +1,4 @@
-#include <stdbool.h>
+#include "carp_stdbool.h"
 
 // Bool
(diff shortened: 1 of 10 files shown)


eriksvedang:
Nice. I'm a little confused though – shouldn't bool be typedef:ed to int?

hellerve:
Not according to most compilers. The `_Bool` data type [is a remnant of pre-C99/`stdbool.h` times](https://en.wikipedia.org/wiki/C_data_types#stdbool.h), but I’m pretty sure it will always be available.

eriksvedang:
Ok!

PR #336, opened 2018-12-01
fix Map and Set behaviour for negative hash values

### LINKED ISSUE #335: Maps don't work with negative int keys
```
(defn main []
  (let [empty (the (Map Int Bool) (Map.create))
        key -7]
    (if (Map.contains? &empty &key)
      (IO.println "yes")
      (IO.println "no"))))
```
crashes with
```
Untitled: ./out//main.c:1482: Bucket__int_bool *Array_nth__Bucket__int_bool(Array *, int): Assertion `n >= 0' failed.
```

The issue seems to be that Map uses `(mod hashcode size)` to find the bucket, and `mod` returns negative numbers for negative first arguments.

opqdonut:
fixes #335

- add new Int.positive-mod function
- use it for Map and Set operations
- add tests

### DIFF
--- a/core/Int.carp
+++ b/core/Int.carp
@@ -53,6 +53,13 @@
           (set! y (/ y 2))
           (set! x (* x x))))
       r))
+
+  (doc positive-mod "Like mod but always returns a positive answer.")
+  (defn positive-mod [k n]
+    (let [r (Int.mod k n)]
+      (if (> 0 r)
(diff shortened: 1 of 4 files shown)


opqdonut:
found a bug

opqdonut:
amended

eriksvedang:
Awesome!

PR #339, opened 2018-12-03
Map.to-array and Set.to-array functions, equality for non-ref Pairs

opqdonut:
Some additions I think would fit the standard library.

Had to add the Pair equality to be able to write the tests for to-array :)

### DIFF
--- a/core/Map.carp
+++ b/core/Map.carp
@@ -225,6 +225,12 @@
           (set! m (put m k v))))
       m))
 
+  (doc to-array "Convert Map to Array of Pairs")
+  (defn to-array [m]
+    (kv-reduce (fn [arr k v] (Array.push-back arr (Pair.init-from-refs k v)))
+               []
+               m))
+
   (defn str [m]
     (let [res (kv-reduce (fn [s k v]
                            (String.join @"" &[s @" " (prn @k) @" " (prn @v)]))
@@ -342,6 +348,10 @@
             (let [e (Array.nth entries j)]
               (set! init (f init e))))))
       init))
+
+  (doc to-array "Convert Set to Array of elements")
+  (defn to-array [s]
+    (reduce (fn [arr elt] (Array.push-back arr @elt)) [] s))
 
   (defn str [set]
     (let [res (reduce (fn [s e] (String.join @"" &[s @" " (prn e)]))
--- a/core/Tuples.carp
+++ b/core/Tuples.carp
@@ -5,7 +5,18 @@
     (Pair.init @r1 @r2))
 
   (defn = [p1 p2]
+    (and (= (Pair.a &p1) (Pair.a &p2))
+         (= (Pair.b &p1) (Pair.b &p2))))
+
+  (defn /= [p1 p2]
+    (not (Pair.= p1 p2)))
+
+  )
+
+(defmodule PairRef
+  (defn = [p1 p2]
     (and (= (Pair.a p1) (Pair.a p2))
          (= (Pair.b p1) (Pair.b p2))))
 
-  )
+  (defn /= [p1 p2]
+    (not (PairRef.= p1 p2))))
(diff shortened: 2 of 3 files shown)


core/Map.carp
```diff
@@ -154,6 +154,18 @@
...
       m))
 
+  (doc to-array "Convert Map to Array of Pairs")
+  (defn to-array [m]
+    (let-do [ret (Array.zero)]
+      (for [i 0 @(n-buckets m)]
+        (let [bucket (Array.nth (buckets m) i)
+              len (Array.length (Bucket.entries bucket))
+              entries (Bucket.entries bucket)]
+          (for [j 0 len]
```
hellerve: You could use `Array.concat` here to avoid code duplication and maybe even make it a little more performant!

core/Map.carp
```diff
@@ -262,6 +274,18 @@
...
       s))
 
+  (doc to-array "Convert Set to Array of elements")
+  (defn to-array [s]
+    (let-do [ret (Array.zero)]
+      (for [i 0 @(n-buckets s)]
+        (let [bucket (Array.nth (buckets s) i)
+              len (Array.length (SetBucket.entries bucket))
+              entries (SetBucket.entries bucket)]
+          (for [j 0 len]
```
hellerve: See my comment above on `Array.concat`, it applies here as well.

core/Map.carp
opqdonut: good call, I was just blindly following the structure of `for-each`

hellerve: (Aside: due to a problem in my Haskell build pipeline I can’t test or reproduce any of your issues or PRs at the moment. I’m sorry!)

opqdonut: I came up with

```
  (doc to-array "Convert Map to Array of Pairs")
  (defn to-array [m]
    (Array.concat
     &(Array.copy-map &(fn [b] @(Bucket.entries b)) (buckets m))))
```

but that copies each bucket twice instead of zero times like the for-loop version.

hellerve: Ah, that’s a bummer, I hadn’t thought of that. Sadly there is not `endo-map` version that changes the type, since that would potentially change the size of the array/its slots.

It’s definitely a shame, because this version looks much cleaner to me. Maybe @eriksvedang has some suggestions?

opqdonut:
If #345 gets merged, these can be reimplemented with `Map.kv-reduce` and `Set.reduce`

hellerve:
It seems like there are some conflicts here as well, should we resolve them now? It looks like #345 might create new conflicts anyway...

opqdonut:
rebased on top of master, reimplemented using reduce

core/Tuples.carp
```diff
@@ -4,8 +4,19 @@
   (defn init-from-refs [r1 r2]
     (Pair.init @r1 @r2))
 
+  (defn = [p1 p2]
```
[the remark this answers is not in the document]
opqdonut: I needed it for the test case. It seems the equality for arrays calls non-ref equality for elements.
```
(defn main []
  (println* (= &[(Pair.init 1 2)]
               &[(Pair.init 1 2)])))
```
```
Can't find matching lookup for symbol '/=' of type (λ [(Pair Int Int), (Pair Int Int)] Bool) at line 59, column 18 in '/home/opqdonut/proj/adventofcode18/Carp/core/Array.carp
```

... but maybe the = for Array should be fixed instead?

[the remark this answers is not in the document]
opqdonut: In any case, sometimes it's more natural to write code that compares non-ref pairs, something like...
```
(if (= (Pair.init x y) (Map.get my-map key))
   ...)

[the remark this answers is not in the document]
eriksvedang: OK, fair enough! The `Array.=` is a good reason for now, if that gets resolved we'll have to think about what instances of `=` to keep around.

PR #341, opened 2018-12-03
fix Map.put and Set.put

### LINKED ISSUE #340: Map.put doesn't update values
```
(defn main []
  (let [m (Map.create)
        m1 (Map.put m &1 &2)
        m2 (Map.put m1 &1 &3)]
    (do
      (IO.println &(str (Map.length &m2)))
      (IO.println &(str &m2)))))
```
prints
```
2
{ 1 2 1 3 }
```

I'll try to make a PR

opqdonut:
fixes #340

### DIFF
--- a/core/Map.carp
+++ b/core/Map.carp
@@ -58,7 +58,21 @@
           (do
             (set! e (Pair.b (Array.nth es i)))
             (break))))
-       @e))
+      @e))
+
+  (defn put [b k v]
+    (update-entries
+     @b
+     &(fn [es]
+        (let-do [found false]
+          (for [i 0 (Array.length &es)]
+            (when (= k (Pair.a (Array.nth &es i)))
+              (do
+                (set! found true)
+                (Array.aset! &es i (Pair.init-from-refs k v)))))
+          (unless found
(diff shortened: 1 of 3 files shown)


core/Map.carp
```diff
@@ -58,7 +58,21 @@
...
+  (defn put [b k v]
+    (update-entries
+     @b
+     &(fn [es]
+        (let-do [found false]
+          (for [i 0 (Array.length &es)]
+            (when (= k (Pair.a (Array.nth &es i)))
+              (do
+                (set! found true)
+                (Array.aset! &es i (Pair.init-from-refs k v)))))
```
hellerve: You can use `break` here for an early loop termination.

core/Map.carp
```diff
@@ -104,7 +118,7 @@
     (let [idx (Int.positive-mod (hash k) @(n-buckets &m))]
       (update-buckets m &(fn [b]
         (let [n (Array.nth &b idx)]
-          (Array.aset b idx (Bucket.grow n (Pair.init @k @v))))))))
+          (Array.aset b idx (Bucket.put n k v)))))))
```
hellerve: Do we need `Bucket.grow`still then?

Issue #343, opened 2018-12-04
Array.sort subtly broken


opqdonut:
Sorting large arrays doesn't always quite sort the data.

Luckily I managed to reproduce this with a fairly small example:

```
(defn hash-comparison [a b]
  (< (hash a)
     (hash b)))

(defn main []
  (let [expected (Array.copy-map &(fn [i] (str @i)) &(Array.range 100 200 1))
        ;; shuffle input by sorting by hash:
        in (Array.sort-by @&expected &hash-comparison)]
    (do
     (let [out (Array.sort in)]
       (do
        (println* &out)
        (println* (= &out &expected)))))))
```
Output:
```
[@"100" @"101" @"102" @"103" @"104" @"105" @"106" @"107" @"108" @"109" @"110" @"111" @"112" @"113" @"114" @"115" @"116" @"117" @"118" @"119" @"120" @"121" @"122" @"123" @"124" @"125" @"126" @"127" @"128" @"129" @"130" @"131" @"132" @"133" @"134" @"135" @"136" @"137" @"138" @"139" @"140" @"141" @"142" @"143" @"144" @"145" @"146" @"147" @"148" @"149" @"150" @"151" @"152" @"153" @"154" @"155" @"156" @"157" @"158" @"159" @"160" @"161" @"162" @"163" @"164" @"165" @"166" @"167" @"168" @"169" @"170" @"178" @"171" @"172" @"173" @"174" @"183" @"175" @"176" @"177" @"179" @"180" @"181" @"182" @"184" @"186" @"185" @"187" @"188" @"189" @"190" @"191" @"192" @"193" @"194" @"195" @"196" @"197" @"198" @"199" @"200"]
false
```

Note the part `@"170" @"178" @"171"`.

hellerve:
That’s interesting! This must be a regression of moving away from `qsort` (a C function) to our own heap-based sort (see also #262 #252 #331). At least I hope it is, because otherwise I’m at a loss what it could be!

opqdonut:
yeah an off-by-one in the heapsort is my guess as well

eriksvedang:
Who wants to fix it?

hellerve:
I can take a look at it when I get my pipeline back on track. I can’t promise I’ll have time for that this week, maybe next. I’ll keep y’all posted if I can’t make it happen!

chrisosaurus:
Ohhh awesome! Thanks for the awesome bug report.

[...]

PR #344, opened 2018-12-07
Bug #343 Fixing off-by-one in Heap parent function.

chrisosaurus:
Adding more comprehensive tests for Heaps to prevent happening again..

### DIFF
--- a/core/Heap.carp
+++ b/core/Heap.carp
@@ -15,7 +15,7 @@
 
     (hidden parent)
     (defn parent [i]
-        (/ i 2))
+        (/ (- i 1) 2))
 
     (hidden max-of-three-until)
     (doc max-of-three-until! "Get the index for the largest (by ord) of an element and its two children.")
--- a/test/heap.carp
+++ b/test/heap.carp
@@ -5,12 +5,12 @@
 
 (deftest test
   (let-do [arr [1 3 4 2 6 1]
-           exp [1 1 2 3 6 4]]
+           exp [1 2 1 3 6 4]]
           (MinHeap.heapify! &arr)
           (assert-equal test
                         &exp
                         &arr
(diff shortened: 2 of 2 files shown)


hellerve:
Is it possible for `(- i 1)` to be `< 0`?

opqdonut:
@hellerve yeah, if somebody tries to find the parent of the root (i=0)

How about adding some tests that check `(= i (parent (lchild i)))` and `(= i (parent (rchild i)))`? The current tests are fine but that's the exact thing that was broken, right?

chrisosaurus:
I is only meant to be an array index, and my code only calls it for I >=1
Even if you call it with I = 0 then (/ (- I )1 2) should still be 0


@opqdonut I constructed the new tests by going through how a heap should work on paper at each step, and then codifying the interesting cases that the previous implementation missed (that our tests didn’t cover).

We can always add more tests, even as follow up prs. I don’t have access to a machine where I can code, I can add some more tomorrow morning.
You’re also welcome to send prs for more tests, the carp tests are generally quite light and are more smoke tests than in-depth tests.


I was thinking of trying to write a fuzzer, or exposing another sort (even just qsort) - that way we could just generate large random sequences and check that our sort matches the other implemention(s).
I might try write a quick sort in carp, and then re-add a qsort wrapper, then we can let the user pick which they want and we can test that they all have the same output for large semi random inputs.

opqdonut:
Yeah I'm definitely not against merging this, just thinking out loud.

eriksvedang:
Thanks for the quick fix, @chrisosaurus!

chrisosaurus:
@opqdonut oh don't get me wrong, I think it is a good idea to add more testing both here and to wider Carp. I suspect your AoC is probably the most complex body of code we currently have in Carp!

PR #345, opened 2018-12-07
core/Map.carp: some new functions

opqdonut:


opqdonut:
some stuff thought might have a place in core

core/Map.carp
```diff
@@ -123,6 +123,25 @@
     (let [idx (Int.positive-mod (hash k) @(n-buckets m))]
       (Bucket.get (Array.nth (buckets m) idx) k)))
 
+  (doc update "Update value at key k in map with function f, if it exists.")
+  (defn update [m k f]
+    (let [idx (Int.positive-mod (hash k) @(n-buckets &m))]
+      (update-buckets m &(fn [b]
+        (let [n (Array.nth &b idx)]
+          (if (Bucket.contains? n k)
+            (Array.aset b idx (Bucket.put n k &(f (Bucket.get n k))))
```
hellerve: If we want this to be a bit speedier, we could add a function `Bucket.update`; this could reuse the index and thus get rid of an extra lookup.

opqdonut: yeah that makes sense

opqdonut: ok I managed to whip up something that works, sidesteps the bug, and even looks kinda neat

core/Map.carp
```diff
@@ -156,6 +178,43 @@
...
+  (defn endo-map [f m]
+    (do
+      (for [i 0 @(n-buckets &m)]
+        (let [bucket (Array.nth (buckets &m) i)
+              len (Array.length (Bucket.entries bucket))
+              entries (Bucket.entries bucket)]
+          (for [j 0 len]
+            (let [e (Array.nth entries j)]
+              (Array.aset! entries j (Pair.init @(Pair.a e)
+                                                (f (Pair.a e) (Pair.b e))))))))
```
opqdonut: I'd love to do something like this:
```
(Array.endo-map (fn [p] (Pair.update-b p f)) entries)
```
but that too is broken by #347. f would have to be a ref, and closing over function refs is broken

hellerve:
Giving this a cursory glance, this looks much better now. Do you think it’s in a mergeable state as-is or would you like to wait until #347 is fixed?

opqdonut:
I think it makes sense to merge this. Some of the implementations can be streamlined once #347 is in, but I think they're nice enough already.

core/Map.carp
eriksvedang: Yeah, definitely that looks like *the* thing to be able to do! Sorry about lambdas being even more broken than I thought... ^^

eriksvedang:
Cool stuff!

Issue #317, opened 2018-10-30
RFC: non-owning filter


hellerve:
The current implementation of `filter` takes ownership of the array it’s being passed. This makes sense, because it modifies the array in-place, but it also means that we always have to copy the array before passing it in if we want to keep working with it, which in my experience is a fairly common workflow.

To make this easier and more performant, I propose that we add a function `filter-copy`, similar to `copy-map`, that copies the array. Whether it should be called `filter-copy` or `copy-filter` I’m not entirely sure about. Aesthetically I prefer the former, but the latter is more consistent (and consistency is what’s ultimately more important).

This needs to be added as an array template, and I’d be up to add this if we reach a consensus whether this makes sense or just adds bloat.

eriksvedang:
I think it's a fairly common use case and in some cases this optimisation will mean a lot (especially if a big array is filtered down to something small). Consistent naming using `copy-filter` seems like the best option, yes. And even though none of the names are particularly beautiful, at least they warn the programmer about their copying nature, which is also good.

hellerve:
Should `filter` then be renamed to `endo-filter`?

chrisosaurus:
+1 to being explicit about copys, it is one of the biggest problems I often face in C++ (accidental copies are too easy).

hellerve:
I think we needn’t be consistent, then!

hellerve:
There is a tradeoff about the characteristics of `copy-filter`, though. It could look like this:

```clojure
(defmodule Array
  (defn copy-filter [f a]
    (let-do [res []]
      (for [i 0 (length a)]
        (when (f @(nth a i)) (push-back! &res @(nth a i))))
      res))
)
```

This would have to allocate every once in a while, but would be good if we filter “aggressively”, i.e. if only few elements remain. On the other hand, if we implemented it like this:

```C
(defmodule Array
  (defn copy-filter [f a] (filter f @a))
)
```

It would copy the whole array once, and thus perform less allocations; it would, however, waste space. This would go away if we optimized `filter` as planned, however (resizing it properly after we’re done filtering).

What do y’all think?

hellerve:
I can do it over the weekend!

eriksvedang:
Awesome!

PR #346, opened 2018-12-07
core/Array.carp: Array.remove and Array.remove-idx functions

opqdonut:


core/Array.carp
```diff
@@ -240,4 +240,18 @@
...
+  (doc remove "Remove occurrences of element el in array arr, in-place")
+  (defn remove [el arr]
+    (filter &(fn [x] (not (= el x)))
+            arr))
+
+  (doc remove-idx "Remove element at index idx from arr")
+  (defn remove-idx [i arr]
+    (do
+      ;;(assert (<= 0 i))
+      ;;(assert (< i (Array.length &arr)))
```
opqdonut: I thought about adding assert as a template, but I don't want a separate wrapper function (line numbers will be misleading, for instance)

hellerve: How about naming this `remove-nth`? That name seems a little more congruent with the rest of the API?

core/Array.carp
```diff
@@ -240,4 +240,18 @@
     (zip &Pair.init-from-refs
          &(range 0 (length xs) 1) ;; Inefficient, creates a temporary array.
          xs))
+
+  (doc remove "Remove occurrences of element el in array arr, in-place")
```
chrisosaurus: I’m not familiar with our filter, Is our filter really an inplace mutation?
what does In place mean here?

Usually our functions which mutate things in place end with an exclamation mark ala scheme, at least within the array libraries. Is that convention no more? (I haven’t been around for a bit).

hellerve: It is, but it’s save because it takes ownership, and thus we do not put a `!` at the end. That’s why I opened #317 (I want a version that doesn’t take ownership as well).

core/Array.carp
opqdonut: definitely

core/Array.carp
eriksvedang: Yeah, we should resolve that (it's basically just naming decisions left to do there, IIRC). I'll merge this for now, since this looks useful on its own.

PR #367, opened 2018-12-18
core/Map.carp: get-with-default

opqdonut:
nice for types with no (zero) defined

### DIFF
--- a/core/Map.carp
+++ b/core/Map.carp
@@ -76,11 +76,11 @@
     (do (Array.push-back! (entries &b) (Pair.init-from-refs k v))
         b))
 
-  (defn get [b k]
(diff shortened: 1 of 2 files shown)


hellerve:
Fun fact: this is how I originally implemented `get` for hashmaps containing types with expensive `zero` calls. I didn’t think of using both, though

opqdonut:
I was thinking of a `get-or-error` function too

hellerve:
Me too, but we don’t have standard error handling yet (I was waiting on sum types for that).

eriksvedang:
Yes, a `Maybe` type will be used for the normal `Map.get` when we have those (sooon...).

But this seems useful in its own right!

PR #372, opened 2019-01-16
core: simplify String.hash

hellerve:
This PR simplifies `String.hash`. Specifically, it removes the distinction between `hash` and `rehash`, because `rehash` isn’t used elsewhere. This also does away with a useless multiplication by `1`—which hopefully got optimized away anyway.

Cheers


chrisosaurus:
Nice find.

PR #373, opened 2019-01-21
core: refactor the vector module; generic types and use module system

hellerve:
This PR refactors the Vector module to pull the types out of the module and rename them to the module they belong in. It also “generifies” the type, so that they’re not bound to the `Double` type. This means we can now also have `Int` vectors and such! Yay!

Cheers


eriksvedang:
Love it!

Issue #358, opened 2018-12-13
String escape parsing broken


opqdonut:
```
(def s "\\")

(defn main []
  (println* s))
```
```
[PARSE ERROR] ".../dbg.carp" (line 5, column 1): unexpected end of input expecting "\\" or "\""
```

opqdonut:
this works:
```
(def s "\\ ")

(defn main []
  (println* s))
```

hellerve:
I have a possible fix for this in my branch [fix-string-escapes](https://github.com/hellerve/Carp/tree/fix-string-escapes), but I haven’t been able to fix my setup yet.

eriksvedang:
Oh? I was just playing around with this but couldn't come up with anything quickly... what's the trick?

eriksvedang:
Checking out your solution now...

eriksvedang:
Doesn't seem to solve this particular error I'm afraid... what's the reasoning behind it?

hellerve:
I completely forgot this issue, sorry! I have a fix for this that seems to work rather nicely, but sadly obfuscates the string parser a little bit. I’m going to make a PR and you can tell me whether you find a more elegant solution!

hellerve:
We should be able to close this (because of #375).

PR #375, opened 2019-02-01
Fix string escaping of backslashes (#358)

hellerve:
This PR fixes the issue that we encountered in #358 where the parser for escaped quotes was just a little too greedy. This PR special-cases this a bit more, which sadly makes the string parser a little more complex. Do tell me if there is a better way to do this!

I also added a test case!

Cheers

### DIFF
--- a/src/Parsing.hs
+++ b/src/Parsing.hs
@@ -87,10 +87,13 @@ number = Parsec.try float <|>
 string :: Parsec.Parsec String ParseState XObj
 string = do i <- createInfo
             _ <- Parsec.char '"'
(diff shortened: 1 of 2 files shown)


eriksvedang:
I don't undertand this patch completely but thanks a lot for fixing the bug :)

hellerve:
So basically what the problem was was that the only thing we escaped was `\"`, but that meant that in the string `"\\" we parsed `Start String -> \ -> \"`, where it should have been `Start String -> \\ -> End String`. This meant that `escaped` returns a `String` instead of a `Char` now, though, and I had to update the call site accordingly.

eriksvedang:
Ok, thanks a lot for the clarification!

PR #376, opened 2019-02-02
Add copy-filter and rename filter to endo-filter

hellerve:
This PR does the work we discussed in #317. It also adds a check to `endo-filter` that resizes the array if appropriate.

Cheers

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -261,8 +261,8 @@
 
   (doc remove "Remove occurrences of element el in array arr, in-place")
   (defn remove [el arr]
-    (filter &(fn [x] (not (= el x)))
-            arr))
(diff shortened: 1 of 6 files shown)


eriksvedang:
I think this looks great!
One comment before merging, the function `templateShrinkCheck` takes an unused argument, perhaps fix that first?

eriksvedang:
Excellent. I think a test in `memory.carp` that triggers the shrinking would be good.

hellerve:
Should I add this or do you want to?

eriksvedang:
If you have the time to do it, please go ahead! I might add some more tests for it later, if I can think of something that might break it...

Issue #374, opened 2019-01-26
Allow for customization within documentation generation


hellerve:
Currently, everything in the documentation generator is hardcoded. This is a little annoying when writing custom documentation, and it would be nice if we could rectify this. An example for some of the things I’d like to be able to do can be found in [the documentation for anima](http://veitheller.de/anima/). I’m hiding the navigation, using a custom icon, and add a documentation prelude. Stuff like this should be supported.

Things I want to be able to customize:

- [x] Styling
- [ ] Which sections to display
- [x] The icon
- [x] Add custom prelude/sections
- Any other ideas?

I’d be up for working on this, and I’d like to hear about any wishes you might have to create a great automatic documentation system! This issue should serve as a conversation starter and brain dump.

Cheers

hellerve:
Another one: I’d like to be able to write markdown in the doc strings (and have it be rendered as HTML on the page). People probably have strong feelings about what markup language to use, but Markdown seems like a pretty common choice and is used by a lot of people already.

eriksvedang:
Sounds like a good idea.

hellerve:
It seems like a lot of this can be fulfilled by changing `projectIndexPage` in `RenderDocs`. I believe rethinking the entire documentation rendering architecture might be worth it, if we have the time, but if push comes to shove, that function might be all that we need to change!

eriksvedang:
OK, yeah I haven't looked at it in a while but I'm definitely open to changing the structure there, if needed.

hellerve:
I’m going to look at this over the weekend, but I’m not sure whether I’ll find a satisfying API just yet. It just hasn’t quite cleared up to me.

eriksvedang:
Sounds good!

eriksvedang:
Can this be closed or are there more features needed?

hellerve:
I think #459 just about made this perfect!

PR #377, opened 2019-02-02
Make some properties about the documentation generator configurable

hellerve:
This PR adds properties to customize a bunch of properties about documentation generation, namely:

- the logo
- the prelude
- the URL
- the styling (CSS) URL

It doesn’t do everything described in #374, most notably it doesn’t allow for customized sections, but it’s a start!

Cheers

### DIFF
--- a/src/RenderDocs.hs
+++ b/src/RenderDocs.hs
@@ -25,7 +25,7 @@ saveDocsForEnvs ctx envs =
       allEnvNames = (fmap (getModuleName . snd) envs)
   in  do mapM_ (saveDocsForEnv ctx allEnvNames) envs
          writeFile (dir ++ "/" ++ title ++ "_index.html") (projectIndexPage ctx allEnvNames)
-         putStrLn ("Generated docs to '" ++ title ++ "'")
+         putStrLn ("Generated docs to '" ++ dirPath ++ "'")
 
 
 projectIndexPage :: Project -> [String] -> String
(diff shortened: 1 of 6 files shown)


src/RenderDocs.hs
```diff
@@ -18,58 +18,68 @@ import Obj
...
-         writeFile (dirPath ++ "/" ++ projectTitle ++ "_index.html") (projectIndexPage projectTitle allEnvNames)
-         putStrLn ("Generated docs to '" ++ dirPath ++ "'")
+saveDocsForEnvs :: Project -> [(SymPath, Env)] -> IO ()
+saveDocsForEnvs ctx envs =
+  let dir = projectDocsDir ctx
+      title = projectTitle ctx
+      allEnvNames = (fmap (getModuleName . snd) envs)
+  in  do mapM_ (saveDocsForEnv ctx allEnvNames) envs
+         writeFile (dir ++ "/" ++ title ++ "_index.html") (projectIndexPage ctx allEnvNames)
+         putStrLn ("Generated docs to '" ++ title ++ "'")
```
bjorn3:
```suggestion
         putStrLn ("Generated docs to '" ++ dirPath ++ "'")
```

hellerve: Whoops, you’re right of course!

eriksvedang: It can't find this variable so the build fails. Where is it defined?

hellerve:
Fair enough! I’ll add that back for now.

hellerve:
I readded the title as a `h1` but omitted the `huge` class for now. If you really want that I can add it too!

PR #378, opened 2019-02-02
memory test: test shrinking

hellerve:
This PR changes one of the filter memory tests so that we test the shrinking. It’s very basic and there a re probably more complex examples that are more likely to contain a memory bug, but this should be useful as a first smoke test.

Cheers


eriksvedang:
Cool, thanks!

PR #382, opened 2019-02-15
Multiple fixes for sumtypes

hellerve:
Okay, this is a big one:

- It reworks the docs for `Array`.
- It makes `Array` safer by making `first` and `last` return `Maybe` (`nth` still needs to be fixed). `unsafe-first` and `unsafe-last` are still there if you need them. Tests have been added for the new functionality and the examples have been fixed.
- add tests to the `Result` module.
- Document the `Result` and `Maybe` modules.
- Add `Maybe.to-result` for conversion—it is the inverse of `Result.to-maybe`.

I think that’s it?

I realize that changing core `Array` functions like that probably breaks a lot of code, but I think it’s necessary to achieve safety going forward.

Cheers

### DIFF
--- a/bench/array_update.carp
+++ b/bench/array_update.carp
@@ -10,7 +10,7 @@
 
 (defn some-updating []
   (let-do [a (replicate n &12345)
-        b (aupdate a 0 inc-ref)]
+        b (aupdate a 0 &inc-ref)]
     (assert (= (nth &b 0) &12346))
     (assert (= (nth &b 1) &12345))))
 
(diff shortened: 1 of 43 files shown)


hellerve:
I also added the following:
- avoid function copies in `Bench`
- fix the `array_update.carp` benchmark
- add `Filepath.file-from-path`
- add tests for the `Filepath` module
- reformat a lot of other documentation

eriksvedang:
Looks great!

PR #383, opened 2019-02-28
Fix for issue 365

xran-deex:
I'm not 100% sure about my second commit, but the first one at least fixes the issue of not being able to compile when filling arrays with functions. The second commit actually fixes a segfault when compiling and running on Windows. Apparently, Windows is pickier with branching on uninitialized values, so I added some code when allocating an array of lambdas to at least initialize the struct fields with NULLs. This appears to fix the Windows issue.

### DIFF
--- a/examples/arrays.carp
+++ b/examples/arrays.carp
@@ -1,0 +1,25 @@
+(use Array)
+(use String)
+
+(defn ex1 []
+    (let [a (Array.allocate 2)
+        b (Array.aset a 0 (fn [] @"Hello"))
+        c (Array.aset b 1 (fn [] @"World"))
+    ]
+    (do
+        (println* (@(Array.nth &c 0)))
+        (println* (@(Array.nth &c 1))))))
+
+(defn ex2 []
+    (let [a (Array.allocate 2)
+        b (Array.aset a 0 (fn [x] (String.append "Hello" x)))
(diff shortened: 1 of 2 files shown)


hellerve:
The first commit looks good to me! The second one seems also fine to me, unless someone has a less costly way of initializing the array.

About Windows: did you run Carp on WSL? I thought we had abandoned Windows support completely for the time being, but if Carp still runs on WSL, that’s great!

eriksvedang:
Looks very good, thanks a lot.

Windows support is not abandoned..! There just isn't anyone working on it very often :)

xran-deex:
@hellerve I compiled on Windows using both the Windows C compiler (from Visual Studio 2017) and GCC (from MinGW/Git Bash). There were only a few small changes needed to build the compiler on Windows. First, replacing the unix Haskell package with an alternative cross-platform terminal package. (I believe the unix package was only used to add color to the terminal output.) Second, I had to add some conditional includes in core.h.

hellerve:
Would you like to share these changes as well, so that we can make Carp proper work on Windows again? That would be great and much appreciated!

eriksvedang:
Yes, this sounds awesome! Especially getting it to work with Visual Studio will be very important moving forward, since that's the one reliable way to make actual windows programs in my experience.

PR #386, opened 2019-02-28
Fix sumtype printing

hellerve:
This PR fixes the printing of sumtypes by accounting for the NULL byte in their calculated size.

Cheers


hellerve:
Oh crap, this also contains my pattern fix from earlier (#385). I guess they need to be backported anyway, but if that’s a problem, I could remove them.

PR #388, opened 2019-03-03
Build Carp on Windows

xran-deex:
I think this is the minimal set of changes that were needed to build Carp on Windows. These changes will at least allow one to compile on Windows using gcc. Compiling with the Windows compiler (cl.exe) still doesn't work due to some specific Posix functions that are used in some carp header files, specifically carp_system.h's clock_gettime usage. There may be other's, too, that will prevent compilation on Windows using cl.exe.

### DIFF
--- a/CarpHask.cabal
+++ b/CarpHask.cabal
@@ -58,7 +58,7 @@ library
                      , blaze-html
                      , blaze-markup
                      , text
-                     , unix
+                     , ansi-terminal >= 0.9
                      , cmark
 
   default-language:    Haskell2010
--- a/core/core.h
+++ b/core/core.h
@@ -4,7 +4,11 @@
 #include <assert.h>
 #include "carp_stdbool.h"
 #include <stddef.h>
+#if defined(WIN32) || defined(_WIN32) || defined(__WIN32) && !defined(__CYGWIN__)
+#include <windows.h>
+#else
 #include <sys/wait.h>
+#endif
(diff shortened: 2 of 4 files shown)


28:
Nice! This is something I really need.

eriksvedang:
Woohoo, awesome!!! I'm merging this asap.

One small question though: what's the reason for putting the dependency into the "extra deps" field? (I'm not very good at Stack, so sorry if this is a stupid question)

xran-deex:
@eriksvedang The version of ansi-terminal that has the method "hSupportsANSIColor" apparently isn't in the snapshot/resolver that is specified in the stack.yaml file ("lts-11.12"), so it needs to be pulled in as an extra dependency. (I'm not a Stack expert, but you can read about it [here](https://github.com/commercialhaskell/stack/blob/master/doc/yaml_configuration.md))

eriksvedang:
@xran-deex OK, thanks – that makes a lot of sense. I'll try to bump the version number of the solver eventually and see if the extra-deps can be removed.

eriksvedang:
So I merged this and have gotten rid of all the warnings from clang-cl.exe etc. The basic examples are working fine, which is very exciting!

There is one annoying thing (on my machine) though - it seems as if `hSupportsANSIColor` reports True on Windows (using Powershell) even though it still renders the escaped codes as text (and not as colors). Is that the same for you, @xran-deex ?

PR #397, opened 2019-03-17
Change "defdynamic" to "defndynamic" in guide

cnasc:
Based on the docs, I expected that `defdynamic` would allow me to define and use a function at the REPL. Instead I got the following error:

```
> (defdynamic square [x]
     (* x x))
Invalid args to `defdynamic`: (defdynamic square [x] (* x x))
(did you try to define a dynamic function - use 'defndynamic' instead) at REPL:1:1.
```

Definitely props on the helpfulness of the error message, which makes this an easily overcome issue. However, I'd like to propose this change to the guide so that any other newcomers don't hit a bump the first time they try something.


hellerve:
Heh, we changed that name 3 days ago! Good catch!

cnasc:
Haha, looked at the commit log after posting this and noticed that was the case. At a glance, there are a few other places in the docs that need the same substitution, if nobody else gets to it I’ll whip up a more thorough PR instead of this one I did through the web interface.

eriksvedang:
Oops, I must have missed some dirs when doing search-and-replace – sorry!

PR #401, opened 2019-03-26
Make `Array.range` more general

hellerve:
This PR generalizes `Array.range`. Prior to this change, `Array.range` assumed that we could use some `Int` functions on the input that prevented it from being generalized. By using `to-int` and making `to-int` and `from-int` interfaces, we can use other types in the function as well, and do cool stuff like `(Array.range 0.0 10.0 0.5)`!

A test case is included.

Cheers


eriksvedang:
Cool!!!

PR #403, opened 2019-03-27
REPL: better autocompletion based on starting environment

hellerve:
This PR injects the starting environment into the REPL autocompletion. This means that all standard functions can now be autocompleted.

You can try this out by typing `In` in your REPL, having it autocomplete (by hitting TAB) to `Int.`, and then, if you hit TAB again, see a list of functions defined in the `Int` module!

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -14,6 +14,7 @@ import Repl
 import StartingEnv
 import Eval
 import Util
+import Lookup
 
 defaultProject :: Project
 defaultProject =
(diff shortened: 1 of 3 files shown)


eriksvedang:
Wow, epic!
This is so good. I should try to incorporate it with the emacs mode also...

Is it a lot more work to make it pick up new definitions?

hellerve:
I am not quite sure how to do this with Haskeline.

In zepto I solved it by writing my own REPL and reimplementing readline in zepto, but I don’t think that’s a good idea :)

eriksvedang:
No, that sounds like a lot of work... one would hope that the creators of Haskeline had thought of this use case. Perhaps it's easier to fix Haskeline.

eriksvedang:
Or some abuse of `unsafePerformIO` :)

hellerve:
An alternative would be to set up a new version of the REPL context every time we enter into the `repl` function. That would work, although I’m not quite sure how to express that. I’ll think about it a bit

PR #407, opened 2019-03-29
update core docs

opqdonut:


hellerve:
Looks good.

A little aside: If you ever feel the need to jazz up your documentation, you can write markdown inside docstrings now and it will be rendered correctly! That’s especially cool if you want to provide code snippets as examples.

PR #408, opened 2019-04-02
Evaluate symbols in `members`

hellerve:
This PR changes `members` so that any arguments passed to it are first evaluated as far as possible. This is useful because then we can use it in macros.

An illustrating example:

```clojure
(defndynamic derive-eq-internal2 [ms]
  (if (= (length ms) 0)
    true
    (list 'and (list '= (list (caar ms) 'o1) (list (caar ms) 'o2))
               (derive-eq-internal2 (cdr ms)))))

(defndynamic derive-eq-internal [a] (derive-eq-internal2 (members a))) ; this would fail saying that a is not a module

(defmacro derive-eq [t]
  (list 'defmodule t
    (list 'defn '= (array 'o1 'o2) (derive-eq-internal t))))
```

Also, the above code implements (probably) correct deriving of `=` for arbitrary structs, ain’t that cool?!

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -265,7 +265,7 @@ eval env xobj =
           return (makeEvalError ctx Nothing ("Invalid args to `meta`: " ++ pretty xobj) (info xobj))
 
         [XObj (Sym (SymPath [] "members") _) _ _, target] ->
-          specialCommandMembers target
+          specialCommandMembers target env
         XObj (Sym (SymPath [] "members") _) _ _ : _ ->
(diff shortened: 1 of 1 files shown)


eriksvedang:
Awesome, that macro is exactly what I had in mind for the `members` function – we should add that kind of thing for other "traits" too!

hellerve:
Totally! The recursion into the types is not a problem, but the composition is probably not generic. Maybe I can think of a factory for derivers, though. I’ll think about it :)

PR #411, opened 2019-04-16
Fix doc for get-args-len

scolsen:
I was running through the docs and it seemed like get-arg was incorrectly documented based on the signature. It looks like the doc for get-args-len was reassigned to get-arg.


eriksvedang:
Thanks!

PR #412, opened 2019-04-17
Remove headers that are problematic for embedded applications

hellerve:
This PR removes some headers—or makes them conditional—that are either unnecessary or only needed under certain circumstances (like `assert`, which is only used for unoptimized builds).

The test cases still compile and run, so this should be fine.

Cheers

CC @mathk


eriksvedang:
Cool!

PR #413, opened 2019-04-18
Add Identifier mangling for invalid identifier strings

scolsen:
If a C reserved word, for example, short or long, is used as an
identifier in carp, the resultant C code contains an invalid identifier
and fails to compile. This change adds a function for mangling reserved
words into valid C identifiers and includes two initial terms, short and
long.

We apply the string mangling after symbol/character mangling as the
string resulting from a successful character mangle is already a valid
identifier even if it contains a reserved word.


scolsen:
We should also mangle digits, or implement another means of invalidating them in identifier positions or converting them into valid identifiers. Starting an identifier with a digit in C is invalid, however, something like:

```clojure
(deftype Flag [4 Int])
```

is valid Carp, at present, and will result in a compilation failure when initialized. 

I gave a go at implementing the obvious solution, adding digits to the existing character mangling, but as of now it seems as though mangling is applied somewhere that makes a bunch of declarations invalid if its implemented this way. 

```
./out/main.c:2043:19: error: no member named 'member0' in 'struct Maybe__int::(anonymous at ./out/main.c:28:5)'
    instance.Just.member0 = member0;
    ~~~~~~~~~~~~~ ^
./out/main.c:2080:19: error: use of undeclared identifier '__TWO__FOUR_'; did you mean '__TWO__FIVE_'?
    String_delete(__TWO__FOUR_);
                  ^~~~~~~~~~~~
                  __TWO__FIVE_
./out/main.c:2078:13: note: '__TWO__FIVE_' declared here
    String* __TWO__FIVE_ = &_24; // ref
            ^
./out/main.c:2242:19: error: use of undeclared identifier '__ONE__EIGHT_'; did you mean '__ONE__NINE_'?
    String_delete(__ONE__EIGHT_);
                  ^~~~~~~~~~~~~
                  __ONE__NINE_
./out/main.c:2240:13: note: '__ONE__NINE_' declared here
```

It results in a bunch of errors as the resulting C is "_ONE__FIVE" (or something similar).

hellerve:
The change looks good to me! :+1:

I think the second thing should actually be a parse error: identifiers should always start with a non-digit! So, while this bug is maybe related, it is different. We can happily work on this together, though, if you’d like to find out how to fix it!

scolsen:
@hellerve I'd be delighted to help out with that! I've opened issue #414 to track it/have some further discussion.

hellerve:
Thanks for that!

eriksvedang:
Very nice patch, this will help a ton!

PR #417, opened 2019-04-21
Make all asserts optional

hellerve:
This PR makes all asserts conditional on non-optimization.

Cheers


eriksvedang:
Excellent, thanks!!!

PR #418, opened 2019-04-22
Option to disable compilation (just generate the C source)

babelchips:
From the carp REPL set the project's no-compile flag:
```(Project.config "no-compile" false)```

or

from the command-line:
```carp -b --no-compile myfile.carp```

### CODE REVIEW

File: src/Obj.hs [hunk 1]
```diff
@@ -558,6 +560,7 @@ instance Show Project where
             , "Search paths for 'load' command:\n    " ++ joinWith  "\n    " searchPaths
             , "Print AST (with 'info' command): " ++ if printTypedAST then "true" else "false"
             , "File path print length (when using --check): " ++ show filePathPrintLength
+            , "Don't compile:" ++ if noCompile then "true" else "false"
```
hellerve: Nit: Could we add a space at the end of the string?

docs/Manual.md
```diff
@@ -94,7 +94,8 @@ When invoking the compiler from the command line you can supply the following fl
 * ```--log-memory``` The executable will log all calls to malloc and free.
 * ```--optimize``` Removes safety checks (like array bounds access, etc.) and runs the C-compiler with the `-O3` flag.
 * ```--check``` Run the compiler without emitting any binary, just report all errors found (in a machine readable way).
-
+* ```--no-compile``` Don't run the compiler, just generate the C source.
+*
```
hellerve: Nit: could we remove this trailing asterisk? It’s adding an empty bullet point.

src/Obj.hs
babelchips: Sure thing.

docs/Manual.md
babelchips: Yes

babelchips:
Updates made commit: 7680ceb1dbdf7f8c8b12e62b890c1ce2f890e9ea

hellerve:
This change looks pretty good to me!

Any reason that you chose to use the negative (i.e. `projectNoCompile`) instead of the positive (i.e. `projectCore`)? It seems to line up with the command line argument, but it seems confusing to me that the boolean expresses the absence instead of the presence (a kind of inverted boolean). I don’t think it’s a problem per se, I was just wondering whether there was any particular reason for it.

babelchips:
Thanks.  Yes I did deliberate that.  I decided to go with the negative form to keep it separate from the setting for the compiler ```projectCompiler```.  I am willing to change it to something else, maybe even ```generateOnly``` or something similar?

hellerve:
That makes sense! If we want to keep the semantics and invert the boolean, something like `compileC` could be used.

It also occurs to me just now that as there are two compilers at play (Carp’s and a C compiler), the flag `--no-compiler` is a bit confusing, but that might just be me.

babelchips:
I think maybe to eliminate any confusion we should drop references to compiling and go with a command line flag of```--generate-only``` and a setting ```projectGenerateOnly```?  What do you think?

hellerve:
That sounds good to me! I suck at naming things and/or features, but I do think that this would make it much clearer.

babelchips:
From the carp REPL set the project's  ``generate-only`` flag:
```(Project.config "generate-only" true)```
or
from the command-line:
```carp -b --generate-only myfile.carp```

PR #419, opened 2019-04-24
Return errors on non symbol identifiers

### LINKED ISSUE #414: Identifiers starting with digits should throw a parse error
At present the Carp REPL will allow for definitions using digits as identifiers like the following:

```clojure
(deftype Flag [4 Int])
(def 4 4)
(defn 4 [x] x)
```

Declarations like the above are more often than not likely erroneous and unintentional. Each of the above declarations results in different behaviors. `(deftype Flag [4 Int])` generates invalid C identifiers and fails to compile, `(def 4 4)` seems to generate a parenthesis parsing error and causes the repl to exit/fail on (build). `(defn 4 [x] x)` defines a function that's not callable (so far as I can discern) but compiles just fine.

For simplification, and to help programmers avoid unwitting errors in declarations, Carp should throw a parse error when a digit is used as the first character of an identifier.  

@hellerve

scolsen:
Function definition identifiers cannot begin with digits. This PR contains changes to ensure identifiers (in defn, def, etc.) can only begin with valid characters (symbol Xobjs).

resolves #414

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -163,11 +163,13 @@ eval env xobj =
                Left err -> return (Left err)
 
         [defnExpr@(XObj Defn _ _), name, args@(XObj (Arr a) _ _), body] ->
-            if all isSym a
-              then specialCommandDefine xobj
-              else return (makeEvalError ctx Nothing ("`defn` requires all arguments to be symbols, but it got `" ++ pretty args ++ "`") (info xobj))
-            where isSym (XObj (Sym _ _) _ _) = True
-                  isSym _ = False
+            case (obj name) of
+              (Sym _ _) -> if all isSym a
+                           then specialCommandDefine xobj
+                           else return (makeEvalError ctx Nothing ("`defn` requires all arguments to be symbols, but it got `" ++ pretty args ++ "`") (info xobj))
+                        where isSym (XObj (Sym _ _) _ _) = True
+                              isSym _ = False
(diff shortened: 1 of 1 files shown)


src/Eval.hs
```diff
@@ -163,11 +163,13 @@ eval env xobj =
...
-            if all isSym a
-              then specialCommandDefine xobj
-              else return (makeEvalError ctx Nothing ("`defn` requires all arguments to be symbols, but it got `" ++ pretty args ++ "`") (info xobj))
-            where isSym (XObj (Sym _ _) _ _) = True
-                  isSym _ = False
+            case (obj name) of
+              (Sym _ _) -> if all isSym a
+                           then specialCommandDefine xobj
+                           else return (makeEvalError ctx Nothing ("`defn` requires all arguments to be symbols, but it got `" ++ pretty args ++ "`") (info xobj))
+                        where isSym (XObj (Sym _ _) _ _) = True
```
hellerve: We could constrain this even more, since the `SymPath` module part must be empty (you couldn’t use `Module.x`, for instance), but that would probably require a different error message and make this a tad more complex.

hellerve:
This looks pretty good to me!

src/Eval.hs
eriksvedang: OK, I'll merge this and let you add that separately later!

eriksvedang:
Excellent! There's still a bug for `let` statements and `fn`, no?

hellerve:
And `def` and `deftype`! I’m not sure we should close #414 just yet, but this is definitely a step in the right direction!

PR #420, opened 2019-04-24
Require defn symbols to be unqualified

hellerve:
This PR follows up on #419 and makes sure that all symbols put in the head of `defn`s are unqualified.

Cheers

CC @scolsen


scolsen:
Whoo! Thanks for the followup!

hellerve:
Thanks for making it possible!

eriksvedang:
```(+ user-experience 1)```

PR #424, opened 2019-04-27
Return errors on non unqualified symbol identifiers for def and deftype

scolsen:
I've added a few more checks to `Eval.hs` to ensure the identifiers passed to `def` and `deftype` members (e.g. `(deftype Name <[members Int]>)`) are unqualified symbols. This prevents the evaluator from allowing instances in which, for example, Nums are passed to defs, `(def 4 5)`, resulting in invalid C. 

@eriksvedang @hellerve Let me know what you think!

...also, let me know if there are other cases we should catch with this. If not, we can probably close up #414 once this looks ok!


hellerve:
This looks great! I think `let` bindings might be another case where we need to check!

PR #428, opened 2019-04-30
Clarify git loading situation in docs

hellerve:
This PR clarifies a little bit what’s necessary for a Git-based Carp package.

Cheers

### DIFF
--- a/docs/Libraries.md
+++ b/docs/Libraries.md
@@ -25,11 +25,21 @@ This line of configuration can be put into a `~/.carp/profile.carp` file to make
 
 ## Loading via git
 
-You can also load libraries via git in this way:
+You can also load libraries via Git like that:
 
-`(load "https://github.com/hellerve/anima.carp@master")`
+```clojure
+(load "git@github.com/hellerve:anima.carp@master")
+```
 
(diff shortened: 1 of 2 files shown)


docs/Libraries.md
```diff
@@ -25,11 +25,21 @@ This line of configuration can be put into a `~/.carp/profile.carp` file to make
...
+```
+
+This will download the [Anima](https://github.com/hellerve/anima) library to
+`~/.carp/libs/<library>/<tag>` and load the file `anima.carp` in it. To get a
+stable version of the library you should specify a git tag rather than
+`@master`.
 
-This will download the [Anima](https://github.com/hellerve/anima) library to `~/.carp/libs/<library>/<tag>` and load the file `anima.carp` in it. To get a stable version of the library you should specify a git tag rather than `@master`.
+Please note that currently only loading through SSH is supported. If you want to
+make a library reading for loading, either prepare a file that has the same name
```
eriksvedang: "reading for loading" sounds like a typo?

hellerve: Fixed! I meant `ready for loading`!

PR #432, opened 2019-05-03
Add Maybe.ptr and Maybe.from-ptr

### LINKED ISSUE #287: Sum types and error / NULL -handling
Carp doesn't have sum types / tagged unions etc. yet but I think it would be good idea to discuss if the standard way of handling errors will be based on them and whether we could use them to abolish most/all(?) usage of null on the Carp-side of things. Should we have things like Option a = Some a | Nothing or Result a = a | Error String/code in the standard library?

It would be worth pondering if we could use for example the Option-type to handle NULLs in the C-interface: autoconvert between Nothing(Carp) <-> NULL(C) so the resulting NULLs would need to be handled.

hellerve:
This PR adds `Maybe.ptr` and `Maybe.from-ptr`, which convert Maybe values to pointers (and `Nothing` to `NULL`). This shoudl fix #287, although it isn’t quite automatic conversion.

Test cases are included.

Cheers

### DIFF
--- a/core/Maybe.carp
+++ b/core/Maybe.carp
@@ -44,4 +44,18 @@ It is the inverse of [`just?`](#just?).")
         (match @b
           (Nothing) false
           (Just y) (= x y))))
+
+  (doc ptr "Creates a `(Ptr a)` from a `(Maybe a)`. If the `Maybe` was
+`Nothing`, this function will return a `NULL` value.")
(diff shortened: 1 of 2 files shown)


eriksvedang:
Sweet, will be interesting to see how this affects API:s when wrapping C libraries.

PR #439, opened 2019-05-10
Safe IO.read->EOF

### LINKED ISSUE #286: IO.read->EOF segfaults if the file does not exist
Minimal crashing example:
```
(defn main []
  (println* (IO.read->EOF "foobar")))
```

hellerve:
This PR fixes #286 by making the result of `IO.read->EOF` a `Result` type.

A regression test case is included.

Cheers


eriksvedang:
Awesome!

PR #442, opened 2019-05-13
Make Array module safer

GrayJack:
- Made Array.find wrapped in a Result (tests included, though I couldn't run it)
- Made Array.maximum and Array.minimum safe (tests included, though I couldn't run it)
- Made Array.index-of safe (tests included, though I couldn't run it)
- Added Pair.zero to maintain compatibility with Array.maximum and Array.minimum
- Also some typo fixes in Array module doc string

Please, try to run the tests on your machine

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,16 +58,16 @@ To build this example, save it to a file called 'example.carp' and load it with 
 [Veit Heller](http://veitheller.de) ([@hellerve](https://github.com/hellerve))
 
 ### Contributors
+* Markus Gustavsson
+* Fyodor Shchukin
 * Anes Lihovac
 * Chris Hall
-* Dan Connolly
-* Eric Shimizu Karbstein ([@GrayJack](https://github.com/GrayJack))
-* Fyodor Shchukin
-* Joel Kaasinen ([@opqdonut](https://github.com/opqdonut))
-* Jonas Granquist
-* Markus Gustavsson
-* Reini Urban
 * Tom Smeding
+* Dan Connolly
+* Reini Urban
+* Jonas Granquist
+* Joel Kaasinen ([@opqdonut](https://github.com/opqdonut))
+* Eric Shimizu Karbstein ([@GrayJack](https://github.com/GrayJack))
 
 Are you missing from the contributors list? Please send a pull request!
 
--- a/README.md
+++ b/README.md
@@ -58,15 +58,16 @@ To build this example, save it to a file called 'example.carp' and load it with 
 [Veit Heller](http://veitheller.de) ([@hellerve](https://github.com/hellerve))
 
 ### Contributors
-* Markus Gustavsson
-* Fyodor Shchukin
 * Anes Lihovac
 * Chris Hall
-* Tom Smeding
 * Dan Connolly
-* Reini Urban
-* Jonas Granquist
+* Eric Shimizu Karbstein ([@GrayJack](https://github.com/GrayJack))
+* Fyodor Shchukin
 * Joel Kaasinen ([@opqdonut](https://github.com/opqdonut))
+* Jonas Granquist
+* Markus Gustavsson
+* Reini Urban
+* Tom Smeding
 
 Are you missing from the contributors list? Please send a pull request!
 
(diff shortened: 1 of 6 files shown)


core/Array.carp
```diff
@@ -108,43 +110,46 @@ Returns `Nothing` if the array is empty.")
   (defn /= [a b]
     (not (= (the (Ref (Array a)) a) b)))
 
-  ; TODO unsafe!
-  (doc maximum "gets the maximum in an array (elements must support `<`).")
+  (doc maximum "gets the maximum in an array (elements must support `<` and `zero`)and wraps it in a `Just`.
+
+If the array is empty, returns `Nothing`")
   (defn maximum [xs]
```
hellerve: This implementation doesn’t quite work as expected. I suggest the following alternative implementation:

```clojure
  (doc maximum "gets the maximum in an array (elements must support `<`) and
wraps it in a `Just`.

If the array is empty, it returns `Nothing`.")
  (defn maximum [xs]
    (if (empty? xs)
      (Maybe.Nothing)
      (let-do [result (nth xs 0)
               n (length xs)]
        (for [i 1 n]
          (let [x (nth xs i)]
            (when (< result x)
              (set! result x))))
        (Maybe.Just @result))))
```

This would also keep the copying to a minimum!

eriksvedang:
It seems a little strange to need both `(zero)` and wrap in an `Either` for the maximum/minimum functions. Do you really need both to implement a safer version?

hellerve:
No, that’s why I updated the documentation in my reimplementation.

eriksvedang:
Oh, missed that. Sorry.

core/Array.carp
```diff
@@ -110,35 +110,33 @@ Returns `Nothing` if the array is empty.")
...
-              (set! result (Maybe.Just x))))))
-      result))
+  (defn maximum [xs]
+    (if (empty? xs)
+      (Maybe.Nothing)
+      (let-do [result (nth xs 0)
+             n (length xs)]
+        (for [i 1 n]
+          (let [x (nth xs i)]
+            (when (< result x)
```
hellerve: This comparison will have to be switched so that we get the minimum. Also the name will have to be switched :)

GrayJack: Sorry, I did that just after I wake up, maybe I was walking sleeping LOL

test/memory.carp
```diff
@@ -231,11 +231,11 @@
 
 (defn array-maximum []
   (let [xs [1 7 2 9 3 8 4 6 5]]
-    (assert (= 9 (maximum &xs)))))
+    (assert (= (Maybe.Just 9) (maximum &xs)))))
```
hellerve: We’ll have to use `(= &(Maybe.Just 9) &(maximum &xs))` here, because comparison is implemented for refs of `Maybe`

core/Statistics.carp
```diff
@@ -5,8 +5,8 @@
 (defmodule Statistics
   (deftype Summary [
     sum Double,
-    min Double,
-    max Double,
+    min (Maybe Double),
```
hellerve: If we change this interface, we’ll also have to update the Bench module.

eriksvedang:
Are you happy with the new solution @hellerve ? Since you wrote most of the library functions being affected by this I want your OK before merge:ing.

hellerve:
This looks pretty amazing to me now! Thank you so much for your relentless work, @GrayJack!

GrayJack:
I made all changes that I wanted, I'm ok with merging

eriksvedang:
Sorry, I had a final comment but I missed pressing the "Start review" button so this dragged out longer than needed. When that is fixed I'll merge immediately!

README.md
```diff
@@ -58,15 +58,16 @@ To build this example, save it to a file called 'example.carp' and load it with
 [Veit Heller](http://veitheller.de) ([@hellerve](https://github.com/hellerve))
 
 ### Contributors
-* Markus Gustavsson
```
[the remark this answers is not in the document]
GrayJack: Yeah, sorry, it was my personal plugging that sorts nom-numbered lists on markdown by default, I usually let it enable for my personal projects. Will fix that asap

eriksvedang:
Thanks!

PR #447, opened 2019-05-15
Only allow Symbols in Let bindings

scolsen:
This change ensures let bindings cannot have invalid identifiers, such as numbers `(let [4 4] true)`. 

I've also adopted guard syntax to handle potential failure conditions.

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -187,24 +187,26 @@ eval env xobj =
         (XObj The _ _ : _) ->
             return (makeEvalError ctx Nothing ("I didn’t understand the `the` at " ++ prettyInfoFromXObj xobj ++ ":\n\n" ++ pretty xobj ++ "\n\nIs it valid? Every `the` needs to follow the form `(the type expression)`.") Nothing)
 
-        [letExpr@(XObj Let _ _), XObj (Arr bindings) bindi bindt, body] ->
-          if even (length bindings)
-          then do bind <- mapM (\(n, x) -> do x' <- eval env x
-                                              return $ do okX <- x'
-                                                          Right [n, okX])
-                               (pairwise bindings)
-                  let innerEnv = Env Map.empty (Just env) (Just "LET") [] InternalEnv 0
-                  let okBindings = sequence bind
-                  case okBindings of
-                    (Left err) -> return (Left err)
-                    Right binds -> do
-                      let envWithBindings = foldl' (\e [XObj (Sym (SymPath _ n) _) _ _, x] -> extendEnv e n x)
-                                    innerEnv
-                                    binds
-                      evaledBody <- eval envWithBindings body
-                      return $ do okBody <- evaledBody
-                                  Right okBody
-          else return (makeEvalError ctx Nothing ("Uneven number of forms in `let`: " ++ pretty xobj) (info xobj)) -- Unreachable?
+        [letExpr@(XObj Let _ _), XObj (Arr bindings) bindi bindt, body] 
+          | odd (length bindings) -> return (makeEvalError ctx Nothing ("Uneven number of forms in `let`: " ++ pretty xobj) (info xobj)) -- Unreachable?
+          | not (all isSym bindings) -> return (makeEvalError ctx Nothing ("`let` identifiers must be symbols, but it got `" ++ concatMap pretty bindings ++ "`") (info xobj))
+          | otherwise -> 
+              do bind <- mapM (\(n, x) -> do x' <- eval env x
+                                             return $ do okX <- x'
+                                                         Right [n, okX])
+                              (pairwise bindings)
+                 let innerEnv = Env Map.empty (Just env) (Just "LET") [] InternalEnv 0
+                 let okBindings = sequence bind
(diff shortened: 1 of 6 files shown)


src/Eval.hs
```diff
@@ -187,24 +187,26 @@ eval env xobj =
...
-                    Right binds -> do
-                      let envWithBindings = foldl' (\e [XObj (Sym (SymPath _ n) _) _ _, x] -> extendEnv e n x)
-                                    innerEnv
-                                    binds
-                      evaledBody <- eval envWithBindings body
-                      return $ do okBody <- evaledBody
-                                  Right okBody
-          else return (makeEvalError ctx Nothing ("Uneven number of forms in `let`: " ++ pretty xobj) (info xobj)) -- Unreachable?
+        [letExpr@(XObj Let _ _), XObj (Arr bindings) bindi bindt, body] 
+          | odd (length bindings) -> return (makeEvalError ctx Nothing ("Uneven number of forms in `let`: " ++ pretty xobj) (info xobj)) -- Unreachable?
```
scolsen: It does seem like this is indeed unreachable. What's happening right now, as far as I can tell, is that any inner lets e.g. `(defn main [] (let [4 4] true))` are passed along to qualify, at which point the preceding example fails since the qualificatory function does not have exhaustive patterns. 

I think we somehow need to make a pass through `eval` on the inner forms before qualifying names? I'm a bit stumped on this one.

src/Eval.hs
```diff
@@ -187,24 +187,26 @@ eval env xobj =
...
+          | not (all isSym bindings) -> return (makeEvalError ctx Nothing ("`let` identifiers must be symbols, but it got `" ++ concatMap pretty bindings ++ "`") (info xobj))
```
hellerve: Shouldn’t this check if all even-indexed things are syms?

scolsen: ...doh -.-. great point. This may also be why I was encountering errors when trying to handle this in Qualify to handle inner lets. I'll make the fix!

scolsen: a9aa5b8 and 8a3b0af fix this. Thanks fo catching this!

src/Qualify.hs
```diff
@@ -51,19 +51,19 @@ setFullyQualifiedSymbols typeEnv globalEnv env (XObj (Lst [the@(XObj The _ _), t
...
-                       let qualified = setFullyQualifiedSymbols typeEnv globalEnv e o
-                       in (extendEnv e binderName s, bs ++ [s, qualified]))
-                    (innerEnv, []) (pairwise bindings)
-           newBody = setFullyQualifiedSymbols typeEnv globalEnv innerEnv' body
-       in  XObj (Lst [letExpr, XObj (Arr bindings') bindi bindt, newBody]) i t
-  else XObj (Lst [letExpr, bind, body]) i t -- Leave it untouched for the compiler to find the error.
+setFullyQualifiedSymbols typeEnv globalEnv env (XObj (Lst [letExpr@(XObj Let _ _), bind@(XObj (Arr bindings) bindi bindt), body]) i t) 
+  | odd (length bindings) = XObj (Lst [letExpr, bind, body]) i t -- Leave it untouched for the compiler to find the error.
+  | not (all isSym (evenIndicies bindings)) = XObj (Lst [letExpr, bind, body]) i t -- Leave it untouched for the compiler to find the error.
+  | otherwise = let Just ii = i
```
scolsen: Prior to this change attempting to define `(defn foo [] (let [4 4] true))` in the REPL caused a 'non-exhuastive patterns in lambda' error in `Qualify.hs` and crashed the REPL. This change exchanges that error for a new one: 
```
carp: Can't create let-binder for non-symbol: XObj {obj = Num Int 4.0, info = Just (Info {infoLine = 1, infoColumn = 20, infoFile = "REPL", infoDelete = fromList [], infoIdentifier = 4}), ty = Nothing}
  error, called at src/InitialTypes.hs:363:22 in CarpHask-0.2.0.0-5zBtEl5uNipJ0fDU65syPR:InitialTypes
```

Ideally, I think we should try and catch this in the REPL and report the error to the user without crashing, but I'm not exactly sure as to how to accomplish this. It seems like we'd need to put the inner let through another Eval pass? At the moment, if I'm understanding it correctly (and there's a chance I'm not!) I think it passes from Eval straight through Qualify and then to `visit` in `InitialTypes`--or something like that. Any ideas?

scolsen: @hellerve @eriksvedang

eriksvedang: Hmm, yeah this one was tricky. The `initialTypes` function detects some type errors and have the ability to return TypeError, so perhaps that's where it should be reported? So keep your code but add a proper type error instead of the call to `error` in `extendEnvWithLetBindings`? Or am I missing something?

scolsen: Thanks for the tips! I've added an error to catch this and it seems to work swimmingly, just need a quick double check to make sure the code is free of fishiness (all puns intended).

eriksvedang:
Awesome work, thank you!

PR #449, opened 2019-05-19
Catch double definitions in let

### LINKED ISSUE #320: let can generate invalid C
When the same variable is bound twice in the same `let` expression, Carp generates invalid C by declaring the variable twice in the same scope. There are multiple ways to go about this; my personal favorite is disallowing it alltogether.

Example program:
```clojure
(defn main []
  (let [x 1
        x 2]
    (IO.println &(str x))))
```

Interestingly, this is not an issue with nested `let` bindings:
```clojure
(defn main []
  (let [x 1]
    (let [x 2]
      (IO.println &(str x)))))
```

This is because every `let` opens a new C scope by adding `{}` around the expression, and C is okay with redeclaring variables in an inner scope.

Cheers

hellerve:
This PR fixes #320 by analyzing `let` bindings for duplicates and returning an error message if there are any.

A test case is included.

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -199,15 +199,19 @@ eval env xobj =
                  let okBindings = sequence bind
                  case okBindings of
                    (Left err) -> return (Left err)
-                   Right binds -> do
-                     let envWithBindings = foldl' (\e [XObj (Sym (SymPath _ n) _) _ _, x] -> extendEnv e n x)
-                                   innerEnv
-                                   binds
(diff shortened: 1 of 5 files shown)


eriksvedang:
Seems like this needs some changes now that I merged #447

hellerve:
Fixed! Hopefuly my rebase didn’t mess anything up, but the tests should still run at least.

eriksvedang:
Very nice!

PR #452, opened 2019-05-20
Fix docs syntax of git url for loading libraries

doublec:
The syntax for loading git libraries didn't work for me on Linux. Changing it to that in this commit works.


hellerve:
Yup, that looks like a typo to me!

PR #460, opened 2019-05-26
Macros: Add unit macro.

scolsen:
Adds a unit macro, a convenience macro for returning () in
side-effecting functions. This is useful, for example, for match forms
in which only one match should result in a side effect. The other
matches still require forms for completeness, but they should simply
return (). Unit might be easier to read in these cases, as parens can
get lost in the sea of surrounding parens in the definition.

I'm not totally convinced of the value of this macro, so please feel free to let me know if it seems redundant and unneeded, or if you have other ideas! Here's an example use-case:

```clojure
;; Without using `unit`                                                                            
    (sig add-propagator (Fn [Cell a] ()))                                                                            
    (defn add-propagator [cell propagator]                                                                           
      (let [props (Cell.propagators cell)]                                                                           
      (match (index-of props propagator)                                                                             
          (Nothing)                                                                                                  
            (do (Cell.update-propagators cell (push-back props propagator))                                          
                (alert-propagator propagator))                                                                       
          (Just x) ()))) ;; do nothing case must return unit to match the sig. Straight up '()' is a little easy to miss here and could be interpreted as an error at first glance.

;; Using `unit`                                                                           
    (sig add-propagator (Fn [Cell a] ()))                                                                            
    (defn add-propagator [cell propagator]                                                                           
      (let [props (Cell.propagators cell)]                                                                           
      (match (index-of props propagator)                                                                             
          (Nothing)                                                                                                  
            (do (Cell.update-propagators cell (push-back props propagator))                                          
                (alert-propagator propagator))                                                                       
          (Just x) (unit)))) ;; Our intentions here are a little bit clearer when we use a macro. We can be confident that doing nothing on this branch is intentional, and not a mistake or something to be filled in later.
```

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -284,6 +284,13 @@
 (defmacro ignore [form]
   (list 'let (array '_ form) (list)))
 
+;; Also expose this as 'void' for those familiar with C?
+(doc unit "Convenience macro for specifing a return type of unit '()' for
+side-effecting functions. Can be slightly clearer than plain () in certain
+contexts, though this form is more verbose.")
+(defmacro unit []
+  (list))
+
 ;; Allows inclusion of C headers relative to the Carp file in which this macro is called.
 (defmacro relative-include [file]
   (list 'local-include


eriksvedang:
Huh, interesting idea! I really like the clarity of it, though I'm not a fan of adding too many ways to write the same thing. Let me ponder it a bit before merge:ing...

eriksvedang:
OK, I'm done thinking. I think it's worth adding this and we'll see if people use it.

There seems to be some bug with the doc generation that chokes on the code in this PR though – it hangs indefinitely :( 

```carp ./docs/core/generate_core_docs.carp -b```

So that needs to be fixed before we can merge... sorry about that!

**Comment by scolsen** [comment]
Shoot! I’ll take a look, thanks!

**Comment by eriksvedang** [comment]
Debugging infinite loops in Haskell is always interesting...

**Comment by scolsen** [comment]
Huh, so I can't seem to reproduce the infinite hang on this branch (using either the carp built on this branch or master carp), it gens the docs just fine...which caused me to notice a tangential issue—the macro docs aren't even generated! At present `Macros.carp` doesn't expose a module (aside from Dynamic), nor is the file loaded in the doc generation script, so the doc strings for top level macros in that file aren't saved to any html. 

Even if we load "Macros.carp" into the doc generation script, the docs aren't saved, since `save-doc` requires a module. So we can either stick all the top-level macros in `Macros.carp` into a module, or we can change `save-docs` to also handle docs for top level forms.

At any rate, I can't reproduce the hang even when I update the code so that the macro docs are actually saved (which I've tested out by sticking a subset of the macros into a module, those after the Dynamic module is defined)!

I'm kind of clueless as to the cause of the hang you're experiencing...is it possible that when you run `carp ./docs/core/generate_core_docs.carp -b` carp is looking at some other `core` directory not on the branch (this has tripped me up since I keep a separate copy of the carp repo for use, which my `CARP_DIR` is set to, and one for tinkering)? If not that, maybe it's some hairy system specific issue?

**Comment by hellerve** [comment]
FWIW, I can reproduce the hanging on Mac, which is what @eriksvedang uses as well, without changing any of my setup (my dev and prod Carp are one and the same).

**Comment by eriksvedang** [comment]
Oh yeah, the problem with the lack of module is a known one, I'll create an issue for it!

If you can't reproduce I guess it's up to me and @hellerve to figure this one out. Perhaps it's something with case insensitivity or similar macOS file system weirdness.

**Comment by eriksvedang** [comment]
Issue opened here: https://github.com/carp-lang/Carp/issues/486

**Comment by eriksvedang** [comment]
It seems like it is the name `unit` that messes it up. Haven't figured out why though...

**Comment by scolsen** [comment]
Huh! This is quite interesting. I'm on mac as well and it works just fine on my end...I'm able to generate docs for the unit macro without an issue. So it sounds more likely I have some weird setup I'm not aware of that's somehow subverting the problem.

When I get a chance later I'll post my OS version and the rest of my setup. Really strange that I'm not encountering the hang.

**Comment by scolsen** [comment]
> It seems like it is the name `unit` that messes it up. Haven't figured out why though...

That sounds reasonable since `unit` is a special term...possibly the name trips up the parser or something in XObjs?

Might also be some kind of environment collision? I could see that being the case if we already define () as 'unit' in the top level env and defining this macro tries to bring in another definition of 'unit' into the top level env --  since the macros aren't in a module --thus the compiler might enter an infinite recursion because we'd wind up having defined 'unit' as 'unit' since (list) === () === 'unit'? (This is just a very sketchy guess at what might be happening with my limited knowledge, but an accidental infinite lazy recursion sounds like it might be the culprit here).

Are you guys able to avoid the hang if you tuck the `unit` definition into a separate module?

File: core/Macros.carp [hunk 1]
```diff
@@ -284,6 +284,13 @@
 (defmacro ignore [form]
   (list 'let (array '_ form) (list)))
 
+;; Also expose this as 'void' for those familiar with C?
+(doc unit "Convenience macro for specifing a return type of unit '()' for
```
hellerve: Nitpick: could we limit this to 80 chars per line?

hellerve:
Putting `unit` in a module—I tried `Foo`—or renaming it to `void` both work for me!

core/Macros.carp
scolsen: Good catch! Will do.

scolsen:
> Putting `unit` in a module—I tried `Foo`—or renaming it to `void` both work for me!

Awesome! So yeah, I'm guessing there's likely some kind of collision going on in the environment when we call this macro `unit`. If that's in fact the issue, this should conceivably happen whenever we define anything called `unit` at the top level (unless types, funcs, etc are handled differently/injected into the env differently)?

I'm cool with naming it `void` instead. So I think we can:

* Fix whatever is causing the hang--presumably something in the compiler?
* rename this `void`
* Stick it in a module (but we should figure out whether other macros in here belong in a module or not, and what said module should be called, and potentially if we should move these moduled macros into a seperate file, like utilities.carp or extras.carp or something, and leave the other important macros in here defined at the top-level)
* drop the macro altogether (this is still an option, since it's more or less just a minor convenience thing)

core/Macros.carp
scolsen: Ah, actually, I'm missing something here--I'm seeing these come in at col 73 and 74

eriksvedang:
I think that fixing the hang would be the best choice, or someone else will trip up on this or something very similar. I'm currently working on another infinite loop bug so I can investigate this one at the same time to see where the loop is. That should give us enough info to fix the bug, hopefully.

core/Macros.carp
hellerve: Huh, somehow I’m now also seeing these as split lines. I dunno what happened with my brain there.

scolsen: No sweat!

hellerve:
Whatever happened to this? It seems like there are a bunch of weird errors with infinite loops in the evaluator so it might be worth consolidating: is this still an issue here, and is the issue that you—@eriksvedang—talked about in your last comment fixed?

eriksvedang:
I did spend some time trying to pinpoint it but never managed to find a fix :/

hellerve:
By using profiling and grepping I found that we bind a variable in `core/Bench.carp` to `unit` in a let statement, and Carp then tries to use the macro symbol inside that let statement, leading to an error.

You can reproduce that error for instance using this statement:

```clojure
(defn main [] (let [unit (+ 2 1)] unit))
```

Also note that this seems to work for any macro, though only macros without arguments actually trigger the infinte loop. If we use for instance `cond` instead it gets really weird:

```clojure
> (defn main [] (let [cond (+ 2 1)] cond))l
Can't find symbol 'xs' at /Users/veitheller/Documents/Code/Github/carp/carp/core/Macros.carp:138:18.
```

`xs` is in fact a macro variable.

I haven’t gotten further debugging this, but it’s already strange and delightful.

eriksvedang:
Thanks for looking into this!

Your example with `(defn main [] (let [unit (+ 2 1)] unit))` doesn't trigger an error for me though, should it?

hellerve:
Does it not trigger an error if you have the `unit` macro defined? Then (and only then) it should trigger the error.

eriksvedang:
Ah, right. I forgot this was in relation to a PR, sorry...

eriksvedang:
A have a feeling we should convert this PR to an issue...

hellerve:
Issue #659 should reflect this.

eriksvedang:
OK, will close this for now then.

PR #461, opened 2019-05-27
Remove StringCopy.append

hellerve:
This PR removes `StringCopy.append` and replaces all instances of it by `String.append`, which is always a better idea.

It also removes it from `Todo.org` and the Language Guide.

Cheers


eriksvedang:
Awesome, this one felt good to merge!

PR #462, opened 2019-05-27
Make map’s higher order functions take refs

hellerve:
This PR makes all the higher-order functions  in `Map` and `Set` be non-owning.

Test cases were updated.

Cheers


eriksvedang:
Good stuff.

PR #463, opened 2019-05-27
Correctly classify sumtypes with single case

hellerve:
This PR fixes a bug where sumtype definitions with a single case didn’t work. It works now, and a regression test has been added.

Technical aside: the problem was that the classification of “regular deftypes” and sumtypes was the number of forms in the definition, which doesn’t word. Instead, we now look at the length and if the element in question is an array. If it is, then it’s a regular type, otherwise it’s a sumtype.

Cheers


eriksvedang:
Much better, thanks for fixing this!

PR #482, opened 2019-06-03
Emit member deletion on destructive setter

### LINKED ISSUE #481: Mutating setter doesn't delete the previous member.
https://github.com/carp-lang/Carp/blob/714aeab80caf59a8332199b67b172d32258dd834/src/Deftype.hs#L147-L155
I'm not sure whether it's intentional behaviour, but it may cause a memory leak.

hellerve:
This PR fixes #481 by emitting member deletion on the mutating setter template.

Cheers


eriksvedang:
Oops. Glad this one is fixed, for sure!

PR #483, opened 2019-06-04
Concretize: Fix top level lambda naming

scolsen:
This change adjusts the way in which function nest levels are
incremented to ensure top level lambdas are assigned the correct nest
level. Top level lambdas used to use their expressions as names,
resulting in invalid c. A check on nest levels in the lambda naming code
now prevents this.

Essentially, this change makes the following sorts of definitions in the REPL ok:

```clojure
[(fn [] ())]
(deftype Foo [f (Fn [] ())])
(Foo.init (fn [] ()))
```

Note that this *doesn't* cover sumtypes, which still can't emit valid C for lambdas. Looks like there's some other portion of the compiler that needs to be modified for Sumtypes to work with lambda/function members.

Please let me know about any stylistic preferences! I just used what Haskell I know to implement this, happy to reimplement in a style more idiomatic for the code base.

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -71,7 +71,7 @@ concretizeXObj allowAmbiguityRoot typeEnv rootEnv visitedDefinitions root =
                                   functionEnv argsArr
              Just funcTy = t
              allowAmbig = isTypeGeneric funcTy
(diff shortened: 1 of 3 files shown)


eriksvedang:
Awesome work, thank you!!!

Issue #294, opened 2018-09-18
Emit `extern`?


roman01la:
Hello! 👋 I'm experimenting with compiling Carp to WASM (it works!), could we have a syntax to emit `extern` annotation somehow? This way you export functions/vars into JS world from compiled wasm code. For now I manually edit emited C code :(

Would be great if someone could drop a quick guide on where too look in the source, so I could try to implement this and learn Carp's internals.

Thanks for amazing work!

hellerve:
Thanks for your interest in Carp!

Late in 2017, I wrote a—now a little outdated—[overview of how the compiler works](https://blog.veitheller.de/The_Carp_Compiler_(as_of_2017).html) on my blog. Short answer: the C is generated in `src/Emit.hs`.

I also experimented with WASM for a bit before (see #178). I dropped it, because I ddidn’t end up having time for it; it would be great if you find the time and interest to actually power through!

hellerve:
This is basically fixed by #484!

PR #484, opened 2019-06-04
Add annotations

hellerve:
This PR adds a mechanism for “annotations”, which is what I call the prefixes that are often used in C by certain libraries—`GLFWAPI` and `EMSCRIPTEN_KEEPALIVE` spring to mind—to signal some kind of special behaviour of the function.

This is important for things like #294 where we need to emit it for the compiler.

A few adjustments to the emitter needed to be made to make sure it honours these annotations, but otherwise this change is relatively trivial, and extensible too!

In pretty great news, it now enables us to do things like this:

```clojure
(system-include "emscripten/emscripten.h")
(Project.config "compiler" "emcc")
(Project.config "title" "example.wasm")
(Project.config "output-directory" ".")
(add-cflag "-s ONLY_MY_CODE=1")

(defn dosth [i z] (Int.+ i z))
(annotate dosth "EMSCRIPTEN_KEEPALIVE")
```

which will emit a standalone WASM file that exports a function called `dosth`! It doesn’t yet handle polymorphic functions, sadly, because they don’t seem to inherit the meta data of their parent, and I couldn’t find where this happens currently.

Anyway, it’s a step into exciting new territory: running Carp in a browser natively!

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -32,6 +32,12 @@
 (defmacro hidden? [name]
   (list 'not (list '= () (list 'meta name "hidden")))) ;; TODO: This is buggy, will report true when meta is set to 'false'!
 
+(defndynamic annotate-helper [name annotation]
+  (cons annotation (meta name "annotations")))
+
+(doc annotate "Add an annotation to this binding.")
+(defmacro annotate [name annotation]
+  (list 'meta-set! name "annotations" (annotate-helper name annotation)))
 
 (defmodule Dynamic
   (defndynamic caar [pair] (car (car pair)))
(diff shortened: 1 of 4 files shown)


eriksvedang:
Cool! This needs a rebase now, it seems.

The problem with polymorphic function inheriting meta data is probably because `concretize`works with XObj:s and not Binders... so that's annoying. Shouldn't be too hard to fix, I hope – only a lot of drudge work.

hellerve:
Yeah, I had to make a lot of functions work with binders already, which was a little annoying.

I can try and fix that!

hellerve:
Rebased!

eriksvedang:
@hellerve Great, let me know how it goes. Binders are probably the correct unit of abstraction in a lot of cases since they contain more information...

PR #485, opened 2019-06-06
Implement zero for Maybe

hellerve:
This PR implements `zero` for `Maybe` and `Result`. This is important for interoperability with other functions.

- `zero` for `Maybe` returns `Nothing`.
- `zero` for `Result` returns `(Success (zero))`.

This is the best I could come up with, although the design is a little questionable. Arguably, `zero` for `Maybe` could also be `(Just (zero))`. I’ll let my peers be the judge of that (by which I mean everyone who reads this PR).

Test cases are included.

Cheers

### DIFF
--- a/core/Maybe.carp
+++ b/core/Maybe.carp
@@ -66,4 +66,7 @@ It is the inverse of [`just?`](#just?).")
    (if (null? a)
      (Nothing)
      (Just @(Pointer.to-ref a))))
+
+  (doc zero "returns `Nothing`.")
+  (defn zero [] (Nothing))
 )
--- a/core/Result.carp
+++ b/core/Result.carp
@@ -106,6 +106,9 @@ It is the inverse of [`success?`](#success?).")
         (match @b
           (Success _) false
           (Error y) (= x y))))
+
+  (doc zero "returns `(Success (zero))`.")
+  (defn zero [] (Success (zero)))
 )
 
 (defmodule Maybe
--- a/core/Result.carp
+++ b/core/Result.carp
@@ -106,9 +106,6 @@ It is the inverse of [`success?`](#success?).")
         (match @b
           (Success _) false
           (Error y) (= x y))))
-
-  (doc zero "returns `(Success (zero))`.")
-  (defn zero [] (Success (zero)))
 )
 
 (defmodule Maybe
--- a/docs/core/Maybe.html
+++ b/docs/core/Maybe.html
@@ -503,6 +503,26 @@ a value using <code>zero</code> if a <code>Nothing</code> is passed.</p>
                 </p>
             </div>
             <div class="binder">
+                <a class="anchor" href="#zero">
+                    <h3 id="zero">
+                        zero
+                    </h3>
+                </a>
+                <div class="description">
+                    defn
+                </div>
+                <p class="sig">
+                    (λ [] (Maybe a))
+                </p>
+                <pre class="args">
+                    (zero)
+                </pre>
+                <p class="doc">
+                    <p>returns <code>Nothing</code>.</p>
+
+                </p>
+            </div>
+            <div class="binder">
                 <a class="anchor" href="#zip">
                     <h3 id="zip">
                         zip
--- a/docs/core/Result.html
+++ b/docs/core/Result.html
@@ -599,26 +599,6 @@
 
                 </p>
             </div>
-            <div class="binder">
-                <a class="anchor" href="#zero">
-                    <h3 id="zero">
-                        zero
-                    </h3>
-                </a>
-                <div class="description">
-                    defn
-                </div>
-                <p class="sig">
-                    (λ [] (Result a b))
-                </p>
-                <pre class="args">
-                    (zero)
-                </pre>
-                <p class="doc">
-                    <p>returns <code>(Success (zero))</code>.</p>
-
-                </p>
-            </div>
         </div>
     </body>
 </html>
--- a/docs/core/Result.html
+++ b/docs/core/Result.html
@@ -599,6 +599,26 @@
 
                 </p>
             </div>
+            <div class="binder">
+                <a class="anchor" href="#zero">
+                    <h3 id="zero">
+                        zero
+                    </h3>
+                </a>
+                <div class="description">
+                    defn
+                </div>
+                <p class="sig">
+                    (λ [] (Result a b))
+                </p>
+                <pre class="args">
+                    (zero)
+                </pre>
+                <p class="doc">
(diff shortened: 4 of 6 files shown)


bjorn3:
> Arguably, zero for Maybe could also be (Just (zero)).

That means that it maybe shouldnt be implemented at all.

hellerve:
@bjorn3 that’s true, except that it’s an important function for a lot of use cases—it’s basically the identity of the type.

Both are reasonable identities, but that doesn’t mean implementing this interface is useless.

I’m currently writing a blog post, about the builder pattern in Carp, for instance. Imagine having a builder that allows you to create a type bit by bit instead of all at once, filling in the elements one by one. This is obviously useful. Ideally it would also allow you to omit fields, and fill them with `zero` values. Having a default for `Result` and `Maybe` means that both of those types can be used as members of a type eligible for a builder.

I hope this example clarifies where I’m heading with this.

hellerve:
Relatedly, his would be equivalent to the `Default` trait in Rust, in a way, which returns `None` for `Option` and is undefined for `Result`.

scolsen:
Awesome! The builder use case sounds cool! Having an identity value defined on types is a prerequisite for implementing recursion schemes over those types (another use-case!), which means this gets us a step toward `Maybe` recursion schemes :) cata, apo, hylo and co. possibly coming soon (meaning if I have the time and can recall how to do it)

swizzard:
does carp have ((plans to implement) something equivalent to) newtypes? something like `Just (zero)` seems akin to the `First`/`Last` newtypes for `Maybe`, whereas `Nothing` seems more like the default `Maybe` instances

hellerve:
@swizzard not for 0.3. (due July) or 1.0 (a long way away), which are the only two releases we have roadmaps for. In general, we’re trying to keep the type system as small as possible—but sometimes expressiveness is more important!

If you have an idea how newtypes could look, feel, and behave in Carp, feel free to open a proposal!

eriksvedang:
Zero for `Just` being `Nothing` seems spot on (like it being `[]´ for Array). Perhaps we should leave the `Result` undefined, since it seems like the difficult one. I can't recall any computer science answer to this, but perhaps it's something we could ask around for :)

eriksvedang:
On the other hand, if you need it for your article @hellerve, that implies usefulness. In that case the suggested implementation for `Result` seems fine. Of course, one simple solution is to keep the definition outside of Core. But if more people will need it, it's probably better to have it centrally defined...

hellerve:
I don’t exactly need it for the article, so I guess I’d just want to leave it for congruence. On the other hand, noone else seems to define it, which seems to be an indication that maybe `Result` doesn’t have a single identity—which makes sense, because it basically has two independent ones.

scolsen:
I think the `Result` case is very similar to ints in that we have two possible options for identity, just like we have `0` as identity for addition and `1` as identity for multiplication.'...so it depends on your use-case.

In Haskell this is problematic for implementing monoid on numbers, since we have two valid options. They get around this by providing additional types that implement monoid for the appropriate operation and select the appropriate identity `Sum` and `Product`.

Maybe we can do something similar here? In carp we have the power of macros, which can make our approach more flexible. We could write a macro that gives the user the option to select the identity they want to use for `Result` and replaces calls to `Result.zero` with the appropriate identity. 

Then again that may be going overboard.

hellerve:
Oof, that sounds super powerful but like a UX nightmare if we do it wrong. I’m happy to implement this (I’m a macro-monger, after all), but I’m not going to be the one who figures out how to make this usable.

scolsen:
There is a definite danger of confusion with that approach...one could conceivably define the identity on `Result` willy-nilly according to one's needs and throw all expectations out the window. I'm not totally sure what this would look like in practice, I can't envision whether it'd actually be problematic or not. 

My intuition says that the identity of the type being variable in a bunch of places would break a lot of expectations and thus be bad. Then again, if the macro is something like `(zero Success)` or `(zero Error)` it's fairly clear at point of use...just not sure whether that leads to usable APIs or whether it leads to painful dives into implementation detail at every corner.

In my own experience Haskell's `Sum` and `Product` types are mildly annoying to use, but it works, and it's not unclear. If we can achieve a fidelity similar to that I think we're fine.

Then again, sometimes the best implementation is no implementation ¯\_(ツ)_/¯

jacereda:
Shouldn't `zero` be `Just(zero)` so you can fold a structure with `+` using an initial value of `zero`?

jacereda:
Maybe there could be something like `sum-zero` for that...

scolsen:
> Shouldn't `zero` be `Just(zero)` so you can fold a structure with `+` using an initial value of `zero`?

That's an interesting idea! It would also be a valid design choice, I think.

I prefer `Maybe.zero = Nothing` because `Maybe a` isn't really a recursive type so its  unlikely `Maybe` would be folded over in common usage, and I think `Maybe.Zero = Nothing` better captures the notion of identity for defining monoidal/binary operations on `Maybe`.

It might be common however, to fold over a recursive structure (e.g. List) that contains Maybes. This can be accomplished with `Maybe.zero = Nothing`. 

```
+ Just x Nothing = Just x
+ Nothing Just y = Just y
+ Just x Just y = Just x + y
+ Nothing Nothing = Nothing 

fold + Maybe.zero [Just 1 Just 2 Just 3 Just 4]
```

Using `Maybe.Zero = Just(zero)` actually amounts to the same thing as the above.

You can define fold on `Maybe` itself but it has to be its projection followed by an "f-algebra" that describes whatever binary operation you want to perform. The scenario of having multiple nested maybes is probably rare in practice. The example below is essentially the `Maybe` projection combined with an f-algebra for computing the binary operation. At which point this `Maybe` is properly a fixpoint--it expresses recursion in general but is not a proper container by itself (it requires an algebra to map its structure to a meaningful values/other types). 

```
project-and-algebra :: Int -> Maybe Int -> Int 
project-and-algebra init Nothing = 0
project-and-algebra init Just x = init + 1
fold project 0 (Just (Just (Just (Just Nothing)))))
```

eriksvedang:
Going through old PR:s... perhaps we should merge this? :)
I'll let you decide on what instances to use @hellerve, I think your initial suggestion works well enough but if someone managed to convince you in the comments, I'm sure that will be good too!

hellerve:
I will remove `zero` for `Result` because I no longer think that that type has an identity. I stand behind `zero` for `Maybe` being `Nothing`, though.

I’ll rebase as well!

hellerve:
Done!

eriksvedang:
Yay!

PR #492, opened 2019-06-13
Compile with link-time optimizations, use standard NDEBUG instead of …

jacereda:
…OPTIMIZE.


jacereda:
The NDEBUG bit should make it easier to integrate with other projects.

The LTO part makes executables smaller by stripping unneeded symbols.

hellerve:
I think getting rid of `OPTIMIZE` is a good idea.

Abou the other C flag: were you aware that `(add-cflag "-flto")` would also be possible? Maybe that would be a better idea, so that people that don’t want this optimization for whatever reason can leave that out?

jacereda:
Right, removed the LTO flag.

eriksvedang:
Excellent, thanks!

PR #497, opened 2019-06-14
Sdl unused

jacereda:


eriksvedang:
What do you mean with it being unused?

jacereda:
`SDL_INIT_EVERYTHING` is just a XOR of all the initializable subsystems. As is, it isn't doing anything.

jacereda:
It would be useful if the library provided `SDL_Init()`.

eriksvedang:
Oh, OK :)

eriksvedang:
It does wrap `SDL_init` though.

jacereda:
Oh, I missed that. I’ll fix it later. Should it wrap the individual values or just the EVERYTHING?

PR #498, opened 2019-06-14
Add some const qualifiers, removed MALLOC() null checks, fixed Float_…

jacereda:
…modf(), use memcpy() instead of memmove().

### DIFF
--- a/core/carp_bool.h
+++ b/core/carp_bool.h
@@ -1,7 +1,7 @@
 #include "carp_stdbool.h"
 
 // Bool
-bool Bool_copy(bool* b) {
+bool Bool_copy(const bool* b) {
   return *b;
 }
 
--- a/core/carp_char.h
+++ b/core/carp_char.h
@@ -20,17 +20,12 @@ char Char_from_MINUS_int(int i) {
   return (char)i;
 }
 
-char Char_copy(char *c) {
+char Char_copy(const char *c) {
   return *c;
 }
 
-String PtrChar_str(char *c) {
+String PtrChar_str(const char *c) {
     size_t len = strlen(c) + 1;
     String ptr = CARP_MALLOC(len);
-
-    if (ptr == NULL) {
-      return NULL;
-    }
-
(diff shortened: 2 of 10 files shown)


jacereda:
I guess removing the NULL checks might be controversial. The rationale is that it's never checked in lots of other places, it adds a lot of noise to static analysis and we have tons of memory.

hellerve:
I’d just like to object to:

> and we have tons of memory

Carp has been used to do work on Eurorack modules that have 16k of RAM + 32k of memory ([these guys](https://www.st.com/en/microcontrollers-microprocessors/stm32f301.html)). There, `malloc` might fail relatively easily; at this rate, returning `NULL` might be our best option. Ideally, I’d opt for fixing the places where it isn’t checked, even if it’s more work.

jacereda:
I've done some embedded programming and in those cases the best thing to do is not to allocate at all and work with static buffers. And I always prefer a crash at the right place than hiding the bug and crashing later.

But in any case, I guess this could be wrapped in a CHECKED_MALLOC() macro for most uses. That could be another PR.

hellerve:
(Adding @trentgill to this conversation in case he has opinions about this)

`CARP_MALLOC` is already a macro, so maybe redefining it to be checked if desirable might be interesting.

Crashing early is great, but undefined behaviour usually isn’t. I’d understand calling `abort()` et al. if allocation fails, but moving on and trying to invoke `memcpy` makes me a little unhappy—I know we do it in other places because of oversights and/or laziness, but that shouldn’t be an excuse. Would you be open to that instead or is there anything that speaks against that?

hellerve:
Bottom line: I’m okay with us merging this, and I’d like to contribute a checked version of malloc unless you want to. Is that okay?

trentgill:
No objections from me. I always add the NULL checks in my programs, but usually out-of-mem issues are found intuitively. Having to pass the failure back through the call stack usually means it never happens anyway.

hellerve:
Cool, so I think we have an agreement on your approach, @jacereda!

Issue #494, opened 2019-06-13
Maybe test failure


jacereda:
I get the following output for the maybe.carp test:

```
Test 'nothing? works on Nothing' passed
Test 'nothing? works on Just' passed
Test 'just? works on Just' passed
Test 'just? works on Nothing' passed
Test 'from works on Just' passed
Test 'from works on Nothing' passed
Test 'unsafe-from works on Just' passed
Test 'apply works on Just' passed
Test 'apply works on Nothing' passed
Test 'or-zero works on Just' passed
Test 'or-zero works on Nothing' passed
Test 'to-result works on Just' passed
Test 'to-result works on Nothing' passed
Test 'ptr works on Just' failed:
	Expected value: '1', actual value: '0'
Test 'ptr works on Nothing' passed
Test 'from-ptr works on Ptr/Just' passed
Test 'from-ptr works on NULL/Nothing' passed
Results:
	|=================|
	Passed: 16	Failed: 1
[RUNTIME ERROR] '"./out/Untitled"' exited with return value 1.
```

jacereda:
On macOS, if that matters.

jacereda:
I think the problem resides here:

```
int* Maybe_ptr__int(Maybe__int a) {
    int* _15;
    if(a._tag == Maybe__int_Nothing_tag) {
        Maybe__int _5_temp = a;
        _15 = NULL;
    }
    else if(a._tag == Maybe__int_Just_tag) {
        Maybe__int _5_temp = a;
        int x = _5_temp.Just.member0;
        _15 = &x;
    }
    else {
      // This will not be needed with static exhaustiveness checking in 'match' expressions:
      fprintf(stderr, "Unhandled case in 'match' expression at line 58, column 5 in '/Volumes/Data/carp/Carp/core/Maybe.carp'\n");
      exit(1);
    }
    return _15;
}

```

Note how it's taking the address of the out-of-scope variable `x`.

hellerve:
I can’t reproduce this error; I’m also not sure `x` is out of scope, since it’s a member of `a` being passed back into the scope where `a` is defined. Am I missing something?

I’m also on OS X btw.

jacereda:
I'm probably using a newer clang, since I'm using nixpkgs.
`_15` contains the address of the stack-allocated variable `x`, so it's returning the address of a stack variable. That's bad.

hellerve:
What version of Clang are you using?

jacereda:
```
clang --version
[...]

PR #500, opened 2019-06-14
Remove Maybe.ptr for now

hellerve:
This PR removes `Maybe.ptr` for now, because as discovered in #494 it leads to a memory error. That’s a little sad, since it removes the dual to `from-ptr`, and the isomorphism is gone. But we’ll figure out how to make it work again, I’m sure!

Cheers


hellerve:
Superseded by #502

Issue #469, opened 2019-05-29
Matching should work on references


hellerve:
Matching is the most integral part of sumtypes. Sadly, using `match` currently takes ownership of the value, meaning that we can’t reuse it. This is annoying, since it leads to a lot of copies in `core/Maybe.carp` and `core/Result.carp`.

Matching should be possible on references. Enabling that has interesting lifetime consequences, though, since currently the matching clause gets full control over the members of the sumtype, which would be a potentially dangerous operation when using references. We might have to change the API to rely on references on the members as well, but that requires some experimenting.

Cheers

eriksvedang:
This is fixed in branch `match-on-refs` which will be merged soon.

eriksvedang:
It's in!

PR #502, opened 2019-06-16
Add Unsafe.coerce

### LINKED ISSUE #494: Maybe test failure

jacereda:
Used for Maybe.ptr, fixes #494.

### DIFF
--- a/core/Maybe.carp
+++ b/core/Maybe.carp
@@ -55,9 +55,10 @@ It is the inverse of [`just?`](#just?).")
   (doc ptr "Creates a `(Ptr a)` from a `(Maybe a)`. If the `Maybe` was
 `Nothing`, this function will return a `NULL` value.")
   (defn ptr [a]
-    (match a
-      (Nothing) NULL
-      (Just x) (address x)))
+    (the (Ptr a) (Unsafe.coerce
+      (match @a
+        (Nothing) NULL
+        (Just _) (the (Ref (Maybe a)) a)))))
 
  (doc from-ptr "Creates a `(Maybe a)` from a `(Ptr a)`. If the `Ptr` was
 `NULL`, this function will return `Nothing`.")
--- a/src/StartingEnv.hs
+++ b/src/StartingEnv.hs
@@ -302,6 +302,28 @@ templateEnumToInt = defineTemplate
                         ,"}"])
   (const [])
 
+-- | The Unsafe module contains dangerous functions
+unsafeModule :: Env
+unsafeModule = Env { envBindings = bindings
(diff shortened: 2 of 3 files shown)


core/Interfaces.carp
```diff
@@ -4,6 +4,8 @@
 ;;(definterface copy (λ [&a] a))
 ;;(definterface str (λ [a] String))
 
+(definterface coerce (λ [a] b))
```
hellerve: This is only needed if there are different implementations for this for different types. Is that going to be how this is used? My understanding was that `Unsafe.coerce` kind of stood on its own.

core/Maybe.carp
```diff
@@ -55,9 +55,10 @@ It is the inverse of [`just?`](#just?).")
   (doc ptr "Creates a `(Ptr a)` from a `(Maybe a)`. If the `Maybe` was
 `Nothing`, this function will return a `NULL` value.")
   (defn ptr [a]
-    (match a
-      (Nothing) NULL
-      (Just x) (address x)))
+    (the (Ptr a) (Unsafe.coerce
+      (match @a
```
eriksvedang: There are plans to make matching work on refs very soon, this would use that – right?

core/Interfaces.carp
jacereda: Oh, right. Fixed.

core/Maybe.carp
jacereda: Can I read about that somewhere?

eriksvedang: https://github.com/carp-lang/Carp/issues/469

jacereda: Yeah, I guess the implementation would change then.

PR #504, opened 2019-06-17
Optionally check malloc

hellerve:
This PR introduces `Dynamic.check-allocations`, which sets a flag to check all calls to `malloc` and raise `SIGABRT` if they fail.

Built in response to #498.

Cheers


eriksvedang:
Sweet!

PR #508, opened 2019-06-20
Stronger types for Color

jacereda:


jacereda:
How can I implement `to-int` cleanly? Also, I thought things like `str` were automatically derived?

eriksvedang:
There's no support for doing `to-int` at the moment but structs have a `get-members` dynamic function. Perhaps adding something similar for sumtypes could help with implementing this?

`str` and `prn` should indeed be automatically generated...

eriksvedang:
This works for example:

```clojure
鲮 (defmodule Color (deftype Id Black Red))
鲮 (type Color.Id.str)
Color.Id.str : (λ [(Ref Id)] String)
鲮 (str (ref (Color.Id.Red)))
(Red)
``

**Comment by jacereda** [comment]
So, I tried to figure out how to use `get-members` and failed. Is it documented somewhere?

**Comment by eriksvedang** [comment]
Oh, so actually it's called `members`, my bad.

You use it like this:

```clojure
鲮 (deftype Foo [x Int y String])
鲮 (members Foo)
=> [(x Int) (y String)]
```

jacereda:
So, you propose something like `constructors` for sum types? Sounds nice, but for this use case it would require a lookup in the returned elements... Isn't something like `get-tag` simpler/faster?

eriksvedang:
I'm fine with merge:ing this as-is, the hack to dispatch on the enum works well for now I think. Implementing an automatic, reusable `to-int` for enums could be done separately, and wouldn't be too hard in my mind at least.

eriksvedang:
@jacereda Do you want to implement a `get-tag` method for sumtypes? Should be fairly straight forward (and I'll happily assist). Otherwise I'll do it eventually...

jacereda:
I'll give it a try.

PR #509, opened 2019-06-20
Enabling address sanitizer yields warnings due to _FORTIFY_SOURCE red…

jacereda:
…efinition.


eriksvedang:
I'm not sure I understand this bug, what do you to do trigger it?

jacereda:
I get it when running `bench/array_resizing.carp`. The `clang` in nixpkgs defines `_FORTIFY_SOURCE`. I know it's possible to disable it in a nixpkgs package, so this isn't very important if it doesn't affect the normal clang distribution in macOS.

PR #510, opened 2019-06-20
Fix string assert, simpler/faster conditions.

jacereda:
The current code would break at the added test in `assert((i+lsrc-1) < linto)`.

### DIFF
--- a/core/carp_string.h
+++ b/core/carp_string.h
@@ -22,20 +22,13 @@ void String_delete(String s) {
 }
 
 void String_string_MINUS_set_BANG_(String *s, int i, char ch) {
-    int l = strlen(*s);
-    (void)l;
-    assert(i >= 0);
-    assert(i < l);
+    assert((size_t)i < strlen(*s));
     (*s)[i] = ch;
 }
 
 void String_string_MINUS_set_MINUS_at_BANG_(String *into, int i, const String *src) {
     char *dest = (*into) + i;
     int lsrc = strlen(*src);
-
-    int linto = strlen(*into);
-    (void)linto;
-    assert(i >= 0);
     /* given a string and indicies
      *
      *  0 1 2 3 4 5 6 7 8 9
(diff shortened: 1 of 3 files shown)


core/carp_string.h
```diff
@@ -22,20 +22,13 @@ void String_delete(String s) {
 }
 
 void String_string_MINUS_set_BANG_(String *s, int i, char ch) {
-    int l = strlen(*s);
-    (void)l;
-    assert(i >= 0);
-    assert(i < l);
+    assert((size_t)i < strlen(*s));
```
hellerve: Should we not check for non-negativity here as well?

jacereda: `size_t` is unsigned, so it shouldn't be needed. There might be problems with that assert for near-4GB strings, but the solution to that would be to use size_t for string sizes/offsets (that would be another PR).

hellerve: I guess, but since we’re passing in an `int` from Carp, you might end up with errors that aren’t super useful, right? I mean I guess if it fails it fails, but if I pass in `-1` I’d still be happier if my assert failed on `i >= 0` than `i < strlen(*s)`.

jacereda: I guess both assertions are useless because you don't know what the value is. `i` could be negative just because it's an overflow of a previous operation. Let me think about this, maybe it would be better to have something built on top of `assert` to display values.

jacereda:
The CHK_INDEX macro would be useful in some other places, I'll leave that for another PR if it's accepted.

core/core.h
```diff
@@ -2,15 +2,30 @@
...
 #include <windows.h>
 #endif
 
 typedef char* String;
 typedef char* Pattern;
 
+#if defined NDEBUG
+#define CHK_INDEX(i,n)
+#else
+#define CHK_INDEX(i,n)				    \
```
hellerve: I really like this macro! :)

PR #522, opened 2019-07-08
More generic for macro

jacereda:
This should allow looping backwards (as long as the step is a constant) and using floating point types.


eriksvedang:
Looks good to me! Thanks!

PR #523, opened 2019-07-09
Fix vector lerp.

jacereda:
Looks like `lerp` for vectors is wrong. I named the plain number interpolator `xlerp` because if I named it `lerp` I got the following error:

```

I found an unresolved generic type `(λ [(Ref (Vector2 r159)), (Ref (Vector2 Double)), Double] (Vector2 Double))` for the expression `XObj {obj = Sym Vector2.lerp (LookupGlobal CarpLand AFunction), info = Just (Info {infoLine = 78, infoColumn = 19, infoFile = "/Volumes/Data/carp/Carp/test/vector2.carp", infoDelete = fromList [], infoIdentifier = 452}), ty = Just (λ [(Ref (Vector2 r159)), (Ref (Vector2 Double)), Double] (Vector2 Double))}` at line 78, column 19 in '/Volumes/Data/carp/Carp/test/vector2.carp'

```

### DIFF
--- a/core/Interfaces.carp
+++ b/core/Interfaces.carp
@@ -77,3 +77,6 @@
 
 (defn not-null? [p]
   (not (null? p)))
+
+(defn xlerp [from to amount]
+  (+ from (* (- to from) amount)))
--- a/core/Vector.carp
+++ b/core/Vector.carp
@@ -91,8 +91,8 @@
 
   (doc lerp "Linearly interpolate between the two vectors a and b by amnt (between 0 and 1).")
   (defn lerp [a b amnt]
-    (init (* (- @(x b) @(x a)) amnt)
-          (* (- @(y b) @(y a)) amnt)))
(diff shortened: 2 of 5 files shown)


hellerve:
It looks like there are no implementations of `xlerp` yet?

jacereda:
> It looks like there are no implementations of `xlerp` yet?

It isn't an interface, just a generic function. Functions at the end of `Interfaces.carp` should probably be moved to a `Generics.carp` file or something like that.

eriksvedang:
I think that actually defining `lerp` as an interface might help with the initial error... Would be nice to have all functions use the same name and use an interface, so that other code can depend on the function in a generic fashion.

jacereda:
Maybe another approach could be to leave `lerp` as a generic function and have `vlerp` as another generic function on vectors. I'll explore this one in a separate PR.

PR #524, opened 2019-07-14
Offline dependencies

hellerve:
This PR introduces a special case to dependency management: if the dependencies were already downloaded and we’re on a tag or commit (so, in `Detached HEAD` state), we can safely proceed to inline the dependencies without redownloading them. This finally re-enables offline builds when the dependency is already there.

Cheers


eriksvedang:
Awesome!!!

PR #525, opened 2019-07-15
Lambda copy generated bad code.

jacereda:
Not very proud of the patch though, maybe there's something in the `concreteCopy` parameters that can be used to determine if it's a pointer?


eriksvedang:
Looks good to me, I think this is a clear way to do it.

eriksvedang:
And thanks a ton for fixing this pesky bug that I had totally missed..!

PR #527, opened 2019-07-22
Better error messages in fmt

hellerve:
This PR introduces better error messages for `fmt` by telling the user which directive needs an argument and which argument needs a directive if they don’t match.

The directive printer is a little wonky, currently we will only display the first letter (which means it works for `%s` but not for `%.2f`). If you have better ideas—short of writing a parser for all kinds of directives—, please let me know!

Cheers


eriksvedang:
Neat, looks good enough to me!

PR #536, opened 2019-08-27
Macros: Add dynamic list to array transforms

scolsen:
This change adds two utility functions to the Dynamic module for
transforming list forms into arrays and array forms into lists.

Using this function, one can easily convert lists and arrays for
ease of symbol manipulation, for example:

```clojure
(Symbol.join (list-to-array (list 'a 'b 'c)))
```

This comes in handy, for instance, when defining macros, as certain
forms in Carp expect arrays in certain places while others expect
list--this enables the programmer to flexible convert between the two as
needed.

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -92,21 +92,16 @@
         acc
         (list-to-array-internal (cdr xs) (append acc (array (car xs))))))
 
-  (doc list-to-array "Transforms a list dynamic data literal into an array dynamic data
-  literal, preserving order")
-  (defndynamic list-to-array [xs]
-    (list-to-array-internal xs (array)))
- 
-  (defndynamic array-to-list-internal [xs acc]
-    (if (= 0 (length xs)) 
-        acc 
-        (array-to-list-internal (cdr xs) (append acc (list (car xs))))))
-  
-  (doc array-to-list "Transforms an array dynamic data literal into a list dynamic data
-  literal, preserving order.")
-  (defndynamic array-to-list [xs]
+  (defndynamic collect-into-internal [xs acc f]
+    (if (= 0 (length xs))
+        acc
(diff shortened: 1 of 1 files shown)


hellerve:
Personally, I’d prefer something like `(collect-into (list 'a 'b 'c) array)` UX-wise. I couldn’t get that to work earlier (I think because `array` is a special form rather than a function), but that’s what my ideal world would look like.

eriksvedang:
Do you guys want to modify this now that #539 is in, or should I just merge this? (I'm fine with either)

hellerve:
This should work now:

```clojure
 (defndynamic collect-into-internal [xs acc f]
    (if (= 0 (length xs))
        acc
        (collect-into-internal (cdr xs) (append acc (f (car xs))) f)))

  (doc collect-into "Transforms a list dynamic data literal into an array dynamic data
  literal, preserving order")
  (defndynamic collect-into [xs f]
    (collect-into-internal xs (f) f))

(collect-into '(1 2 3) array)
'(collect-into [1 2 3] list)
```

scolsen:
Sehr fantastisch, @hellerve! If you have time, would you like to clone and push your changes to my branch (scolsen:array-list-dynamics)? That way the history reflects your contribution accurately? If you're too busy, I can snag your implementation and add it to the PR, but I'd really love it if you got the proper cred in the commits!

hellerve:
@scolsen You can definitely snatch it! If you really want to cred me, just `@` me in the commit.

Cheers

scolsen:
Fixed it up! Made one small change to your macro @hellerve, so that users don't have to remember to quote list results.

hellerve:
Looks great!

eriksvedang:
Yay!

PR #538, opened 2019-09-02
Streamline macro-log

### LINKED ISSUE #528: macro-log duplicate
Currently we register both `commandPrint` and `commandMacroLog` as `macro-log`. One treats strings specially and one accepts multiple arguments. Which of the two should we prefer?

My personal preference is fusing the two: both unwrapping strings and accepting multiple arguments!

Cheers

hellerve:
This PR fixes #528 by streamlining `macro-log` and removing the duplicate. Specifically, you can now write stuff like

```clojure
(macro-log "Hiya number " 2 "!")
```

In other words, it’s multi-arity and type-generic now!

Cheers

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -479,12 +479,6 @@ commandProject args =
      liftIO (print (contextProj ctx))
      return dynamicNil
 
--- | Command for printing a message to the screen.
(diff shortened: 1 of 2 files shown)


eriksvedang:
Great!

PR #539, opened 2019-09-02
Make list and array commands

hellerve:
Thsi PR makes `list` and `array` commands rather than special forms. They can then be used in dynamic higher order functions (relevant for #536, for instance).

Cheers


eriksvedang:
That's a lot better!

Issue #516, opened 2019-07-06
Can't compile using Vector


jacereda:
If I try to compile:

```
(load "Vector.carp")
(defn main []
  0)
```

`Maybe__bool` ends up being undefined (only `Maybe__int` is defined).

```
[nix-shell:~/src/Carp/Carp]$ cabal new-run carp -- -b /tmp/test.carp
Up to date
./out/main.c:699:25: error: unknown type name 'Maybe__bool'
void Maybe_delete__bool(Maybe__bool p);
                        ^
./out/main.c:2015:25: error: unknown type name 'Maybe__bool'
void Maybe_delete__bool(Maybe__bool p) {
                        ^
./out/main.c:2016:16: error: use of undeclared identifier 'Maybe__bool_Just_tag'
  if(p._tag == Maybe__bool_Just_tag) {
               ^
./out/main.c:2019:21: error: use of undeclared identifier 'Maybe__bool_Nothing_tag'
  else if(p._tag == Maybe__bool_Nothing_tag) {
                    ^
4 errors generated.
carp: callCommand: clang -fPIC -lm -g ./out/main.c -o "./out/Untitled"  -I.//core/   (exit 1): failed

[nix-shell:~/src/Carp/Carp]$ 
```

jacereda:
Oops, looks like it has to do with some local changes I have made in the `Vector` module. I'll try to isolate the problem.

jacereda:
I reduced it to:

```
(defn angle-between [a b]
    (Maybe.Just (zero)))
(defn anti-parallel? [a b]
    (Maybe.apply (angle-between a b) (fn [x] (= x pi))))

(defn main [] 0)
```

That should yield the same errors.

hellerve:
It seems like the specialized type 'Maybe__bool' is not being emitted. Do you know from the top of your head what the conditions for a type being emitted are, @eriksvedang?

hellerve:
I think I know what the issue is. It seems as if `fn` forms aren’t actually annotated like
`defn`s are. I’m not quite positive about this part, but I’m guessing that leads to the
types that the `fn` depends on not being put in the type environment. @eriksvedang does
this sound like it makes sense?

eriksvedang:
Yeah sounds about right. Should I fix it?

hellerve:
That’d be great. I’d love to see how that’d work, but I’m not sure I could fix it myself.

eriksvedang:
OK, I’ll give it a go then!

[...]

PR #541, opened 2019-09-08
Working on generics.

jacereda:
- Added Generics module.
- Some Geometry/Vector* functions hardcoded parameters to Double.

### DIFF
--- a/core/Generics.carp
+++ b/core/Generics.carp
@@ -1,0 +1,106 @@
+;; The following generic functions make use of the interfaces
+
+(defmodule Generics
+  (defn one [] (inc (zero)))
+  (defn minus-one [] (dec (zero)))
+  (defn two [] (inc (one)))
+  (defn three [] (inc (two)))
+  (defn minus-two [] (* (two) (minus-one)))
+  (defn four [] (+ (two) (two)))
+  (defn five [] (+ (four) (one)))
+  (defn ten [] (+ (five) (five)))
+  (defn hundred [] (* (ten) (ten)))
+  (defn thousand [] (* (hundred) (ten)))
+  (defn hundred-thousand [] (* (hundred) (thousand)))
+  (defn million [] (* (thousand) (thousand)))
+  (defn minus-four [] (* (four) (minus-one)))
+  (defn twice [x] (* (two) x))
+  (defn halved [x] (/ x (two)))
+  (defn squared [x] (* x x))
+  (defn inverse [x] (/ (one) x))
+  (defn half [] (halved (one)))
+  (defn half-pi [] (halved pi))
+  (defn tau [] (twice pi))
+  (defn ninety [] (* (two) (* (* (three) (three)) (five))))
+  (defn one-hundred-eighty [] (* (two) (ninety)))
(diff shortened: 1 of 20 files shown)


jacereda:
Is there some way to have generic literals, so we can get rid of all those constants in `Generics`?

hellerve:
Yeah, that’s what I’ve been fighting when I tried to make the vector namespace generic as well back in the day (see also #373). I think your solution with incrementing `zero` and working from there is kind of ingenious, honestly, even if it’s a bit clunky.

I’d also like a unified number literal, and have the inference figure out which number is appropriate in the end. This sounds like a pretty big but worthwhile change to me.

eriksvedang:
Right now I think that would require setting the type of literals to unconstrained type variables which would be annoying in a lot of places. So we would need some kind of grouping (a la "type classes") to not lose the information about them being "numeric" (and "string like", etc). I think your solution here is pretty awesome though, and I'm perfectly happy merge:ing this as-is. But yes, would be nice to be able to write the literals of course. Perhaps we can start a new issue and discuss that there, and I'll merge this after I'm done reviewing it.

core/Vector.carp
```diff
@@ -253,37 +273,35 @@
...
+  (doc vlerp "Linearly interpolate between the two vectors a and b by amnt (between 0 and 1).")
+  (defn vlerp [a b amnt]
+    (zip (fn [a b] (lerp a b amnt)) a b))
 
-  (doc lerp "Linearly interpolate between the two vectors a and b by amnt (between 0 and 1).")
-  (defn lerp [a b amnt]
-    (Maybe.apply (zip - b a)
-      (fn [x] (init @(n a) @(v &(zip- * &(Array.replicate @(n a) &amnt)
-                           (v &x)))))))
+  (defn workaround-516 []
```
jacereda: Note this horrible workaround for #516, this should be removed when that one gets fixed.

hellerve: It doesn’t seem like this code is used anywhere. Am I reading this right as just being there to appease the compiler?

If so, this is extremely weird.

jacereda: Just remove that and try to run `test/vector3.carp`, you'll hit #516. I tried to fix it but my limited knowledge of the compiler internals didn't help.

core/Generics.carp
```diff
@@ -0,0 +1,106 @@
...
+The margin of error is 0.00001.")
+  (defn approx [a b]
+    (approx-margin a b (small)))
+
+)
+
+
+;; These are defined in the global scope.
+
+(use Generics)
```
eriksvedang: Since you're using the `use` command in the global scope, the Generics module will be included from here on, carrying into the user's session.
The best way to solve this is probably to use the `with` command instead, like this:
```
(with Generics
  <forms...>)
```
This will avoid the problem!

eriksvedang: Also, this feature might be completely undocumented... yikes :/

scolsen: > Also, this feature might be completely undocumented... yikes :/

I think it is! I saw it when digging through the source yesterday and couldn't recall any references to it in docs. Seems super useful!

PR #542, opened 2019-09-09
Add gensym

hellerve:
This PR adds a very simple version of `gensym`.

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -323,3 +323,12 @@ Otherwise it will fail with the message `msg`.
 The expression must be evaluable at compile time.")
 (defndynamic const-assert [expr msg]
   (if expr () (macro-error msg)))
+
+(doc *gensym-counter* "is a helper counter for `gensym`.")
+(defdynamic *gensym-counter* 1000)
+
+(doc gensym "generates symbols dynamically as needed.")
+(defndynamic gensym []
+  (do
+    (defdynamic *gensym-counter* (inc *gensym-counter*))
+    (Symbol.join ['gensym-generated- (Symbol.from *gensym-counter*)])))
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -327,8 +327,12 @@ The expression must be evaluable at compile time.")
 (doc *gensym-counter* "is a helper counter for `gensym`.")
 (defdynamic *gensym-counter* 1000)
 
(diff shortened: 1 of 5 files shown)


scolsen:
Very cool! Do we have a means of documenting commands yet?

hellerve:
I don’t think so. If I try to call `(info Symbol.from)` the REPL tells me that `Symbol.from` is unbound, which is a bug I guess.

scolsen:
What if you try `(info Sym.from)` -- that shouldn't really make a difference since we bind the module to `Symbol` (so that should just work), but the module name is "Sym" in the environment--perhaps that's an issue? (contrarily the String and Project modules are both named "String" and  "Project" in the environment, and bound to those strings in Dynamic.

```haskell
dynamicSymModule :: Env
dynamicSymModule = Env { envBindings = bindings
                       , envParent = Nothing
                       , envModuleName = Just "Sym"
                       , envUseModules = []
                       , envMode = ExternalEnv
                       , envFunctionNestingLevel = 0 }
  where bindings = Map.fromList [ addCommand "join" 1 commandSymJoin
                                , addCommand "prefix" 2 commandSymPrefix
                                , addCommand "from" 1 commandSymFrom
                                ]
```

hellerve:
Neither of those work, and changing the module name to `Symbol` doesn’t seem to do anything either. Bummer.

hellerve:
I renamed the symbol to make the documentation consistent.

While doing that I realized that using `Dynamic.Symbol.from` in `info` works—d’oh!—, but documenting it still doesn’t seem to save anything.

eriksvedang:
Please open an issue on the documentation problem!

eriksvedang:
There is a slight potential for bugs here since the counter resets on `(reload)`, which could lead to multiple symbols getting the same number in some rare cases (I think). But the file Macros.carp is reloaded very early so it shouldn't be a problem in reality, plus the symbols only need to be unique in local scope, right? So I think we're good. Just important to think about, since the bugs related to this are nasty :) The bugs when not using `gensym` are even nastier though, so I'm very happy you added this!

hellerve:
The symbols need to be unique in whatever scope they are defined in (I want it to work in global scope as well if necessary), but I think the cases in which values introduced by `gensym` are duplicated are rare enough to justify existence.

It’s definitely important to think about it, though!

PR #548, opened 2019-09-10
Implement get-tag for sum types.

jacereda:
I tried also to simplify the code a bit, but maybe those deps were there for some future purpose...


eriksvedang:
The deps are a relic from copying the code from the Deftype module, perhaps I can remove them there too now actually...

PR #549, opened 2019-09-11
Add Dynamic map

scolsen:
This PR adds a `map` dynamic function to the Dynamic module. `map` takes some function and one or more forms and returns a `list` dynamic data literal containing the result of applying the function to each form.

Here are some examples of it's usage:

```clojure
'(map symbol? 'a 'b c)                                                   
=> (true true false) ;; In this case (since unquoted) 'c' refers to the `c` function!                                                       
(collect-into (map length '[a b c] '[d] '[e f]) array)                    
=> [3 1 2]                                                                
'(map car '(Maps of the world) '(Are never lacking) '(Great expanses))    
=> (Maps Are Great) 
```

## Some Questions and Alternatives

- Why map?

It's sometimes useful to apply a function to a whole bunch of forms and use the resulting forms in some meaningful way. For example, one might want to pull the first item out of several forms one is processing to e.g. ensure each one of them begins with a `definterface` or conforms with some other rule (length, whatever).

- Why not have `collect-into` take an `f`?

This is definitely a valid alternative design, but I kind of like how simple `collect-into` is at present, this function composes well with it.

---

Let me know what you think!


eriksvedang:
Very useful! But I think it should take lists as arguments and use the variadicity (?) for the number of args to the function being mapped, i.e:
```clojure
(map + [1 2 3] [4 5 6]) ;; => [5 7 9]
```

scolsen:
Ah interesting, that seems closer to traditional `zipWith`, and that might be more useful!

I can see how having `map` take a `:rest` argument is somewhat confusing, since `map` traditionally only works on lists--the idea here is that the list formed by the `:rest` arg is the list it's operating on, but that's not so clear when you're just reading something like: `map symbol? 'a 'b 'c` (makes it look as though it's not operating on a list and thus isn't really `map` in the traditional sense, even though it is under the hood)

What about providing something like:

```clojure
;; arguments to `map` must be an explicit list or array form
(map symbol? '(a b c))
;; => (true true true)
(map symbol? [a b c])
;; => [true true true]
(map symbol? 'a 'b 'c)
;;=> error

;; emulates traditional zipWith but on forms
(zip-with + [1 2 3] [4 5 6])
;; => [5 7 9]
;; we can also make it super generic by using a rest arg 
;; and checking that each arg is a list or array e.g.
(defn add-triadic [x y z] (+ (+ x y) z))
(zip-with add-triadic [1 2 3] [4 5 6] [7 8 9])
;; [12 15 18]
(zip-with add-triadic [1 2 3] [4 5 6])
;; macro-error

;; the original `map` proposal, `apply-to` maps some function 
;; over an arbitrary number of provided forms 
;; by taking a rest argument and mapping over the list built up of :rest args 
(apply-to symbol? 'a 'b 'c)
;; => (true true true)
(apply-to length '[a b c] '[d e f])
;; => (3 3) ;; since arrays are symbols.
;; or if we want to be more verbose but more explicit we can call it `apply-to-forms`
(apply-to-forms symbol? 'a 'b 'c)
;; or if we want to be less verbose and less explicit 
;; and let module context convey our meaning, we can call it `apply`:
(apply symbol? 'a 'b 'c)
;; or we can make its meaning a bit more concrete
;; transform--has `form` in the name
(transform-each symbol? 'a 'b 'c)
```

eriksvedang:
Having to quote the whole form (and it still evaluating!) is a bug btw, I've filed an issue #555.

scolsen:
Alrighty, I've updated the implementation to make it similar to clojure and other lisps. I wound up having to implement a couple of utility functions, so I've described them in detail (with code samples) below:

## `empty?`

Returns true if a data literal is empty, false otherwise

```clojure 
(defndynamic empty? [xs]
    (if (= 0 (length xs))
      true
      false))
```

This one is fairly straightforward. I just found it tedious to write length checks on lists, and this makes it a bit faster to do that.

## `and-internal`

For the implementation of `zip` (aka `zip-with`), I needed an easy means of checking whether or not any of the forms provided as arguments were empty, since by zip's semantics, we stop processing as soon as we exhaust one of the provided lists.

```clojure
(defndynamic and-internal [x y]
    (if x y false))
```

It's precisely the same as the existing `and` macro, only it returns the result instead of the if form.

The and macro doesn't work for this case because it returns a form. In this case I needed to evaluate the result eagerly since I had to reduce a list of booleans using some boolean combinator--replacing the booleans with forms was no good.

## `or-internal`

The reasoning . behind implementing this function was the same as that of `and-internal`.  Ultimately, this is the one I use in the implementation of zip, since we should cease processing if *any* of the arguments are empty.

```clojure
(defndynamic or-internal [x y]
    (if x true y))
```

**Note: Arguably, we should move both `or-internal` and `and-internal` into the Bool module (and probably just call them `and` and `or`). However, I wanted to keep everything scoped to Dynamic until we have some discussion on it.**

## `reduce`

The clearest semantics for checking whether or not any of the forms provided to `zip` are empty that I could think of was reducing over an application of map to check whether the lists are `empty?`. Let me know if there's a better way to do this, and we can excise this function from the Dynamic module.

```clojure
(defndynamic reduce [f x xs]
    (if (empty? xs)
      x
      (reduce f (f x (car xs)) (cdr xs))))
```

## `apply`

`zip` is a bit trickier than map since we need to build the application form before applying the provided function. We first need to fetch the first member of each provided form and build up an application of `f`--so we need `apply` here. Apply, just like @hellerve referenced earlier, behaves as it does in other lisps. Given a function name and a list of arguments, it returns the result of a function application using those arguments.

```clojure
(defndynamic apply-internal [argument-list acc]
      (if (empty? argument-list)
          acc
          (apply-internal (cdr argument-list) (append acc (list (car
          argument-list))))))

  (doc apply
    "Applies the function `f` to the provided argument list, passing each value
    in the list as an argument to the function.")
  (defndynamic apply [f argument-list]
    ;; The let clause here is a tad mysterious at first glance. When passed a
    ;; standalone function name (i.e. not an application (f x), carp evaluates
    ;; it into the function's defining form, e.g. foo becomes (defn foo [x] x),
    ;; commands such as + become (command +) etc.  ;; The binding here accounts
    ;; for that case, allowing users to pass the function name to apply
    ;; unquoted.
    ;;
    ;; This is necessary for parity across map-internal, zip, and apply.
    ;; Since map calls its function directly, it takes it as is. Apply, on the
    ;; other hand, would have to take the quoted form, since it builds a list
    ;; that serves as the actual application.
    ;;
    ;; This is problematic for the user facing map function, since it makes
    ;; calls to map or zip (which uses apply) as appropriate--unless we support
    ;; the quoted function name argument in map-internal or the unquoted one in
    ;; apply, we can't use zip and map-internal in map.
    ;;
    ;; TODO: Once issue #555 is fixed, we shouldn't need to call (list 'quote...
    ;; here
    (let [function-name (list (cadr f))]
      (apply-internal argument-list function-name)))
```

As you can see from the comment, this one's a tad tricky since given a standalone binding name, e.g. `+` and *not* an application `(+ 1 2)` carp returns the defining form `(command +)`. This means we can't simply use the `f` argument as is when building up the function application in `apply`, since it will evaluate to e.g. `(command +)` or `(defn blah [] 2)`--which would result in an erroneous form such as: `(apply + '(1 2)) => ((command +) 1 2)` or `(apply blah '(1 2)) => ((defn blah [] 2 ) 1 2)`

There are two ways around this issue:

1. Enforce that the user pass only quoted function names as arguments to `apply` and all the functions that depend on it.
2. Adapt apply to accept unquoted function names.

I thought the second option was better for three reasons:
1. It's nicer for the user, as they don't have to remember to quote the function name.
2. The implementation of map, since it can apply the function directly, does not require quoting the function name. The module API as a whole is easier to modify and use if we can keep things consistent, since `map-internal` can take unquoted function names, why not `apply`?
3. It's extra cognitive load and code for implementers of other functions that use `apply` to recall that it requires a quoted function name. This would enforce the same requirement on their function's arguments, and possibly force them to implement code checking that the function name is in fact quoted (e.g. `(= (car (list? f)) quote)`. 

All the above points considered, I felt it was a nicer api to just write the brief let binding that pulls the name out of the functions declaration (returned upon encountering the naked function name). 

## `zip`

Finally, as we discussed, to emulate the behavior of `map` in other lisps, we need an implementation of zip to handle the arguments > 1 case. 

```clojure
(defndynamic zip-internal [f forms acc]
    (if (reduce or-internal false (map-internal empty? forms (list)))
        acc
        (zip-internal
          f
          (map-internal cdr forms (list))
          (let [result (list (apply f (map-internal car forms (list))))]
            (append acc result)))))

  (defndynamic zip [f :rest forms]
    (zip-internal f forms (list)))
```

There's a current limitation with this implementation in that it's results *must* be quoted. Presently, the quoted result of running zip is, e.g.:

```clojure
'(zip + '(1 2 3) '(4 5 6))
=> ((+ 1 4) (+ 2 5) (+ 3 6))
```

Which is all well and good, and what we expect when calling `apply`.

However, call this unquoted, and the following error occurs:

```clojure
(zip + '(1 2 3) '(4 5 6))
=> 
I can’t match the types `Int` and `(λ [t3, t5] t0)` within `(+ 1 4)`.

  1 : Int
  At line 1, column 10 in 'REPL'

  Expected first argument to '+' : (λ [t3, t5] t0)
  At no info
```

Bizarre, no? Strangely enough, calling other functions to force the evaluation of the resulting list members works as expected:

```clojure
(car (zip + '(1 2 3) '(4 5 6)))
=> 5
```

I'm not sure what the source of this error/behavior is, but for now it places a pretty big limitation on using both `map` and `zip`. Incidentally, this behavior is in keeping with clojure's `map` insofar as it is also "lazy" in a weird, illegitimate way. Any help on this would be greatly appreciated!

hellerve:
Does this change when you use `Dynamic.+` instead of `+`?

scolsen:
> Does this change when you use `Dynamic.+` instead of `+`?

Good question. I tried it out with `Int.+` earlier to no avail... :man_facepalming: probably should be using `Dynamic` for dynamics, duh. 

I can't check this until later tonight, but I'll let you know how it goes!

If it works using the `Dynamic.+` we'll probably have to add some kind of note about this to the docs for the functions. Or perhaps we should even restrict these functions to taking only `defndynamic` functions? I don't think that would actually help this particular case, since I *think* `Dynamic.+` is still a command, but it may be a good policy regardless, since these functions are expected to operate on form arguments and not on actual carp types.

scolsen:
> Does this change when you use `Dynamic.+` instead of `+`?

No dice! When I use `Dynamic.+` I still encounter the same error.

```clojure
(zip Dynamic.+ '(1 2 3) '(4 5 6))
;; => I can’t match the types `Int` and `(λ [t3, t5] t0)` within `(+ 1 4)`.

  1 : Int
  At line 1, column 18 in 'REPL'

  Expected first argument to '+' : (λ [t3, t5] t0)
  At no info
;; like before, evaluating any particular one of the forms using another function seems to work fine:
(car (zip Dynamic.+ '(1 2 3) '(4 5 6)))
;; => 5
(cadr (zip Dynamic.+ '(1 2 3) '(4 5 6)))
;; => 7
```

Seems fishy, possibly an evaluator bug or an unhandled case?

eriksvedang:
Yeah this is annoying. I think it's very related to the bugs in #555. Making the dynamic evaluator more robust when it comes to evaluation order and semantics is probably the best way to fix this. For example, this result is illuminating:

```
鲮 (collect-into (zip Dynamic.+ [1 2 3] [4 5 6]) array)
=> [(+ 1 4) (+ 2 5) (+ 3 6)]
```

I think we should wait with merge:ing this until we can get the desired results, it'll confuse people a lot if they try to use it.

hellerve:
Looking at this again, `and-internal` and `or-internal` look like they are the same as `Dynamic.and` and `Dynamic.or` (they have to be prefixed with `Dynamic`, which is more than a bit annoying).

scolsen:
Ah you're right! I think I somehow missed the fact that Dynamic.and/or existed at the time. Nice catch!

Also, hoping to get back to carp again soon in general--I still have a mission to write a doc on the evaluator semantics :sweat_smile: --but I've been so busy lately I haven't had a time to look at anything.

hellerve:
No problem, we’re all volunteers, so being available consistently is not to be expected :)

core/Macros.carp
```diff
@@ -119,6 +119,103 @@
...
+  (defndynamic zip-internal [f forms acc]
+    (if (reduce or-internal false (map-internal empty? forms (list)))
+        acc
+        (zip-internal
+          f
+          (map-internal cdr forms (list))
+          (let [result (list (apply f (map-internal car forms (list))))]
+            (append acc result)))))
+
+  (defndynamic zip [f :rest forms]
```
hellerve, line 213: Should we add documentation for this as well?

core/Macros.carp
```diff
@@ -119,6 +119,103 @@
...
+    ;; apply, we can't use zip and map-internal in map.
+    (let [function-name (list (cadr f))]
+      (apply-internal argument-list function-name)))
+
+  (defndynamic map-internal [f xs acc]
+    (if (empty? xs)
+        acc
+        (map-internal f (cdr xs) (cons-last (f (car xs)) acc))))
+
+  (defndynamic zip-internal [f forms acc]
```
hellerve, line 204: Maybe the internal functions should be marked as `hidden`?

core/Macros.carp
```diff
@@ -119,6 +119,103 @@
...
+    ;; Since map calls its function directly, it takes it as is. Apply, on the
+    ;; other hand, would have to take the quoted form, since it builds a list
+    ;; that serves as the actual application.
+    ;;
+    ;; This is problematic for the user facing map function, since it makes
+    ;; calls to map or zip (which uses apply) as appropriate--unless we support
+    ;; the quoted function name argument in map-internal or the unquoted one in
+    ;; apply, we can't use zip and map-internal in map.
+    (let [function-name (list (cadr f))]
+      (apply-internal argument-list function-name)))
```
hellerve, line 197: Can’t we just use `cons` here?

core/Macros.carp
```diff
@@ -119,6 +119,103 @@
   (defndynamic collect-into [xs f]
     (list 'quote
       (collect-into-internal xs (f) f)))
+
+  (doc empty?
+    "Returns true if the provided data literal is empty, false otherwise.")
+  (defndynamic empty? [xs]
+    (if (= 0 (length xs))
```
hellerve, line 126: Can’t we just remove the `if` and return the expression directly?

scolsen:
Alright, the latest commits make some tiny updates, as well as add `flip` and `curry`.

I've tested this with the new evaluator and it looks like everything works as expected!

I guess we could add some more tests, otherwise I think it's ready for a second look when you get a chance, @hellerve and @eriksvedang

core/Macros.carp
scolsen: Yep! b2bb8d0

core/Macros.carp
scolsen: Sounds good! d1155aa

core/Macros.carp
scolsen: yep `append` does the trick! good catch! 0c512c5

(the conses maintain the list structure around the function name or arguments, whereas append just adds the symbol)

core/Macros.carp
scolsen: done! ba0bfe7

scolsen:
Thanks for the comments @hellerve! Should be all resolved now.

eriksvedang:
I'm getting an error...

```
Can't compare 1 with (length a b d) at /Carp/test/macros.carp:63:34.

Traceback:
(= 1 (Dynamic.car mapped)) at /Carp/test/macros.carp:63:31.
(eval (list (quote defn) (quote main) (array) (list (quote l...) at /Carp/core/Test.carp:134:3.
(deftest test (assert-true test (test-let-do) "let-do works ...) at /Carp/test/macros.carp:66:1.
```

hellerve:
It seems like CI is getting the same error.

scolsen:
> I'm getting an error...
> 
> ```
> Can't compare 1 with (length a b d) at /Carp/test/macros.carp:63:34.
> 
> Traceback:
> (= 1 (Dynamic.car mapped)) at /Carp/test/macros.carp:63:31.
> (eval (list (quote defn) (quote main) (array) (list (quote l...) at /Carp/core/Test.carp:134:3.
> (deftest test (assert-true test (test-let-do) "let-do works ...) at /Carp/test/macros.carp:66:1.
> ```

Whoops, that's certainly not right. Let me dig into that.

scolsen:
Should be fixed now! I had some errors in the call to `map`. I also added a test for `zip.`

hellerve:
Seems like the tests are working now!

PR #552, opened 2019-09-11
Updates docs to reflect removal of local-include

TimDeve:
Replaces reference to local-include with relative-include in the language guide

local-include was removed in #543


TimDeve:
Unsure if the example is clear enough

hellerve:
I think it would be worth saying a word what the include is relative to, i.e. the path of the file in which `relative-include` is called.

TimDeve:
Just to be clear, it is relative to the carp file calling the function/macro, correct?

hellerve:
It is!

eriksvedang:
Thanks!

PR #563, opened 2019-09-16
Make String.join take ref

### LINKED ISSUE #562: String.join should take ref to string as first argument

hellerve:
This PR changes `String.join?` to take a reference to a string. Fixes #562.

Cheers

### DIFF
--- a/core/Map.carp
+++ b/core/Map.carp
@@ -279,7 +279,7 @@
 
   (defn str [m]
     (let [res (kv-reduce &(fn [s k v]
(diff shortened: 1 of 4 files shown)


hellerve:
I’ve also surveyed the eco system again, and it seems as if mostly my libraries are affected. I’ll fix the once this is merged.

hellerve:
All of my instances of `String.join` found in [this search](https://github.com/search?q=extension%3Acarp+join) should now be accounted for. It seems, however, as if code from @allenj12 and @titan is also affected.

eriksvedang:
Thanks for fixing this!

PR #564, opened 2019-09-17
Misc fixes

jacereda:


eriksvedang:
This looks great! But I get an error in the test suite:

```
./test/array.carp

I can’t find any implementation for the interface `/=` of type (λ [&Double, &Double] Bool) at line 103, column 18 in "Array.carp"
```

jacereda:
Is it worth having a `/=` interface? The C compiler will probably be smart enough to do the right thing anyways, or am I missing something?

jacereda:
I mean, could we just implement `/=` as a generic function negating `=`?

jacereda:
Wow, the C compiler is VERY smart. I changed `/=` to a generic and the following:

```
(Project.config "compiler" "/usr/bin/clang")
(add-cflag " -std=c99 -march=native -O2 -fomit-frame-pointer -fvisibility=hidden -flto")
(defn main []
  (let [x (random)
        y (random)]
    (cond (/= x (the Float y))
          1
          0)))
```

Yields this code:
```
out/Untitled:
(__TEXT,__text) section
_main:
0000000100000f60	movslq	%edi, %rax
0000000100000f63	movq	%rax, _System_args(%rip)
0000000100000f6a	movq	%rsi, 0x9f(%rip)
0000000100000f71	movl	$0x1, %eax
0000000100000f76	retq
```

eriksvedang:
I have a hunch we implemented the `/=` for some Carp type level reason, but I don't remember... (do you have an opinion @hellerve ?)

Just relying on the `=` interface and having a single generic `/=` seems nicer, of course. Though since we can't have a constraint on the generic type variable in its signature it will lead to slightly more confusing errors when using `/=` on a type without a `=` function implemented.

The emitted assembler looks nice and small :)

hellerve:
I think we didn’t do this way back when the compiler was much less stable, especially around generics (I think this was around the time I implemented the testing library, which would’ve been late 2017). The idea was always to go back to a generic I believe.

jacereda:
OK, I'll open a separate PR with the `/=` removal.

hellerve:
Yup, it was PR #90 back in the day. I don’t think we even had `definterface` back then.

eriksvedang:
Getting new errors when running the test suite on this one.

jacereda:
Passes for me on Mac. What's the error?

hellerve:
I can confirm that it runs fine for me as well.

eriksvedang:
In examples/basics.carp:
```
I can’t match the types `&r7` and `r7`.

  a : &&r7
  At line 71, column 13 in '/Users/erik/Projects/Carp/core/Double.carp'

  (defn DoubleRef./= [a b] (/= (copy a) (copy b))) : (λ [&r7, &r9] t2)
  At line 71, column 3 in '/Users/erik/Projects/Carp/core/Double.carp'
```

jacereda:
Wasn't that one fixed by https://github.com/carp-lang/Carp/pull/564/commits/336b43a2af9984acb38a1d91b6f08ae7bad0e03b ?

Maybe you need to pull again?

eriksvedang:
No, I already have that... not sure what's up :/

eriksvedang:
Seems to work when I remove that function definition.

eriksvedang:
I guess it interacted with the other patch that made ´/=` generic.

jacereda:
Yeah, this one is supposed to be applied before the `/=` removal.

Issue #558, opened 2019-09-13
Make type variable names beautful in error messages


eriksvedang:
Type variables often have weird internal names in error messages, like `r5`, etc. This can be avoided by running the type variable name beautification function that is used in some other places (I think?)

PR #565, opened 2019-09-17
Beautify type variables in error messages.

jacereda:


jacereda:
My probably broken attempt at #558

jacereda:
Do you have any preference on how to break the circular imports?

eriksvedang:
What caused the need for circular imports? Since you seem to have solved it I have a little bit of trouble following? :)

jacereda:
I extracted a `TypeErrorDef` module that is imported from both `TypeError` and `AssignTypes`, but maybe you prefer some other way?

eriksvedang:
This didn't seem to beautify anything that I could see (I tried type errors). How did you test it?

jacereda:
Can't remember the example I used. I tried now with:

```
(load "Vector.carp")
(defn foo []
  (Vector3.add (Vector3.init 1f 1f 1f) (Vector3.init 1.0 1.0 1.0)))
```

And found it's broken , because it turned:

```
I can’t match the types `(Vector3 r4)` and `(Ref (Vector3 r2))` within `(Vector3.add (V ... 1 1))`.

  (Vector3.init 1.0f 1.0f 1.0f) : (Vector3 r4)
  At line 3, column 20 in '/private/tmp/err.carp'

  Expected first argument to 'Vector3.add' : (Ref (Vector3 r2))
  At line 3, column 8 in '/private/tmp/err.carp'
```

Into:

```
./carp.sh -b /tmp/err.carp
I can’t match the types `(Vector3 a)` and `(Ref (Vector3 a))` within `(Vector3.add (V ... 1 1))`.

  (Vector3.init 1.0f 1.0f 1.0f) : a
  At line 3, column 20 in '/private/tmp/err.carp'

  Expected first argument to 'Vector3.add' : (Ref (Vector3 a))
  At line 3, column 8 in '/private/tmp/err.carp'


```

Notice the `: a` in the second report is erasing important info.
I'll try to figure out something more robust.

jacereda:
Now it seems to behave better for that example. I've only addressed `UnificationFailed` for now. 

This is the current error:

```
I can’t match the types `(Maybe c)` and `(Ref (Vector3 b))`. within `(Vector3.add (M ... 1.0))`

  (Maybe.Just (Vector3.init 1.0f 1.0f 1.0f)) : (Maybe c)
  At line 3, column 16 in '/private/tmp/err.carp'

  Expected first argument to 'Vector3.add' : (Ref (Vector3 b))
  At line 3, column 4 in '/private/tmp/err.carp'

```

eriksvedang:
While I think the approach here is a good start, I think we need to rewrite it to make it work with the current code and fix the issues that were found. Thanks for the effort, feel free to submit a new PR if you have the time!

PR #567, opened 2019-09-20
NixOS/Nixpkgs support

jacereda:


bench/map.carp
```diff
@@ -4,21 +4,21 @@
 (defn insert []
   (let [m {}]
     (for [i 0 100]
-      (set! m (Map.put &m &i &1)))))
+      (Map.put! &m &i &1))))
```
hellerve: It seems like these are unrelated to the NixOS change, or am I mistaken?

bench/structs.carp
```diff
@@ -8,7 +8,7 @@
 (defn creation [] (Point.init 0 0))
 (defn bench-access [] (ignore (Point.x &p))) ;; The name 'access' collides with an existing C-function.
 (defn set [] (Point.set-x @&p 5))
-(defn update [] (Point.update-x @&p Int.inc))
+(defn update [] (Point.update-x @&p &Int.inc))
```
hellerve: Same here; is this related to the change to NixOS?

bench/map.carp
jacereda: I changed those so the Nix build can execute the tests correctly.

hellerve: Ah, interesting! What about Nix made it necessary?

jacereda: The nix recipe runs the test suite unless told otherwise. If it fails, the whole installation fails.

jacereda:
This would:

- Allow setting up a consistent development environment across machines with nixpkgs installed.  Just run nix-shell in the Carp directory and start developing using carp.sh to test with all the dependencies installed.
- Simplify the packaging for nixpkgs. The nixpkgs recipe could delegate on this one. The current script in nixpkgs for instance doesn't include the SDL/GLFW... dependencies.

bench/map.carp
hellerve: Oh, so you’re saying that the old implementation was actually broken?

jacereda: Yes, it simply didn't compile.

jacereda:
Oh, I just realised it's using ghc881. That means it will build everything, since the prebuilt packages in the nix cache use another GHC version. I did that because it was failing to recognise ansi-terminal, but maybe I was doing something wrong. I'll try to revert that to the default compiler.

bench/map.carp
hellerve: Oh wow, that’s weird. For me `test/map.carp` works fine, or was it triggered by a different part of the test suite? For me `run_carp_tests.sh` seems to run just fine on the latest version in `master`.

jacereda: It's `bench/map.carp`, so I guess It wasn't due to the test suite but due to me manually testing the `benchmarks.sh` script.

hellerve: Ah, perfect, thank you! For some reason I was thinking we were looking at `core/Map.carp` and was confused how it was possible that it didn’t compile. Sorry about that!

jacereda:
I think the problem I was having is due to mixing normal cabal builds and `cabal new-run`. I'll switch it to `cabal run`.

jacereda:
I have tested with nixpkgs 19.09 and unstable. 19.03 distributes an old `ansi-terminal`.
I haven't tested the `stack` workflow though.

eriksvedang:
I don't like the complexity of this... can't Nix use `stack` like everybody else? :)

run_carp_tests.sh
```diff
@@ -8,11 +8,6 @@ if [[ $# -gt 0 ]] && [[ "$1" == "--no_sdl" ]]; then
     NO_SDL=1
 fi
 
-stack build;
```
eriksvedang: I want this line to stay...

jacereda: > I don't like the complexity of this... can't Nix use `stack` like everybody else? :)

Nix prefers pulling packages from hackage directly and curate them themselves. 

Some info on why:

https://nixos.org/nix-dev/2016-June/020585.html

IMHO nixpkgs provides a better development experience than stack. It's also great for managing non-haskell dependencies for your project.

As to whether this PR adds complexity, it does, but it would simplify my life quite a lot. Every time I switch branches I have to do some `git apply` dance to put the nix configuration in place so I can build, then I've to be careful not to commit any of those bits.

eriksvedang: OK, interesting. I'll test your PR on my machines and get back after some more evaluation.

build.sh
```diff
@@ -0,0 +1,9 @@
+if [ -z "$CARP" ]
+then
+    if [ -z "$NIX_CC" ]
+    then
+	stack build
```
eriksvedang: Thanks :)

hellerve:
It seems to run fine for me. As long as I can still run stack (also not a NixOS user here, because I am a caveman) I’m fine with it!

eriksvedang:
OK, good. I don't want to enforce Nix on anyone, so we should make it keep working with Stack for sure. But it's nice if both works, since Nix seems very popular and well thought out.

Issue #572, opened 2019-09-23
Can't use `load` with HTTP on Windows


TimDeve:
Trying to load a library from a git repo using `load` and HTTP on Windows fails with this error:
```
$ carp -b main.carp --generate-only
I can't find a file named: 'http://github.com/carpentry-org/sockets.git'

I tried interpreting the statement as a git import, but got: fatal: Not a git repository (or any of the parent directories): .git
 at C:\Users\myuser\dev\sandbox\carpload\main.carp:2:8.
```
This is the content of main.carp:
```clojure
(load "https://github.com/carpentry-org/sockets.git@master")
```

hellerve:
I think for now we decided on supporting SSH only.

jacereda:
@TimDeve Can you try that PR?

hellerve:
Oh wow, did we actually leave the `http`/`https` code in? My memory is failing me hard on this one.

TimDeve:
@jacereda Thanks, I'll try it when I get home.

@hellerve Why was it planned to be removed, it works on Linux/Mac.

jacereda:
I would just leave https.

hellerve:
The problem with HTTPS is that it doesn’t work with private repos (because we don’t mirror Git’s stdout which asks for your username and password). That’s often a tad confusing. Go does the same thing IIRC.

The drawback is that this just flat out won’t work if you don’t have an SSH key associated with your Git account.

bjorn3:
> The drawback is that this just flat out won’t work if you don’t have an SSH key associated with your Git account.

Or you dont even have an account. For example CI, or an end user of your program.

TimDeve:
@jacereda I tried your branch, it doesn't seem to have had any effect, sorry.

hellerve:
Is this still a problem or was this fixed with the latest changes?

hellerve:
Clarification: it should work now, since for now we don’t strip `https:` from the final target path and the colon was most likely the problem.

The whole include path system could probably use an overhaul. I’m surprised it isn’t more of a problem already.

TimDeve:
This is the error I get with latest:
```
carp: CreateDirectory "C:\\Users\\myuser\\.carp\\libs\\https:\\github.com\\TimDeve\\hoquet.git\\master": invalid argument (The filename, directory name, or volume label syntax is incorrect.)
```
So it seems it stills try to create a path with a colon in it.

[...]

PR #573, opened 2019-09-23
Blind attempt at fixing #572.

jacereda:
Maybe the problem is it can't write to the home directory?

In any case, shouldn't it write to XdgDirectory instead of home?

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -4,7 +4,7 @@ import Control.Monad
 import qualified System.Environment as SystemEnvironment
 import System.IO (stdout)
 import System.Console.Haskeline (runInputT)
-import System.Directory (doesPathExist, getHomeDirectory)
+import System.Directory (doesPathExist, getXdgDirectory, XdgDirectory(..))
 import GHC.IO.Encoding
 
 import ColorText
@@ -25,7 +25,7 @@ defaultProject =
           , projectFiles = []
           , projectAlreadyLoaded = []
           , projectEchoC = False
-          , projectLibDir = ".carp/libs/"
+          , projectLibDir = "carp/libs/"
           , projectCarpDir = "./"
           , projectOutDir = case platform of
                               Windows -> ".\\out"
@@ -80,8 +80,7 @@ main = do setLocaleEncoding utf8
                                   ""
                                   execMode)
           context <- loadFiles startingContext coreModulesToLoad
-          home <- getHomeDirectory
-          let carpProfile = home ++ "/.carp/profile.carp"
+          carpProfile <- getXdgDirectory XdgConfig "carp/profile.carp"
           hasProfile <- doesPathExist carpProfile
           context' <- if hasProfile
                       then loadFiles context [carpProfile]
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -6,8 +6,8 @@ import Control.Monad.State
 import Control.Monad.State.Lazy (StateT(..), runStateT, liftIO, modify, get, put)
 import System.Exit (exitSuccess, exitFailure, exitWith, ExitCode(..))
 import qualified System.IO as SysIO
-import System.Directory (doesFileExist, canonicalizePath, createDirectoryIfMissing, getCurrentDirectory, getHomeDirectory, setCurrentDirectory)
-import System.FilePath (takeDirectory)
+import System.Directory (doesFileExist, canonicalizePath, createDirectoryIfMissing, getCurrentDirectory, getXdgDirectory, setCurrentDirectory, XdgDirectory(..))
+import System.FilePath (takeDirectory, (</>))
 import System.Process (readProcess, readProcessWithExitCode)
 import Control.Concurrent (forkIO)
 import qualified Data.Map as Map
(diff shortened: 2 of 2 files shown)


hellerve:
I don’t really know about `XdgDirectory` vs `home`, but I’m seeing that this changes the behavior on OS X from writing to `~/.carp/libs/` to `~/.cache/carp/libs`. I’m personally okay with that change.

We shouldn’t, however, just change the behavior of where we look for `profile.carp` without at least a notice; otherwise I for instance probably wouldn’t figure out where to put the file quickly. (Although of course now I looked)

jacereda:
`profile.carp` should probably be loaded from the `XdgConfig` directory.  That would map to `$HOME/.config/carp" on Unix. 

Most software is transitioning from `$HOME/.foo` to `$HOME/.config/foo` nowadays, maybe we should too.

And yes, it should be notified somehow (or provide a backward-compatible fallback for some period).

jacereda:
So, what do you think? Should we go all-in and rename the directories, provide backwards-compatibility or forget about it completely?

hellerve:
I think we should adhere to best practices, so I guess that yes, we should move from the home directory to the XDG directories! Thanks for your work on this!

app/Main.hs
```diff
@@ -80,8 +80,7 @@ main = do setLocaleEncoding utf8
                                   ""
                                   execMode)
           context <- loadFiles startingContext coreModulesToLoad
-          home <- getHomeDirectory
-          let carpProfile = home ++ "/.carp/profile.carp"
+          carpProfile <- getXdgDirectory XdgConfig "carp/profile.carp"
```
hellerve: Should we use `</>` here as well?

app/Main.hs
```diff
@@ -25,7 +25,7 @@ defaultProject =
           , projectFiles = []
           , projectAlreadyLoaded = []
           , projectEchoC = False
-          , projectLibDir = ".carp/libs/"
+          , projectLibDir = "carp/libs/"
```
hellerve: And here (plus the coming lines)?

jacereda: In this case I think it would be better to stick to unix paths and do a `toNative` where needed, since we don't want conditional compilation for something like configuration.

Or we could just use `System.FilePath.Posix` paths everywhere and perform the `toNative` before calling the OS. 

I'll try to evaluate which one looks better.

hellerve:
While we fixed it by #579 I think this is still a good PR to merge!

eriksvedang:
I think this broke repl history, no? I'm fine with merge:ing but we need to update the docs also (ideally in the same PR).

hellerve:
Yeah, we moved directories here, so both `profile.carp` and the history would have to move.

eriksvedang:
The profile file works though.

hellerve:
Do you mean the file works without this PR or having it in the old location still works after this PR? The latter would mean that either I’m misremembering or there is something wrong.

eriksvedang:
The profile file works with the new location, if I move it there. But the profile history doesn't seem to save at all when using this PR. Does it save for you?

hellerve:
I don’t have that problem. In fact it seems that the history file is untouched by this PR (and should be changed). FYI @jacereda the offending line is [this](https://github.com/carp-lang/Carp/blob/master/src/Repl.hs#L75).

It seems like this is an unrelated problem in your setup then, @eriksvedang?

eriksvedang:
I guess so, then :)

eriksvedang:
I figured it out – for history to work you need a `.carp` in your home directory. I removed this to test out the new setup and then the history stopped working. So this PR should update the history part also, to avoid forcing the user to have two directories, right?

hellerve:
Yes, I think we should migrate everything into one directory, and ideally we’d also stop cluttering the home directory. So I think moving everything to XDG is the best way forward!

jacereda:
> I figured it out – for history to work you need a `.carp` in your home directory. I removed this to test out the new setup and then the history stopped working. So this PR should update the history part also, to avoid forcing the user to have two directories, right?

I changed it but I'm not sure if the history should be a config file or a cache file.

jacereda:
I think I would just use System.FilePath.Posix everywhere (so all paths are Unix paths internally) and only before entering the syscall level perform a `toNative` conversion. Otherwise it's hard not to mix different path separators. 

This would mean creating something like a `Path` module with all the required functionality and centralising there all the filesystem accessing functions. Are you OK with this change?

eriksvedang:
> This would mean creating something like a Path module with all the required functionality and centralising there all the filesystem accessing functions. Are you OK with this change?

@jacereda Sounds perfect!

eriksvedang:
> I changed it but I'm not sure if the history should be a config file or a cache file.

I'm not sure what effect it has if it's either, but a cache file sounds right. It shouldn't be very important information after all (you wouldn't want to add it to version control for example).

bjorn3:
I think it should be a config file, as caches are normally for rebuildable things, which a history file is not.

jacereda:
Changes included as part of https://github.com/carp-lang/Carp/pull/587

Issue #526, opened 2019-07-16
Wrong declaration order in generated output


jacereda:
The following code:

```
(load "Vector.carp")

(deftype (V f) [v (Vector3 f)])
(deftype (X f) [x (Maybe (V f))])

(defn main []
  (let [x (X.init (the (Maybe (V Float)) (Maybe.Nothing)))]
    0))
```

yields:

```
./out/main.c:445:5: error: unknown type name 'Maybe__V__float'
    Maybe__V__float x;
    ^
1 error generated.
```

because the generated file declares types in the wrong order:

```
...
// Depth 103
typedef struct {
    Maybe__V__float x;
} X__float;
...
// Depth 105
typedef struct {
    union {
    struct {
        V__float member0;
    } Just;
    // Nothing
    };
    char _tag;
} Maybe__V__float;
...
```

jacereda:
Could it be a problem in `scoreTypeBinder`?

eriksvedang:
Probably, yes!

jacereda:
I think I have a fix, is https://github.com/carp-lang/Carp/pull/529 going to be merged? If so I can PR it on top of it.

jacereda:
OTOH, it's a trial and error fix. Shouldn't the scoring be something like a `Data.Graph.topSort` of the dependencies of each binding?

eriksvedang:
Sorry, I'm moving houses (and having a vacation) so I haven't had the time to merge PR:s, sorry.

eriksvedang:
I merged #529 now, but don't have time to look at the others until in a couple of days.

jacereda:
No hurries, I just wanted to know on top of which branch to PR this one. Enjoy your vacation.

PR #575, opened 2019-09-25
[WIP] Update scoring of defalias

hellerve:
(Duplicate of #574, but as a proper Draft)

This PR is a hack that updates the scoring of defalias by incrementing by more than 1. This is because somehow the scoring that `depthOfType` comes up with for the aliased type is lower (usually 1-2) than what the final score of that type is. The problem can be seen in [my queues repo](https://github.com/hellerve/queues.carp).

As this is just a hack, I’d rather investigate the reason for this more deeply and figure it out for real. I’m quite sure that the problem is related to some scoring discrepancy between wrapped and top-level types, but I can’t quite figure it out as of yet. Someone with more experience with the scoring might be able to illuminate more here.

Independent of all of this, I changed some baseline values from `1` to `0`, since there doesn’t seem to be a reason for `1` to be the baseline value (the outcome isn’t changed and the tests still run). This came out of some prior playing around with the module.

Cheers

### DIFF
--- a/src/Scoring.hs
+++ b/src/Scoring.hs
@@ -19,7 +19,7 @@ scoreTypeBinder typeEnv b@(Binder _ (XObj (Lst (XObj x _ _ : XObj (Sym _ _) _ _ 
       -- we add 1 here because deftypes generate aliases that
       -- will at least have the same score as the type, but
       -- need to come after. the increment represents this dependency
-      in  (depthOfType typeEnv Set.empty selfName aliasedType + 1, b)
+      in (depthOfType typeEnv Set.empty selfName aliasedType + 2, b)
     Deftype s -> depthOfStruct s
     DefSumtype s -> depthOfStruct s
     _ -> (500, b)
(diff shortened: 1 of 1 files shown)


jacereda:
The program that motivated my previous scoring PR is happy with these changes.

Am I the only one who thinks this scoring approach is too hard to maintain? Wouldn't something like toposorting the whole thing prior to emitting be more robust?

hellerve:
Well, the scoring tries to emit a key for sorting, basically, but I do think that the current approach uses a bit too much voodoo (not in a derogatory sense so much as in noone-quite-understands-why). I know why it’s there, and I don’t have any good alternatives at hand, but I don’t like it. The part about not knowing how to do it better keeps me from talking about it too much, though.

hellerve:
By the way, do you have that file somewhere still?

jacereda:
Do you mean the example that triggered the previous bug? It's an implementation of a toy raytracer, the one described in https://github.com/RayTracing/raytracing.github.io, but I haven't uploaded it. I'll do it soon.

jacereda:
In any case, I reduced the problem to the example at https://github.com/carp-lang/Carp/issues/526

eriksvedang:
I'd love to find a simpler approach. But yes, the score is just used to have something to base the sorting on... what could be used instead?

bjorn3:
Maybe crearte a result list and then keep iterating over the input list and when an input has all dependencies already in the result list, append to the result list.

Another option would be to use forward declarations when a dependency hasnt already been emitted.

jacereda:
Only semi-serious here. Let's write a zig emitter and forget completely about scoring: https://ziglang.org/#Order-independent-top-level-declarations

eriksvedang:
Interesting! The algorithm in the compiler that checks the dependencies of functions actually has an implicit graph containing all that is needed (if we would emit in the reverse order, so to speak) so perhaps there is something to be done with that. Need to think more about it but I’m starting to understand that there must be way more elegant ways to do this than what’s currently in the code.

jacereda:
> Do you mean the example that triggered the previous bug? It's an implementation of a toy raytracer, the one described in https://github.com/RayTracing/raytracing.github.io, but I haven't uploaded it. I'll do it soon.

https://github.com/jacereda/raytracing-in-one-weekend-carp

eriksvedang:
Hmm, were these changes merged in a separate PR? @hellerve

PR #576, opened 2019-09-27
Reduce amount of calls to doesFileExist

jacereda:


jacereda:
This could as well use `Control.Monad.Loops.firstM`, but I guess adding another dependency for this is overkill.

eriksvedang:
Nice!

PR #577, opened 2019-09-27
Long type to ensure longs are actually 64 bits.

jacereda:
Plain `long` is 32 bits on some platforms (most notably Windows). Code like the following fails on those:

```
long Double_to_MINUS_bytes(double x) {
    long y;
    memcpy(&y, &x, sizeof(double));
    return y;
}
```


hellerve:
It’d be great if we could have this in core! It’s been open for a while, what’s holding it back?

eriksvedang:
Hmm, not sure. I'm open to merge:ing it as-is but it needs a rebase.

hellerve:
Since this seems to be the only thing preventing a passing Windows CI build, I’d be happy to take this over if you want to hand it over, @jacereda! If you’d like to rebase yourself, though, that’s of course also totally okay.

eriksvedang:
Yes please!

eriksvedang:
This is fixed now.

PR #579, opened 2019-10-02
Replace colon in git path

### LINKED ISSUE #571: Can't use `load` with SSH on Windows
Trying to load a library from a git repo using `load` and SSH on Windows fails with this error:
```
$ carp -b main.carp --generate-only
carp: CreateDirectory "C:\\Users\\myuser\\.carp\\libs\\github.com:carpentry-org\\sockets.git\\master": invalid argument (The filename, directory name, or volume label syntax is incorrect.)
```
This is the content of main.carp:
```clojure
(load "git@github.com:carpentry-org/sockets.git@master")
```

hellerve:
This PR fixes #571 by replacing the colon in a file path produced by a Git checkout by `_COLON_`. This should probably be generalized later on, but maybe it’s good enough for now?

@TimDeve Could you check that this fixes your problem?

Cheers


TimDeve:
Seems to work well, I get a folder called `github.com_COLON_carpentry-org` in my `.carp/libs` folder and no errors, thanks 👌

eriksvedang:
Seems reasonable!

PR #580, opened 2019-10-02
Preserve includes order in generated output.

jacereda:
There're some other bits here I'm unsure about, mostly the changes in `src/StartingEnv.hs` and `src/Workbench.hs`.


eriksvedang:
The workbench file should probably be removed, it was just a place for me to save things to evaluate in the repl when testing out things. Haven't used it in a long time.

eriksvedang:
Thanks for the PR though, I think this is neater.

PR #581, opened 2019-10-04
Numeric protocols

jacereda:
This PR builds on top of https://github.com/carp-lang/Carp/pull/580

It's a big one, but IMHO it's worth headache. At some point I'd like to have sized integers to interact with external libraries and this would help.

I don't like the macros in `Integral.carp` and friends. Is there some way to do that in a less cumbersome way?


jacereda:
Oh, and I don't like the `fconv` and `iconv` names, any suggestion?

jacereda:
And if this gets merged, https://github.com/carp-lang/Carp/pull/577 doesn't make sense anymore.

hellerve:
Wow, this is a super interesting idea! I didn’t have time to look at this in-depth yet, but the general idea looks nice.

I also agree about the `iconv` and `fconv` names, I wasn’t quite sure what they meant for a bit. I actually had to play around with them for a bit to figure out why you need them. I’m not quite sure what would be a good name for it either, sadly; let me mull that over for a bit.

Anyway, great work!

jacereda:
Maybe `integral` and `floating`? Or `to-integral` and `to-floating`, renaming the interfaces to something else?

hellerve:
I like `to-*` and `from-*` as a naming convention; but I think it makes more sense for a simple transmutation than for a double one. I’m not sure that `Integral` and `Floating` are good names for concrete types, though, since they seem very protocol-y (but maybe that’s just Haskell creeping into my thinking).

Maybe `Integer` would be a better name for a concrete type and `Integral` for the conceptual one? 

As for the floating point types, `Floating` seems like a good conceptional name. I was contemplating `Real` for the concrete type, but `Real` should probably be both `Floating` and `Integral` (again, like in Haskell). We could also make `Floating` the concrete type and `Discrete` the conceptual one, but I’m not sure whether that encapsulates the idea of floats at all.

eriksvedang:
Yeah this looks cool! I need to play around with it a bit before I can give any constructive criticism. I think having "from"/"to" in the names make them much easier to understand. On the other hand it's pretty annoying in Haskell to remember when to use which one. But yes, "iconv" sounds more like something that has to do with icons.

core/carp_double.h
```diff
@@ -1,132 +1,8 @@
...
-double Double_floor(double x) {
-    return floor(x);
-}
-
-double Double_mod(double x, double y) {
-    return fmod(x, y);
-}
+typedef double Double;
+#define T Double
+#define SUF(x) x
```
eriksvedang: What does SUF stand for?

src/Emit.hs
```diff
@@ -739,9 +739,10 @@ paramListToC xobjs = intercalate ", " (map getParam xobjs)
         getParam invalid = error (show (InvalidParameter invalid))
 
 projectIncludesToC :: Project -> String
-projectIncludesToC proj = intercalate "\n" (map includerToC (projectIncludes proj)) ++ "\n\n"
+projectIncludesToC proj = intercalate "\n" (map includerToC includes) ++ "\n\n"
   where includerToC (SystemInclude file) = "#include <" ++ file ++ ">"
         includerToC (RelativeInclude file) = "#include \"" ++ file ++ "\""
+        includes = reverse $ projectIncludes proj
```
eriksvedang: This is merged now in the other PR, right?

eriksvedang:
On Windows there are 1 warning and 2 errors:

```
In file included from .\out\main.c:13:
.//core\core.h(50,9): error: unknown type name 'int64_t'
typedef int64_t Integral;
        ^
.//core\core.h(52,15): error: 'static_assert_52' declared as an array with a negative size
STATIC_ASSERT(sizeof(Integral)==sizeof(Floating));
~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.//core\core.h(48,41): note: expanded from macro 'STATIC_ASSERT'
#define STATIC_ASSERT(x) STATIC_ASSERT2(x,__LINE__)
                         ~~~~~~~~~~~~~~~^~~~~~~~~~~
.//core\core.h(47,62): note: expanded from macro 'STATIC_ASSERT2'
#define STATIC_ASSERT2(x,l) extern char CAT(static_assert,l)[x? 0 : -1]
                                                             ^~~~~~~~~
In file included from .\out\main.c:20:
In file included from .//core\carp_double.h:4:
In file included from .//core/carp_floating.h:3:
.//core/carp_numeric.h(16,3): warning: 'memcpy' will always overflow; destination buffer has size 4, but size argument is 8 [-Wfortify-source]
  __builtin_memcpy(&y, &x, sizeof(x)); // FIXME will fail on big-endian
  ^
```

core/carp_double.h
jacereda: It's a macro to potentially append a suffix to a the function name. In the case of `double` it will just use the identifier. For `float` it will append a `f`, so `fabs` becomes `fabsf`.

src/Emit.hs
jacereda: Yes. I can rebase on current master if you want.

jacereda:
Which compiler yields errors? 
Shouldn't we use the `gcc` provided with `ghc` instead of requiring an additional dependency on Windows?

jacereda:
AFAICS the only error is the lack of `int64_t` in that compiler.
It can be typedefed as `typedef signed __int64 int64_t`.

The other errors seem produced by this one.

eriksvedang:
That solved it nicely.

Actually being able to use the Visual Studio compiler is a big selling point when doing gamedev at least, plus people might use the compiler binaries without installing Haskell (especially on Windows) so making it run there is pretty important (in an ideal world it wouldn't be so annoying...)

eriksvedang:
Unless you have any objections @hellerve I'd like to get this merged, it's very neat.

Just need to sort out the naming...

Is there anything prevents us from using `from-int` and `from-float` instead of `iconv` and `fconv`? I think those names are OK, but I guess there must be some reason you changed them @jacereda ?

Oh, and I think there's still some issue on Windows so I need to make sure that works before merge:ing to. Works great on Mac though.

eriksvedang:
Oh, I didn't read the implementation of the conversion functions until now. I see, they rely on `from-` and `to-`... Then we need a third name, right.

eriksvedang:
@jacereda I think your suggestion with `floating` and `integral` are pretty great then. 

Alternatively just name it `convert`, but put the functions into the `Floating` and `Integral` modules, so that using them becomes `Floating.convert` / `Integral.convert`.

core/Int.carp
```diff
@@ -1,72 +1,10 @@
...
-  (doc to-int "acts as the identity function to implement the interface.")
-  (sig to-int (Fn [Int] Int))
-  (defn to-int [a] a)
-
-  (doc from-int "acts as the identity function to implement the interface.")
-  (sig from-int (Fn [Int] Int))
-  (defn from-int [a] a)
 )
 
 (defmodule IntRef
```
eriksvedang: The Integral.register macro could maybe generate these functions too?

examples/guessing.carp
```diff
@@ -14,7 +14,7 @@
       (println "(Enter q to quit.)\n")
 
       (set! guessing true)
-      (set! answer (random-between 1 100))
+      (set! answer (the Int (random-between 1 100)))
```
eriksvedang: This is a little obnoxious? It still works without the explicit cast, right?

src/Emit.hs
eriksvedang: Yes please!

examples/guessing.carp
jacereda: Oh, right. I don't know how that ended up there.

core/Int.carp
jacereda: Do you mean moving the `IntRef` functions to the `Int` module?

jacereda: Addressed this explicitly.

src/Emit.hs
jacereda: Done.

eriksvedang:
Sorry for being slow merge:ing this, have been caught up in lifetimes. I have a few concerns that I need to investigate before pushing the button, but I just wanted to say that I haven't forgotten about it! :)

jacereda:
Sure, no problem.

jacereda:
Rebased on top of current master.

hellerve:
Where are we on this PR?

jacereda:
I'll rebase it again.

hellerve:
Thanks @jacereda! I think @eriksvedang still has to review this, too, so I didn’t want to rush you or anything; this was more meant in his direction :)

eriksvedang:
I now have looked at it more. 

While I think the general idea for this new numerical stack is *very* good I don't think it should be added before the type system supports constraints on type variables. As it is written now (while extremely clever) it relies on pseudo "type classes" in the form of made up types with no instances. The Carp compiler knows nothing of those types and performs no checks on them. I think this will lead to problems for end users since they get almost no help from the system in getting things right.

The other part of the PR are the C macros for generating numeric code. I'm open to merge:ing that in a separate PR. While I think it makes the C code a bit harder to read it will also save a lot of time and make silly copy and paste errors less likely, so it would be a clear win overall.

eriksvedang:
I'll close this for now because of the reasoning above. Still open to merge:ing the C macros separately.

PR #582, opened 2019-10-07
Unify printing booleans in project

hellerve:
This PR unifies printing booleans in `(project)` and other commands that show the current state of the project environment. Previously, a bunch of the boolean variables were printed Haskell style (id est using `True` and `False) and other were printed Carp style (using the lower case versions).

I took the liberty of unifying this to Carp style and introducing a local function for that.

Cheers


eriksvedang:
Neato!

PR #588, opened 2019-10-10
Fix regression when loading via git

hellerve:
This PR fixes a regression that was introduced by #579: we check for `https:` and `http:` after having the colon replaced, meaning this will never work. We have to check for `https_COLON_` and `http_COLON_` instead.

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -1145,8 +1145,8 @@ commandLoad [xobj@(XObj (Str path) i _)] =
     fromURL url =
       let split = splitOn "/" (replaceC ':' "_COLON_" url)
           fst = head split
-      in if fst `elem` ["https:", "http:"]
-        then joinWith "/" $ tail $ tail split
+      in if fst `elem` ["https_COLON_", "http_COLON_"]
+        then joinWith "/" (tail (tail split))
         else
           if '@' `elem` fst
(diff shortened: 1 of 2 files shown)


TimDeve:
It seems to have done the trick, thanks.

hellerve:
Glad to hear this worked! :)

eriksvedang:
Needs a rebase.

hellerve:
Done! I had to update one of the tests. The tests in `master` don’t currently work for me (`test/args.carp`fails)!

eriksvedang:
@hellerve Weird! That change makes the test fail on my machine... 

```
Test 'System.get-arg works as expected' failed:
	Expected value: '.../Carp/outUntitled', actual value: '.../Carp/out/Untitled'
```

eriksvedang:
Turns out it was because I had left off a '/' in the out dir in my config file (where I set a custom "output-directory"). Would be nice if that didn't matter (must be something with the new Path module, I assume...)

hellerve:
Huh, interesting! We should definitely make this more robust. I’ll look into it!

jacereda:
I think the problem with that test is using `String.join` for joining paths. We should probably have `Path.join` at some point.

hellerve:
@jacereda Ah, I didn’t see that comment when I commented in #594. That’s exactly it!

PR #590, opened 2019-10-14
Add write-file

hellerve:
This PR adds the dynamic function `write-file`, which is the writing equivalent to `read-file`. It enables us to write code like this:

```clojure
(defdynamic *c-counter* 0)

(defndynamic inline-c [s]
  (let [f (String.join [(Project.get-config "output-directory")
                        "carp_inline_c_generated" (str *c-counter*) ".h"])]
    (do
      (write-file f s)
      (system-include f))))


(inline-c "int x() { return 1; }")
(register x (Fn [] Int))

(defn main [] (println* &(x)))
```

I hope this is useful enough to motivate why this is a good idea!

Cheers

### DIFF
--- a/docs/core/Dynamic.html
+++ b/docs/core/Dynamic.html
@@ -1644,6 +1644,25 @@
                     
                 </p>
             </div>
+            <div class="binder">
+                <a class="anchor" href="#write-file">
+                    <h3 id="write-file">
+                        write-file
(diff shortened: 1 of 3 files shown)


eriksvedang:
Cool stuff. The `inline-c` function looks handy.

PR #593, opened 2019-10-15
core.h: Use pragma once for multi-include

hellerve:
This PR adds `#pragma once` to `core.h`, to make it possible for external C files to include `core.h` without nameclashes. It shouldn’t do anything else.

Cheers

### DIFF
--- a/core/core.h
+++ b/core/core.h
@@ -1,3 +1,4 @@
+#pragma once
 #if defined(WIN32) || defined(_WIN32) || defined(__WIN32) && !defined(__CYGWIN__)
 #include <windows.h>
 #endif


jacereda:
I did the opposite everywhere in https://github.com/carp-lang/Carp/pull/581

I made header inclusion explicit in Core.carp and removed all includes and guards from the header files.

hellerve:
But wouldn’t that still mean that an external header file that needs to include `core.h`(because of the Array or String types, for instance) runs into a clash, since `core.h` still gets loaded at least twice?

jacereda:
Since the first thing we do in Core.carp is including "core.h", there's no
need to include it again at any point below. Notice that the PR preserves
order of includes to respect whatever order we use in Core.carp.

hellerve:
But that won’t help for external, freestanding C headers that need to include it and get included by Carp again. See also carpentry-org/hydrogen.carp#1.

jacereda:
I think the solution would be just to remove the `#include <core.h>` line
if the numeric protocols PR gets accepted.

hellerve:
No, it would not. That would do nothing for the C header file. It would change things for Carp code only, but this is not the concern here.

eriksvedang:
I think @jacereda is correct, the suggestion works for me at least...

hellerve:
I’m happy to admit that I was completely wrong! I believe that’s because `include`s just expand their input files inline and through moving the dependencies down the import lists we’re able to reference earlier inputs? I’m not quite sure I grasp how this worked, but I’m happy to know that this works and is thus a non-issue.

Thanks for your patience, @jacereda, and sorry for me misunderstanding your point/how C works entirely.

PR #594, opened 2019-10-16
Ensure directory paths have trailing slash in Project.config

hellerve:
This PR fixes the problem @eriksvedang encountered in #588 by ensuring that all paths fed into `Project.config` have a trailing slash.

Cheers

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -108,9 +108,9 @@ commandProjectConfig [xobj@(XObj (Str key) _ _), value] =
                      "title" -> do title <- unwrapStringXObj value
                                    return (proj { projectTitle = title })
                      "output-directory" -> do outDir <- unwrapStringXObj value
-                                              return (proj { projectOutDir = outDir })
+                                              return (proj { projectOutDir = addTrailingPathSeparator outDir })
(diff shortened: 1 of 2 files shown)


jacereda:
Why is this needed? It might be a symptom of a missing `</>` somewhere...

hellerve:
In `tests/args.carp` we do `(Dynamic.String.join [(Project.get-config "output-directory") "somefile"])` for test purposes; if the user sets their output directory, then having a trailing slash suddenly becomes important in their tests. While it might also be interesting—and certainly more principled—to add a function like `Dynamic.Path.join` that does the right thing, I think uniformizing the format might be a good idea for our users anyway (i.e. the two are not mutually exclusive).

If you think that not uniformizing that format might be a better call or have an idea where to move this instead, I’d be happy to hear about that!

eriksvedang:
I'm not sure this PR works, when I set the output-directory without a trailing slash it stays that way..?

eriksvedang:
I never got this to work. Is it still relevant? @hellerve

hellerve:
We fixed this by changing the test!

PR #600, opened 2019-10-24
Add a byte type

hellerve:
This PR adds the type `Byte` to the core types (which desugars to `uint8_t`). It mostly works like the other numerical types, except that it’s unsigned, and so `neg`, `abs`, and `positive-mod` don’t really make sense.

Test cases are included. I suspect that this clashes quite a bit with #581, and I’m happy to backport that work into this (or the other way around, whichever is easier).

Cheers

### DIFF
--- a/core/Byte.carp
+++ b/core/Byte.carp
@@ -1,0 +1,61 @@
+(system-include "carp_byte.h")
+
+(defmodule Byte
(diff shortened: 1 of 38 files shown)


hellerve:
Now that #581 is on the backburner, what’s the status on this?

eriksvedang:
Let's get this merged then.

PR #603, opened 2019-10-28
into lifetimes: unsafe-first and unsafe-last return references

hellerve:
This PR changes the unsafe array functions to returns references directly. According to GitHub search, only Carp internally uses those functions, so they should be safe for change.

Cheers


hellerve:
Superseded by work on master.

PR #604, opened 2019-10-28
Simplify Char.meaning

hellerve:
This PR simplifies `Char.meaning` by removing a bunch of copies and using `case` instead of `cond`.

Cheers


eriksvedang:
Much nicer!

PR #605, opened 2019-10-28
Reduce copies in heap

hellerve:
This updates @chrisosaurus wonderful heap library by removing the copies, since we now know how to call function references!

Cheers


eriksvedang:
Way to go!

chrisosaurus:
Awesome job, thanks for this.

PR #609, opened 2019-10-30
Add .clang-format

hellerve:
This PR adds a `.clang-format` file and reformats our codebase according to it. The changes should exclusively expand or contract lines.

Cheers

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,8 +3,8 @@ BasedOnStyle: Google
 AlignEscapedNewlinesLeft: 'true'
 AlignTrailingComments: 'true'
 AllowAllParametersOfDeclarationOnNextLine: 'true'
-AllowShortBlocksOnASingleLine: 'true'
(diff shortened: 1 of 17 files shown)


eriksvedang:
Looks great mostly, but can we please keep function declarations on multiple lines instead of cramming them all into one? :)

/ The bike shedding BDFL

hellerve:
Done! Now a bunch of short functions were expanded to multiple lines, but if you think that’s a better default I’m all for it!

eriksvedang:
Well, I think that letting the author decide on that detail on a case-by-case basis would work too, but if there has to be the same for all functions I like this a lot more.

Should we add something that runs this in the test script or so?

hellerve:
It already does this, because I took the liberty of adding it to `build.sh`!

As for case-by-case, you can always disable it by saying `clang-format: off` for a certain region of code!

PR #610, opened 2019-10-30
Disallow non-symbols in fn arguments

### LINKED ISSUE #414: Identifiers starting with digits should throw a parse error

hellerve:
This PR checks `fn` arguments for non-symbols and returns an error if it does. This finally accounts for all the cases and thus finally fixes #414!

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -147,6 +147,11 @@ eval env xobj =
                    _ -> return (makeEvalError ctx Nothing ("`if` condition contains non-boolean value: " ++ pretty okCondition) (info okCondition))
                Left err -> return (Left err)
 
+        [XObj (Fn _ _) _ _, args@(XObj (Arr a) _ _), _] ->
+            if all isUnqualifiedSym a
+            then return (Right listXObj)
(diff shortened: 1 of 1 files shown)


eriksvedang:
This doesn't work for me, I get an error when I define a lambda with a number as an argument:

```
(let [x (fn [1] 123)] ())
```

```
carp: src/Concretize.hs:(97,36)-(98,63): Non-exhaustive patterns in lambda
```

eriksvedang:
Perhaps you choose the wron branch for the PR?

eriksvedang:
Or wait, I didn't get the code in the PR... something weird with it.

PR #611, opened 2019-10-30
Filter out unit types from call

### LINKED ISSUE #557: Functions taking unit type as argument emits invalid C code
Functions taking (and returning?) the unit type as an argument, i.e. `f : (Fn [()] ())` will emit invalid C code.

hellerve:
This PR fixes #557 by removing unit types from calling forms. It makes programs like this work:

```
(defn f [x] x)

(defn main [] (f ()))
```

Which before errored by trying to emit a parameter of type `void` in the argument list.

Cheers


eriksvedang:
Yay!!!

PR #615, opened 2019-10-31
Add 2 missing includes on core/Core.carp

GrayJack:
Looks like there was 2 missing includes on the Core.carp file, making compilation fail on systems that those are not imported by default

### DIFF
--- a/core/Core.carp
+++ b/core/Core.carp
@@ -12,6 +12,8 @@
 (system-include "carp_stdbool.h")
 (system-include "core.h")
 (system-include "carp_memory.h")
+(system-include "sys/wait.h")
+(system-include "unistd.h")
 
 (load "Interfaces.carp")
 (load "Bool.carp")


hellerve:
I’m not sure whether this applies, but should we wrap this in a `(linux-only)` form or something like that?

Which systems does it fail on?

GrayJack:
Failed on my Linux system, but both headers are supposed to exist, be used and have problems compiling in it's absence on all Unix/Unix-Like OS, so it would be prudent to be in a `(unix-only)` form

GrayJack:
Interesting, maybe MacOS import them as default, well, he have to test this on other BSDs to be sure

jacereda:
On MacOS, something we’re already including is indirectly including those. 
I would leave them without any ifdef, just to be more explicit. Now it’s clear we’re using things from unistd (fork() at least). 
If at some point that’s optional, we can add the ifdefs.

hellerve:
That makes sense! Let’s do that then :)

eriksvedang:
Thanks @GrayJack !

PR #616, opened 2019-10-31
Rename known unsafe function to make clear that they're unsafe

GrayJack:
- [X] Maybe.ptr -> Maybe.unsafe-ptr
 - [x] Array.nth -> Array.unsafe-nth


hellerve:
Hey! Before we merge this, it might be interesting to look at [where nth is used in external projects](https://github.com/search?q=extension%3Acarp+nth&type=Code) so we can decide who all needs to know about this!

jacereda:
What about having a `Compat.carp` module where `nth` and `ptr` are defined by just calling the corresponding`unsafe-xxx`? Then it would be nice to have a deprecation annotation stating that they'll be removed in version 0.x.0...

hellerve:
@jacereda That sounds like a good idea! We can write macros like those:

```
(defmacro nth [a i]
   (do
     (macro-log "This version of nth will be renamed to unsafe-nth in version 0.4.0.")
     (macro-log "The new version of nth will return a `Maybe`.")
     (unsafe-nth a i)))
```

GrayJack:
> Should we in this pull request also add a new version of `nth`, something like the version below?
> 
> ```clojure
> (defn nth [a i]
>   (if (and (>= i 0) (< i (length a)))
>     (Maybe.Just @(unsafe-nth a i)) ; the copy will go away with lifetimes
>     (Maybe.Nothing)))
> ```

Would that be in the scope of this PR? I wondered the same thing

GrayJack:
@jacereda I like the idea, but I also can be a shoot in the foot if people start to `(use Compat.Array)` and later in the code just use `nth`, I think it could easily lead to wrong usage 

Can we have something in between? Like modules marked as `hidden` or whatever can only be called using full path, or any other thing

jacereda:
@hellerve That could work for now, the problem with the approach is forgetting to remove it :)

I guess the right thing would be to add metadata with the version in which it will be removed, so that when the version is bumped the `nth` function itself stops compiling.

Something like:
```
(deprecated 0 4 0)
(defn nth ...
```

jacereda:
@GrayJack Well, when I find a deprecation I can decide to ignore the warning or adapt to the code. If I choose the former, I can only blame myself if my code stops compiling in a future release.

Issue #617, opened 2019-11-01
Wrap all C stlib functions that return a pointer in a Result type


sdilts:
Many C library functions return a null pointer on error, i.e. `malloc` and `fopen`, etc. To make these more safe, their return values should be wrapped in a `Result` type.

sdilts:
Most of these functions also set `errno` to indicate how the function failed. Is that something that should be left up to the caller to check, or should we try to interpret it in the error we return?

eriksvedang:
I think this sounds like a great idea. Since most C functions have weird names I think it would work well to keep those around but use more lispy names for the safe ones.

PR #618, opened 2019-11-01
Add open-file function

sdilts:
`fopen` returns a nullable pointer, which doesn't fit very well with Carp's notions of saftey and error handling. Add the function `open-file` that returns a `Result` type that forces the user to check if the operation was successful.

Partial fix for #617


jacereda:
Wouldn't it be better to return `(Result Int (Ptr FILE))`, with the `Int` part being the `errno`?

sdilts:
That would make a lot more sense.

hellerve:
Could we add a test to the test suite while we’re at it?

sdilts:
Not sure what the tests need to be since we are just testing a wrapper for a system function, but I think these are enough. Is there anything else I should add?

hellerve:
No, that looks great to me now!

eriksvedang:
Very nice!

PR #619, opened 2019-11-03
Add FP exception handling, x86 specific

jacereda:
This would need some #ifdef work on other platforms.

### DIFF
--- a/core/fpe.h
+++ b/core/fpe.h
@@ -1,0 +1,11 @@
+#include <fenv.h>
+
+static void enable_exceptions(int e) {
+    fenv_t fe;
+    fegetenv(&fe);
+#if !defined __linux__
+    fe.__control &= ~e;
+#endif
+    fe.__mxcsr &= ~(e << 7);
+    fesetenv(&fe);
+}
--- a/core/FPE.carp
+++ b/core/FPE.carp
@@ -1,0 +1,12 @@
+(defmodule FPE
+  (relative-include "fpe.h")
+
+  (register division-by-zero Int "FE_DIVBYZERO")
+  (register inexact Int "FE_INEXACT")
+  (register invalid Int "FE_INVALID")
+  (register overflow Int "FE_OVERFLOW")
+  (register underflow Int "FE_UNDERFLOW")
+  (register denormal-operand Int "FE_DENORMALOPERAND")
+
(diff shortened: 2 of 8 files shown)


core/fpe.h
```diff
@@ -0,0 +1,8 @@
+#include <fenv.h>
+
+static void enable_exceptions(int e) {
+    fenv_t fe = *FE_DFL_ENV;
+    fe.__control &= ~e;
+    fe.__mxcsr &= ~(e << 7);
+    fesetenv(&fe);
+}
```
bjorn3: This works only for x86, right?

jacereda: Yes, x86/x64 AFAIK.

eriksvedang:
This is cool! There is a problem on macOS though, because of case insensitivity. The test file `/test-for-exceptions/fpe.carp` loads itself because it's called the same thing as the core library file. Please change the name of the test file to something more descriptive.

eriksvedang:
I tried renaming the test from `fpe.carp` to `fpe_test.carp` but I still can't make it pass.

hellerve:
I also can’t trivially get this to work. My errors do not align with what is expected.

We should probably revisit this at a later point, because I quite like the idea of this PR.

eriksvedang:
Works on my Mac now, so that's great!

Fails in Windows, though the tests are not run there? So not sure what that is about...

hellerve:
> Fails in Windows, though the tests are not run there? So not sure what that is about...

I’m not quite sure what you mean by that, the output seems to be from that test:

```
./test-for-exceptions/test-fpe.carp
In file included from out\main.c:32:
D:/a/Carp/Carp/core\fpe.h(7,8): error: no member named '__control' in 'struct fenv_t'
    fe.__control &= ~e;
    ~~ ^
D:/a/Carp/Carp/core\fpe.h(9,8): error: no member named '__mxcsr' in 'struct fenv_t'
    fe.__mxcsr &= ~(e << 7);
```

eriksvedang:
Oh, OK. I forgot that the Windows CI also use the bash script now, sorry.

eriksvedang:
I guess we need a workaround for that then...

jacereda:
There must be some way to do this in a more portable way, but I still have to figure it out.

eriksvedang:
Will close for now, please open again when if you get it to work on all platforms!

PR #620, opened 2019-11-04
Update md files

GrayJack:
Not complete, lack some information for me

`with` form is a special form?
Can `match` form pattern match primitive types?

### DIFF
--- a/docs/LanguageGuide.md
+++ b/docs/LanguageGuide.md
@@ -76,6 +76,7 @@ and other static analysis. The first three of them are also available in dynamic
 (do <expr1> <expr2> ... <return-expression>) ;; Perform side-effecting functions, then return a value
 (if <expression> <true-branch> <false-branch>) ;; Branching
 (while <expression> <body>) ;; Loop until expression is false
+(use <module>) ;; Brings all symbols inside <module> into the scope
 (with <module> <expr1> <expr2> ...) ;; Locally scoped `use` statement where all expressions after it will look up symbols in the <module>
 (match <expression> <case1> <expr1> <case2> <expr2> ...) ;; Pattern matches <expression> against a set of sumtype constructors
 (ref <expression>) ;; Borrow an owned value
--- a/docs/LanguageGuide.md
+++ b/docs/LanguageGuide.md
@@ -76,7 +76,8 @@ and other static analysis. The first three of them are also available in dynamic
 (do <expr1> <expr2> ... <return-expression>) ;; Perform side-effecting functions, then return a value
 (if <expression> <true-branch> <false-branch>) ;; Branching
 (while <expression> <body>) ;; Loop until expression is false
-(match <expression> <case1> <expression1> <case2> <expression2> ...) ;; Pattern matches <expression> against a set of sumtype constructors
+(with <module> <expr1> <expr2> ...) ;; Locally scoped `use` statement where all expressions after it will look up symbols in the <module>
+(match <expression> <case1> <expr1> <case2> <expr2> ...) ;; Pattern matches <expression> against a set of sumtype constructors
 (ref <expression>) ;; Borrow an owned value
 (address <expression>) ;; Takes the memory address of a value, returns a C-style pointer
 (set! <variable> <expression>) ;; Mutate a variable
--- a/docs/LanguageGuide.md
+++ b/docs/LanguageGuide.md
@@ -25,7 +25,7 @@ To learn more about the details of memory management, check out [Memory.md](http
 10.0    ;; Double
 true    ;; Bool
 "hello" ;; &String
-\#"hello" ;; &Pattern
+#"hello" ;; &Pattern
 \e      ;; Char
 [1 2 3] ;; (Array Int)
 {1 1.0 2 2.0} ;; (Map Int Double)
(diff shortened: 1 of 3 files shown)


docs/LanguageGuide.md
```diff
@@ -217,6 +219,23 @@ Sometimes, it's more convenient to bring a module's declarations into scope only
 (Vector2.update-x my-pos inc) ;; => (Vector2 11 20)
 ```
 
+### Enums / Algebraic types
+```clojure
+(deftype (Either a b)
+  (Left a)
```
hellerve: The members here must be wrapped in array, i.e. `(Left [a])`.

docs/LanguageGuide.md
```diff
@@ -75,6 +76,7 @@ and other static analysis. The first three of them are also available in dynamic
 (do <expr1> <expr2> ... <return-expression>) ;; Perform side-effecting functions, then return a value
 (if <expression> <true-branch> <false-branch>) ;; Branching
 (while <expression> <body>) ;; Loop until expression is false
+(match <expression> <case1> <expression1> <case2> <expression2> ...) ;; Pattern match the first expression in a exhaustive cases
```
hellerve: `match` only works on sumtypes for now.

docs/LanguageGuide.md
```diff
@@ -217,6 +219,23 @@ Sometimes, it's more convenient to bring a module's declarations into scope only
 (Vector2.update-x my-pos inc) ;; => (Vector2 11 20)
 ```
 
+### Enums / Algebraic types
```
hellerve: We usually prefer the term sumtypes over algebraic types in Carp speak.

docs/LanguageGuide.md
```diff
@@ -217,6 +219,23 @@ Sometimes, it's more convenient to bring a module's declarations into scope only
...
+```clojure
+(deftype (Either a b)
+  (Left a)
+  (Right b))
+
+;; A Variant can be created with the same syntax as call expression
+(Left 10)
```
hellerve: You either need to `use` `Either` before you can do that, or you’ll have to write `Either.Left`.

docs/LanguageGuide.md
GrayJack: True, forgot about that when I wrote that

docs/LanguageGuide.md
GrayJack: It would be better to just add a note or change de entire comment?

hellerve: Good question. How about `Pattern matches <expression> against a set of sumtype constructors`? I don’t know whether that is clear enough, though.

GrayJack: There is plans to support pattern match on non sumtypes?

> I don’t know whether that is clear enough, though.

It's clear enought for me

hellerve: Eventually yes, but we want to sort out a bunch of other fundamental things first. Contributions are always welcome, though! (Though this is likely to be a little too hard to add for someone who doesn’t have a deep understanding of the Carp compiler)

GrayJack:
Another question pop up by looking at the stdlib source code,
In core/Color.carp has shown this pattern
```closure
(deftype Id
     Black
     Red
     Green
     Yellow
     Blue
     ...)
```

That means that sumtypes variants doesn't need to be between parenthesis when doesn't depend on anything?

I ask cause the pattern used by `Maybe`, when declared the variant `Nothing` put it between parenthesis
```closure
(deftype (Maybe a)
  (Nothing)
  (Just [a]))
```

hellerve:
I stumbled upon the same thing within #621. Apparently this does work! I just didn’t remember.

eriksvedang:
Yeah, perhaps we should have an example of that in the document too?

eriksvedang:
This is awesome btw, thanks a lot @GrayJack !!!

docs/LanguageGuide.md
```diff
@@ -220,12 +220,26 @@ Sometimes, it's more convenient to bring a module's declarations into scope only
...
+
+**Enumeration:**
+```clojure
+(deftype MyEnum
+  Kind1
+  Kind2
+  Kind3)
+```
+
+**Data:**
```
GrayJack: I was not sure what to put in here, any suggestions?

docs/Tooling.md
```diff
@@ -27,6 +27,9 @@ For Flycheck:
 
 To start an interactive session, make sure `carp` is in your path (inside Emacs) and execute `M-x run-carp`.
 
+# Atom
+[language-carp](https://atom.io/packages/language-carp) offers both TextMate and Tree-sitter, where tree-siiter grammar is more powerful
```
jacereda: There's a typo there (siiter)

GrayJack: There is no typo in 'sitter', pro project call tree-sitter, look out their site https://tree-sitter.github.io/tree-sitter/

jacereda: There is, continue reading after the comma.

GrayJack: Yeah, sorry, misted that, I was without glasses xD
Was hard to spot on

docs/Tooling.md
```diff
@@ -27,6 +27,9 @@ For Flycheck:
 
 To start an interactive session, make sure `carp` is in your path (inside Emacs) and execute `M-x run-carp`.
 
+# Atom
+[language-carp](https://atom.io/packages/language-carp) offers highlight with both TextMate and Tree-sitter, where tree-sitter grammar is more powerful
```
hellerve: Is there a period missing at the end?

eriksvedang:
Thanks a lot!

PR #621, opened 2019-11-04
macros: add enum form

hellerve:
This PR adds the `enum` macro form. It is basically a convenience form around `deftype` with sumtypes without arguments. That means that this

```clojure
(enum TwoState
  First
  Second
)
```

desugars to this

```clojure
(deftype TwoState
  (First [])
  (Second [])
)
```

It is thus purely a vehicle for convenience and clarity, and not for functionality. Therefore it is subject to all concerns for style that anyone might have. If you think the name needs to change, or the form needs to change, let me know!

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -342,3 +342,12 @@ The expression must be evaluable at compile time.")
 (doc gensym "generates symbols dynamically as needed.")
 (defndynamic gensym []
   (gensym-with 'gensym-generated))
+
+(defndynamic make-enum-constructors [constructors]
+  (if (= (length constructors) 0)
+    ()
+    (cons (list (car constructors) [])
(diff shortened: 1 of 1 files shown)


eriksvedang:
Deftype should already support the short version, no?

hellerve:
Aw, I totally forgot about this feature! Well, I guess that is preferable.

I will say, however, that moving bits out of the compiler and reducing overloading of forms (like deftype, which has a whole lot of different forms) might be a preferable way forward.

eriksvedang:
Yeah, perhaps you're right! The macros would have to be helpfully written then to tell you when you mess up the syntax, but that's perfectly doable of course. I think keeping this as they are for now works well though, so I'll close this PR and we can get back to this at some point.

PR #622, opened 2019-11-09
macros: add cxr, nthcar, and nthcdr

hellerve:
This PR adds the dynamic functions `cxr`, `nthcar`, and `nthcdr`. They build on top of `car` and `cdr` to allow arbitrary accesses, like so:

```clojure
(defdynamic x '(1 2 ((3) 4) 5))
(cxr '(3 a 2 d) x) ; => 3, equivalent to caaaddr
(nthcdr 2 x) ; => (((3) 4) 5)
(nthcar 2 x) ; => ((3) 4)
```

I’m not quite sure about the naming of `nthcar`, because it suggests `n` `car` operations, when in fact it’s `n` `cdr`s followed by one `car`. In Common Lisp, this is just called `nth`, but since we already use that name, I’d like to avoid that. Any alternative suggestions are welcome!

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -73,6 +73,25 @@
   (defndynamic cdddar [pair] (cdr (cdr (cdr (car pair)))))
   (defndynamic cddddr [pair] (cdr (cdr (cdr (cdr pair)))))
 
+  (defndynamic cxr [x pair]
+    (if (= (length x) 0)
+      (list 'quote pair)
+      (list
+        (if (= 'a (cadr x))
+          'car
+          (if (= 'd (cadr x))
+            'cdr
+            (macro-error "`cxr` expects either `a` or `d` symbols, got " (cadr x))))
+        (if (= 1 (car x))
+          (cxr (cddr x) pair)
+          (cxr (cons (- (car x) 1) (cdr x)) pair)))))
+
+  (defndynamic nthcdr [n pair]
+    (cxr (list n 'd) pair))
+
+  (defndynamic nthcar [n pair]
+    (cxr (list 1 'a n 'd) pair))
+
   (defndynamic eval-internal [form]
     (list 'do
           (list 'defn 'main [] (list 'IO.println* form))


jacereda:
Isn't `nthcar` just Haskell's `drop`?

jacereda:
As for `nthcar`, what about `at`?

hellerve:
You are right about `nthcdr` and `drop`. As for `at`, that describes it pretty well, but creates a new name for something we already know as `nth`. We could also reuse `nth`, knowing that it would behave differently at compile time than it would at runtime (which is maybe okay?)

In sum: both of your suggestions work super well, but since we already have Lispy terminology for these concepts, I’d like to reuse them if possible (but that’s just personal preference).


Nota bene: In Clojure, they use `nth` and `nthrest`, the latter of which I find a little yucky.

scolsen:
Super cool! 

I think using `nth` is fine, actually. Since it's defined in the `Dynamic` module, I think it should be clear that it has a compile time meaning, and it should usually be fairly clear which instance is called.

I like `at` too! Concise and clear.

If we really can't use `nth` or `at`,  `nthcar` is the composition `init . drop` -- can this composition of functions maybe generate ideas? 

How about `car-from` or `car-at`? The implication being you're getting the `car` from the n position:

`car-at n list`

We could then rename `nthcdr` for parity to `cdr-at`, Then we'd have:

```clojure
(defdynamic x '(1 2 ((3) 4) 5))
(cxr '(3 a 2 d) x) ; => 3, equivalent to caaaddr
(cdr-at 2 x) ; => (((3) 4) 5)
(car-at 2 x) ; => ((3) 4)
```

However, I think we would need a slight rewrite to use these names, since the function behavior isn't totally symmetrical, unlike the names would suggest.

```clojure
For (cdr-at/nthcdr 2 x)
(1  2 ((3) 4) 5))
    ^ -- cdr called from here giving (3 4 5)
-- thus 1 == index 1 since n == 2
For (car-at/nthcar 2 x)
(1  2 ((3) 4) 5))
        ^ -- car called from here giving (3 4)!
-- thus 1 == index 0 since n == 2
```

If my thinking's right (there's always a decent chance it isn't!) the two functions actual index the lists differently, one from 0, one from 1. I personally think we should make their indexical view into the list symmetrical whether or not it breaks from precedent in other languages. So, calling `nthcdr` or `nthcar` with the same n would call the respective function from the same point in the list.

hellerve:
Sorry @scolsen, I didn’t see your comment until after this was merged. I think your thinking is right, and `nthcdr/cdr-at` should actually apply `cdr` one more time, so that we both have symmetry, and `the first tail` aka. `nthcdr 0` is actually a tail and not the list!

PR #625, opened 2019-11-13
RFC: Add Array.unreduce

hellerve:
This PR adds `Array.map->`. This seems like a very useful function to me, but I may be alone with that judgement, so please tell me if you’d rather not have this in the standard library.

`map->` represents a functional generalization of looping in a `map`-like interface. To quote from its docstring:

`map->` creates an array by producing values using `step` until they no longer satisfy `test`. The initial value is `start`.

Example:
```clojure
; if we didn’t have Array.range, we could define it like this:
(defn range [start end step]
  (map-> start &(fn [x] (< x (+ step end))) &(fn [x] (+ x step)))
)
```

It is thus a little hard to understand, but once you do it is fairly powerful.

Cheers

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -257,17 +257,17 @@ This function copies the array. If you don’t want that, use [`endo-map`](#endo
         (aset-uninitialized! &na i (~f (unsafe-nth a i))))
       na))
 
-  (doc map-> "creates an array by producing values using `step` until they no
-longer satisfy `test`. The initial value is `start`.
+  (doc unreduce "creates an array by producing values using `step` until they
+no longer satisfy `test`. The initial value is `start`.
 
 Example:
 ```
 ; if we didn’t have Array.range, we could define it like this:
 (defn range [start end step]
-  (map-> start &(fn [x] (< x (+ step end))) &(fn [x] (+ x step)))
+  (unreduce start &(fn [x] (< x (+ step end))) &(fn [x] (+ x step)))
 )
 ```")
-  (defn map-> [start test step]
+  (defn unreduce [start test step]
     (let-do [elem start
              acc []]
       (while-do (~test elem)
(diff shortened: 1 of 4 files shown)


scolsen:
Very cool!

If I'm understanding it correctly, it's an implementation of [`anamorphism`](https://en.wikipedia.org/wiki/Anamorphism)?

Actually, the name is quite apt since morphisms rely on functors (which are intimately related to map)!

I wanted to implement all the morphisms for Carp in a library, but I couldn't do it in the standard "functional" way. Using the looping constructs as an alternative is a nice approach!

Still, I wonder if there's a better name than `map->`? `anamorphism` is too haughty and academic, but I think `unfold` is another name for this function that is a nice indicator of what it does if you know what `fold` does--I forget, do we have `fold` in Carp? Having a clearer name or more traditional name might help others intuit what it does.

Seems useful to me!

GrayJack:
So what about `unreduce` to maintain the `reduce` name?  :stuck_out_tongue_winking_eye: 

Kinda joking, but kinda serious as well

scolsen:
I'm cool with `unfold`, `unreduce`, or....`deduce`! Get it? Because we're *deducing* what the contents of the data structure should be, kind of like a programmatic sherlock holmes? Maybe, yes? no? ok... :(

But in all seriousness both `unfold` or `unreduce` sound fine to me. I think keeping symmetry between the names is nice. The most important thing is making sure what the function does is clear, I think both of those names serve that purpose. I'm also a-ok with the original `map->` too and trust @hellerve's judgement.

GrayJack:
Well in that case, I prefer the `unreduce` to keep the symmetry with `reduce`

hellerve:
I’ve adjusted the name to `unreduce`!

eriksvedang:
The name of this branch (with the `->`) doesn't work on my machine...

hellerve:
Does it work if you enclose it in quotes (like `"map->"`)?

eriksvedang:
It did!

PR #626, opened 2019-11-20
[win] attempt to fix windows compilation

nateupstairs:
Can Dynamic.or be used in Core.carp like this? or will this never evaluate true on non-win systems? This could be hacky in some way, but it's small and fixed compilation on my system.


eriksvedang:
Finally got around to starting my Windows machine... :) Thanks a lot for this!

PR #630, opened 2019-11-29
Add comp macro

hellerve:
This PR adds `comp`, a macro that translates function composition.

Example:

```clojure
((comp inc inc) 10)
; gets rewritten to
((fn [<gensym>] (inc (inc <gensym>))) 10)
```

Cheers


hellerve:
Rebased this against current master, should be mergeable again.

PR #632, opened 2020-01-06
Fix example in Array.reduce documentation

### LINKED ISSUE #631: The example for Array.reduce is incorrect
[The documentation for `Array.reduce`](https://github.com/carp-lang/Carp/blob/c2d479424674538caeb2696baee8dc4779f9c1eb/core/Array.carp#L8) has this alternative definition of Array.sum as an example:
```
(defn sum [x]
  (reduce + 0 x))
```

But this does not work since reduce passes each element by reference. So I assume the example should be something more similar to the actual definition of `Array.sum` like
`(Array.reduce &(fn [x y] (+ x @y)) 0 xs)`.

Or is the plan to change the language so that the example definition will run as it is?

hellerve:
This PR fixes #631 by adjusting the example in `Array.reduce`. Thank you @Emiluren for pointing it out!

Cheers

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -6,7 +6,7 @@ As an example, consider this definition of `sum` based on `reduce`:
 
 ```
 (defn sum [x]
-  (reduce + 0 x))
+  (reduce &(fn [x y] (+ x @y)) 0 x))
 ```
 
 It will sum the previous sum with each new value, starting at `0`.")


eriksvedang:
Thanks for the fix! Would it be useful enough to have a macro that does this code transformation?

PR #634, opened 2020-01-13
core: make result functions take refs

hellerve:
This PR makes sure that `Result` functions take refs. Before, they would take ownership over the function, which we generally do not want.

Cheers


eriksvedang:
Good fix!

PR #635, opened 2020-01-13
core: add String.from-cstr

hellerve:
This PR adds `String.from-cstr`. It’s a noop function that only appeases the typechecker.

Cheers

### DIFF
--- a/core/String.carp
+++ b/core/String.carp
@@ -10,6 +10,7 @@
   (register copy       (Fn [&String] String))
   (register length     (Fn [&String] Int))
   (register cstr       (Fn [&String] (Ptr Char)))
+  (register from-cstr  (Fn [(Ptr Char)] String))
   (register str        (Fn [&String] String))
   (register prn        (Fn [&String] String))
   (register index-of   (Fn [&String Char] Int))
--- a/core/carp_string.h
+++ b/core/carp_string.h
@@ -93,7 +93,7 @@ char *String_cstr(const String *s) {
 }
 
 String String_from_MINUS_cstr(char *s) {
-    return s;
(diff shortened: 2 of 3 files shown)


hellerve:
Actually, this is unsafe for constant strings, because they then get picked up by the deleter. Can we mark strings to not get picked up by the deleter without marking them as ref?

hellerve:
Alternatively we could also copy the string, which would be a little annoying.

eriksvedang:
> Actually, this is unsafe for constant strings, because they then get picked up by the deleter. Can we mark strings to not get picked up by the deleter without marking them as ref?

Not as it stands now, and I have an uneasy feeling about it anyways. 

I think copying should be the default way to solve this, to ensure safety. In what situation is this necessary though? (I haven't looked at the code that requires this)

hellerve:
When C gives us back strings that we don’t own but want to work with. For instance, the C struct `dirent` has a `d_name` property that is the directory name. We’re not supposed to delete that, but we may want to work with it. The solution I came up with was to make it a `(Ptr Char)`, and then transmute it to a `String`. Does that make it a little clearer?

hellerve:
I’m fine with copying, by the way.

eriksvedang:
Right, thanks for the clarification. `(Ptr Char)` is something I've used with great success, `Ptr` is kind of the hint to the compiler to keep its hands off the memory. So if that works for you, use that. Or do the copy:ing, if you like the interface more that way.

hellerve:
Yeah, that works, this is just the way to get a `String` back from that, because from a utility standpoint a `(Ptr Char)` is useless in Carp.

PR #637, opened 2020-01-14
Check index in char-at

### LINKED ISSUE #636: Error message for `Dynamic.String.char-at` is not informative
When using `Dynamic.String.char-at`, the error message is uninformative. This is probably most obvious in `fmt`, but it can be most easily reproduced by calling it directly, for instance like so:

```clojure
> (Dynamic.String.char-at "hi" 3)
carp: Prelude.!!: index too large
```

This should definitely throw a better error message.

Cheers

hellerve:
This PR fixes #636 by checking the index of `char-at`. It also makes the same error in `fmt` more informative.

A test case is included.

Cheers

### DIFF
--- a/core/Format.carp
+++ b/core/Format.carp
@@ -5,32 +5,34 @@
         len (String.length s)]
     (if (= idx -1)
       (list 'copy s) ; no more splits found, just return string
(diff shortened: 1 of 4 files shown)


eriksvedang:
Nice!

PR #641, opened 2020-01-18
Add Array.unsafe-raw

### LINKED ISSUE #591: Array.raw should take a reference
Currently, the type signature of `Array.raw` is `(Fn [(Array a)] (Ptr a))`. I believe it should be `(Fn [(Ref (Array a))] (Ptr a))` instead. The reason I believe that would be better is that it’s currently impossible to interface with side-effecting functions that make use of the raw memory and then getting it back in Carp. While this is supremely unsafe, it is also how a bunch of low-level C libraries operate—writing into preallocated buffers rather than allocating and returning their own. I think we should provide the means to interoperate with those libraries.

What do y’all think?

Cheers

GrayJack:
This PR add the function `Array.unsafe-raw` in the _ArrayTemplate.hs_ 

I included 2 test checking

This should fix #591

### DIFF
--- a/test/array.carp
+++ b/test/array.carp
@@ -289,7 +289,7 @@
   (assert-equal test
                 &2
                 (Pointer.to-ref (Pointer.inc (unsafe-raw &[1 2 3])))
-                "unsafe-raw works2")
+                "unsafe-raw works II")
   (assert-equal test
                 &[1 3 5]
                 &(remove &2 [2 1 3 2 5])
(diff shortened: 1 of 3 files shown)


GrayJack:
@hellerve Can you check if this is really what you wanted?

test/array.carp
```diff
@@ -282,6 +282,14 @@
...
                 (predicate-count &[1 8 5 10 3] &(fn [i] (Int.even? @i)))
                 "predicate-count works")
+  (assert-equal test
+                &1
+                (Pointer.to-ref (unsafe-raw &[1 2 3]))
+                "unsafe-raw works")
+  (assert-equal test
+                &2
+                (Pointer.to-ref (Pointer.inc (unsafe-raw &[1 2 3])))
+                "unsafe-raw works2")
```
hellerve: Usually we comment like this: `unsafe-raw works II`

PR #645, opened 2020-01-25
Make Maybe.apply take function ref

hellerve:
This PR changes `Maybe.apply` to take a function ref.

Test cases are updated.

Cheers


eriksvedang:
Neat, thanks!

PR #646, opened 2020-01-25
core: add IO.getenv

hellerve:
This PR adds `IO.getenv`, which makes it possible to inquire about environment variables. It returns a `Maybe`.

Test cases are included.

Cheers


scolsen:
Sweet, I was just checking whether we had a way to do this carp yet, now we will :D 

Makes it possible to write some terminal manipulation libs which should be fun.

hellerve:
Yes! Also check out [`path`](https://veitheller.de/git/carpentry/path), which should make this even simpler :)

eriksvedang:
Cool.

PR #647, opened 2020-01-27
Better performance in Map./Set.put!

hellerve:
This PR increases the performance of `put!` for `Map` and `Set` in the presence of collisions by reducing indirection and allocation.

The peformance can be tested with, for instance, this program:

```clojure
(load "Bench.carp")

(add-cflag "-O3 -flto -march=native -fomit-frame-pointer")

(defn main []
  (let-do [m (Set.create-with-len 16777216)
           x (Bench.get-time-elapsed)]
    (for [i 0 16777216] (Set.put! &m &(Int.random)))
    (Bench.print "Time: " (- (Bench.get-time-elapsed) x))))
```

On my machine (a recent MacBook Pro), this prints around 7.5 seconds versus around 13 seconds before, so it’s pretty substantial.

Cheers

### DIFF
--- a/core/Map.carp
+++ b/core/Map.carp
@@ -73,9 +73,15 @@
     (do (Array.aupdate! (entries &b) i &(fn [p] (Pair.set-b @p @val)))
         b))
 
+  (defn set-idx! [b i val]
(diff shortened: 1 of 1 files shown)


eriksvedang:
Friggin' awesome.

PR #650, opened 2020-01-27
Use asprintf in *_format and *_str

hellerve:
This PR changes `carp_string.h` to use `asprintf` wherever we used allocation and `snprintf`. It’s part of GNU/BSD and thus not guaranteed to be there, so we probably won’t be able to add it. I want to open this PR to discuss the change anyway, because it does provide a marked speed improvement.

I don’t think reimplementing it ourselves is feasible, sadly.

Cheers

### DIFF
--- a/core/carp_string.h
+++ b/core/carp_string.h
@@ -76,11 +76,8 @@ bool String__LT_(const String *a, const String *b) {
 }
 
 String String_append(const String *a, const String *b) {
-    int la = strlen(*a);
-    int lb = strlen(*b);
-    int total = la + lb + 1;
-    String buffer = CARP_MALLOC(total);
-    snprintf(buffer, total, "%s%s", *a, *b);
+    String buffer;
+    asprintf(&buffer, "%s%s", *a, *b);
     return buffer;
 }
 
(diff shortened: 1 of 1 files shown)


jacereda:
What's the rationale against it? It's in GNU, BSD and musl. That might easily cover 99.99% of use cases.

https://github.com/rofl0r/musl/blob/master/src/stdio/asprintf.c

hellerve:
I don’t know about its protability (I’m especially worried about Windows). But I’d be delighted to be wrong, since it’s better than the hacks in #651, in my opinion.

eriksvedang:
I can try this on WIndows too and see if it works. If so, it's in.

hellerve:
Sounds good. It’s also somewhat superseding #651 then. Let me know whether it works, if it does I’m open to reworking it; it seems a bit more principled (and faster) than #651.

eriksvedang:
OK!

hellerve:
Did you have time to test this yet?

eriksvedang:
Sorry, no. It has conflicts now but I can try it on Windows as soon as those are fixed.

eriksvedang:
I have tried it now and unfortunately the `asprintf` function is not found when compiling via `clang-cl.exe` (which is the one supported right now on Windows).

hellerve:
Alright, in this case I’d drop support for it right now since I can’t add the magic incantations to make it work (I’m sure there is a way).

PR #651, opened 2020-01-27
Use snprintf only when checking length

hellerve:
This PR is basically the vanilla version of #650. We’re removing the length checks with `snprintf` when we know the size already. This avoids a few extra checks.

Cheers

### DIFF
--- a/core/carp_pattern.h
+++ b/core/carp_pattern.h
@@ -670,13 +670,13 @@ Array Pattern_global_MINUS_match(Pattern *p, String *s) {
 String Pattern_internal_add_char(String a, char b) {
     if (!a) {
         String buffer = CARP_MALLOC(2);
-        snprintf(buffer, 2, "%c", b);
+        sprintf(buffer, "%c", b);
         return buffer;
     }
 
     int len = strlen(a) + 2;
     String buffer = CARP_MALLOC(len);
-    snprintf(buffer, len, "%s%c", a, b);
+    sprintf(buffer, "%s%c", a, b);
     CARP_FREE(a);
     return buffer;
 }
@@ -747,7 +747,7 @@ String Pattern_substitute(Pattern *p, String *s, String *t, int ns) {
 
     int l = strlen(res) + strlen(str) + 1;
     String buffer = CARP_MALLOC(l);
-    snprintf(buffer, l, "%s%s", res, str);
+    sprintf(buffer, "%s%s", res, str);
     CARP_FREE(res);
     return buffer;
 }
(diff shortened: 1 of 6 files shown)


jacereda:
This will yield tons of warnings on Windows due to sprintf() deprecation.

hellerve:
That’s problematic of course. Do you know how to avoid those? As we’ve discussed on Gitter, those checks actually have a noticable performance impact on some programs, and it’d be nice to avoid duplicate work!

hellerve:
Googling tells me `#define _CRT_SECURE_NO_WARNINGS` should avoid those, but I’m not clear whether this shadows other possible problems that we actually do want to see?

jacereda:
For things like `Int_str` I usually do:

```
String Int_str(int x) {
     char buf[64]; 
     size_t size = snprintf(buf, sizeof(buf)-1, "%d", x);
     buf[size] = 0;
     return memdup(buf, size+1);
 }
```

hellerve:
That makes sense! I’m open to doing this for the primitive types.

It won’t help us on more flexible functions such as `String_append` or (most prominently) the array templates, though, and I’d like them to profit as well.

`String_append` should probably be rewritten using `strncat` anyway...

jacereda:
For append I would just use:
```
size_t sa = strlen(a);
size_t sb = strlen(b);
size_t sz = sa + sb;
char * buf = CARP_MALLOC(sz+1);
memcpy(buf, a, sa);
memcpy(buf+sa, b, sb);
buf[sz] = 0;
```

hellerve:
Heh, fun! If you look at my latest commit, you’ll see that that’s exactly what I ended up doing :)

eriksvedang:
I'll try this PR on Windows before merge:ing.

eriksvedang:
This one works fine on Windows so I'll merge it.

hellerve:
Did it not throw errors on `MSVC`? That’d be weird.

eriksvedang:
Using Clang that is.

hellerve:
Ah, I think the crux is supporting MSVC, but I sadly don’t have access to Windows atm.

eriksvedang:
Basically nothing in the Carp generated C works with the MSVC compiler front-end. I'm using their linker though, so no "Linux Subsystem for Windows" or whatever it's called. I think this is the most reasonable approach for Windows compatibility going forward, I've been able to compile the minimal SDL example like that.

eriksvedang:
Having the generated code work with the Visual Studio compiler would be great of course, but it's a lot harder and more annoying so that's for the future, if anything.

hellerve:
Ah dang. Alright, I’m not the right person to fix this, but it is a big bummer that msvc is unsupported. Should we mention this in the README for now?

eriksvedang:
It's kinda mentioned in Install.md under C compiler but it could probably be clarified further, yes.

PR #656, opened 2020-01-29
Make foreach (almost) hygienic

hellerve:
This PR changes `foreach` to be more hygienic. I changed the variables `len` and `xs` to `gensym`s to make sure that we don’t introduce unwanted captures. The variable `i` I deliberately left unhygienic, because it’s often useful to know the iteration number we’re on (though arguably, that’s what `for` is designed to do). I’d be happy to introduce a gensym for that as well.

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -170,11 +170,13 @@
 ;;         (list f (list 'Array.nth (list 'ref xs) 'i))))
 
 (defndynamic foreach-internal [var xs expr]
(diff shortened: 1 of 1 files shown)


eriksvedang:
Yeah I think making `i` hygenic would be great too, since nested loops gets messed up right now I think.

hellerve:
Good point, I fixed it!

PR #658, opened 2020-01-29
Add missing function list-to-array

sdilts:
+ Before this commit, list-to-array-internal existed, but no list-to-array.

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -109,6 +109,9 @@
         acc
         (list-to-array-internal (cdr xs) (append acc (array (car xs))))))
 
+  (defndynamic list-to-array [lst]
+    (list-to-array-internal lst []))
+
   (defndynamic collect-into-internal [xs acc f]
     (if (= 0 (length xs))
         acc


hellerve:
This is what `collect-into` was meant for as well. `(collect-into '(1 2 3) array)` should work. Why there even still is `list-to-array-internal` is a bit of a mistery to me.

eriksvedang:
I'm not sure I understand your comment @hellerve, do you approve of the PR or is it redundant? :)

hellerve:
I think the function generally makes sense, though I’d rather make `collect-into` the default; if we do use the other two functions, they should probably rather be based on it rather than implementing their own traversals.

scolsen:
I think `list-to-array-internal` was supposed to be deleted but wasn't--as @hellerve mentioned, we decided to use a generic `collect-into` to cover both cases instead.

sdilts:
Okay. I'll close this and open another one to delete `list-to-array-internal` when I get the time`.

PR #666, opened 2020-02-07
Make *-internal functions in dynamic private and hidden

sdilts:


hellerve:
This looks good, although `private` and `hidden` only really work inside of a module (the idea is that no documentation is generated for them and they can’t be accessed from outside the module).

eriksvedang:
Yeah, sorry this has no effect as-is now. It's a good idea though, but it would require us to put those functions in modules which has a lot of usability repercusions. I'll close it for now.

PR #667, opened 2020-02-07
Add plists

sdilts:
A plist, or property list, is a data structure for storing a small amount of key-value pairs in a list. Among other things, they are useful for implementing keyword arguments in function or macro parameter lists.

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -147,7 +147,7 @@
         (plist-remove-internal (cdr (cdr plist)) key (Dynamic.cons (cadr plist) (cons (car plist) accum))))
       ()))
 
-  (doc plist-remove "Return a new plist with the given key and its corresponding item removed")
+  (doc plist-remove "Return a new plist with the given key and its corresponding item removed.")
   (defndynamic plist-remove [plist key]
     (plist-remove-internal plist key ()))
 
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -119,6 +119,16 @@
   (defndynamic collect-into [xs f]
     (list 'quote
       (collect-into-internal xs (f) f)))
+
+  (private reverse-internal)
+  (hidden reverse-internal)
+  (defndynamic reverse-internal [lst accum]
+    (if (not (= 0 (length lst)))
+      (reverse-internal (cdr lst) (cons (car lst) accum))
+      accum))
+
+  (defndynamic reverse [lst]
+    (reverse-internal lst ()))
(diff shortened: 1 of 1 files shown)


core/Macros.carp
```diff
@@ -119,6 +119,52 @@
...
+  (private plist-remove-internal)
+  (hidden plist-remove-internal)
+  (defndynamic plist-remove-internal [plist key accum]
+    (if (not (= 0 (length plist)))
+      (if (= (car plist) key)
+        (append (reverse accum) (cdr (cdr plist)))
+        (plist-remove-internal (cdr (cdr plist)) key (Dynamic.cons (cadr plist) (cons (car plist) accum))))
+      ()))
+
+  (doc plist-remove "Return a new plist with the given key and its corresponding item removed")
```
hellerve: Nitpick: can we put a period at the end of the sentence here?

sdilts:
I would prefer having a hashmap too, but plists are essential for implementing the `destructuring-bind` macro. Keyword argument lists *are* plists, so they are what you use to extract variable values out of the argument list. Seeing what `destructuring-bind` expands to might be the best way to see what is going on. Here's how it works in Common Lisp. Recall that `sb-c::getf` does something similar to `plist-get`

``` Common Lisp
(destructuring-bind (&key foo bar) '(:bar 2 :foo 1)
  (print foo)
  (print bar))
```
Produces the output:
```
1
2
```
And basically expands to:
``` Common Lisp
(let* ((#:g496 (:bar 2 :foo 1)
       (#:g497 (sb-c::ds-getf #:g496 ':foo))
       (foo
        (if (not (eql #:g497 0))
            (car #:g497)))
       (#:g498 (sb-c::ds-getf #:g496 ':bar))
       (bar
        (if (not (eql #:g498 0))
            (car #:g498))))
  (print foo)
  (print bar))
```
Also, plists are slightly easier to implement.

hellerve:
Alright, I understand their usefulness in this context!

I want to point out that there is no reason that you couldn’t implement this using hashmaps if you convert the plist first, which is slightly awkward but reuses a more generally useful data structure instead of using one that is only useful in that context; I’m in favor of adding it, though and am merely trying to make this alternative (I don’t want to present it as if I think you haven’t thought about that, I don’t know whether you have) explicit.

eriksvedang:
While this code looks good, I think it better belongs in the library that needs it. We should add support for dynamic dictionaries instead.

PR #670, opened 2020-02-09
Fix multiple things in pattern

hellerve:
This PR fixes multiple things in `Pattern`:

### Newlines & Tabs

Firstly, this PR fixes how newlines and tabs are treated in patterns and adds carriage returns. Previously, the match was aborted on the first match of a newline due to a bug (`goto init;` wasn’t called). This PR simplifies their handler also.

### Captures

Secondly, this PR fixes capture groups; there was the possibility for memory corruption if capture groups weren’t initialized, but the length calculation for the strings doesn’t need to depend on those capture groups in the first place. We also introduce `Pattern.split` and `Pattern.global-match-str`, which first triggered that bug.

Cheers


eriksvedang:
Great work.

PR #674, opened 2020-02-11
Introduce slice

hellerve:
This PR introduces `slices` an interfaces that creates subcollections from collections. It supersedes `substring` and `subarray`. I also took the liberty of renaming `prefix-<collection>` and `suffix-<collection>` to just `prefix` and `suffix`.

Since `suffix` depends on `length` we can’t make it a generic just yet, but I hope we’ll be able to rectify that soon.

This PR breaks existing behavior, thus our libraries should be vetted for usage of these functions. Fore reference [here’s the current usage of `subarray`](https://github.com/search?q=extension%3Acarp+subarray&type=Code). Once this has been approved I’ll update all Carpentry packages accordingly.

Cheers

### DIFF
--- a/bench/array_subarray.carp
+++ b/bench/array_subarray.carp
@@ -8,7 +8,7 @@
 
 (defn some-subarray []
   (let-do [a (replicate n &1)
-           b (subarray &a 0 (/ n 2))]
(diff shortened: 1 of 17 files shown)


eriksvedang:
Very nice!

PR #676, opened 2020-02-12
core: add check in starts-with?/ends-with? for string length

hellerve:
This PR fixes a problem in `starts-with?` and `ends-with?`, namely they didn’t check whether the substrings were actually shorter than the strings to test against, which is a requirement.

Cheers

### DIFF
--- a/core/String.carp
+++ b/core/String.carp
@@ -76,9 +76,15 @@
 
   (doc starts-with? "Check if the string `s` begins with the string `sub`.")
   (defn starts-with? [s sub]
-    (= sub &(prefix s (length sub))))
+    (let [ls (length sub)]
+      (and (>= (length s) ls) (= sub &(prefix s ls)))))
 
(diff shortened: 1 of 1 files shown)


core/String.carp
```diff
@@ -76,9 +76,15 @@
...
+      (and (>= (length s) ls) (= sub &(prefix s ls)))))
 
   (doc ends-with? "Check if the string `s` ends with the string `sub`.")
+  (defn ends-with? [s sub]
+    (let [ls (length s)
+          lsub (length sub)]
+      (and (>= ls lsub) (= sub &(suffix s (- ls lsub))))))
+
   (defn ends-with? [s sub]
     (= sub &(suffix s (- (length s) (length sub)))))
```
davidrusu: Did you mean to keep this old definition of `ends-with?` here?

hellerve: Nope, sorry 🤦‍♂ Fixed now!

PR #678, opened 2020-02-13
Make evaluated let sequencable

hellerve:
This PR changes the semantics of `let` in an evaluated context so that later bindings can refer to prior ones, i.e. this is now possible:

```clojure
(let [x 1
      y (+ x 1)]
   y)
```

This was already possible in compiled code, just not when evaluating. Now this is possible as well.

Cheers


eriksvedang:
Useful, indeed!

PR #680, opened 2020-02-13
Add additional custom comparator to for

hellerve:
This PR adds custom comparators to `for`. This is necessary especially when using expressions as steps. We can implement a generic `for` loop with arbitrary numerical types like so, for instance:

```clojure
(for [(zero) n (one) <] (IO.println &(str n))
```

`n` is the only thing that determines which numerical type to use here, which is pretty cool.

But we can also use more advanced comparators, if we really want to.

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -143,12 +143,15 @@
     (let [variable (car settings)
           from (cadr settings)
           to (caddr settings)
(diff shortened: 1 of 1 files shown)


eriksvedang:
Oh that's cool. I like the example usage!

PR #683, opened 2020-02-16
Add Binary module

scolsen:
This PR adds a binary module for converting byte sequences to and from commonplace int encodings (16, 32, 64). 

## Open Questions:

1. *Do we need a new module for this?* We could simply add this functionality to the `Byte` module (which inspired this module)--my original reason for separating the two is that the `Byte` module is currently nice and small, and while these modules are interrelated, their functionality is separable. That said, it might be nice to keep it all in one place--I'm open to suggestions on this point.

2. *Is there a way we can use the `Byte` functions directly to implement this?* I couldn't figure out how. Seems like we need to do the type conversion (and unfortunately thus the actual bit twiddling in C to make it work). If that is the case, this might be one argument for merging this with `Byte` to reduce the number of headers we rely on.

## Progress
- [x] Add unsafe byte sequence -> intN functions
- [x] Add safe byte sequence -> intN functions
- [x] Add intN -> byte sequence functions
- [x] Add tests for the Binary module 

## Out of Scope/ Future Work

I'd like to pursue these improvements, but I'd like to get the baseline conversion functions we've defined at this point into core first.

- Add safe binary sequence -> intVar for variable length encodings.
- Add functions for retrieving n bits from a byte sequence

### DIFF
--- a/core/Binary.carp
+++ b/core/Binary.carp
@@ -68,6 +68,9 @@
   (register to-int16 (λ [Byte Byte] Uint16))
   (register to-int32 (λ [Byte Byte Byte Byte] Uint32))
   (register to-int64 (λ [Byte Byte Byte Byte Byte Byte Byte Byte] Uint64))
+  (register int16-to-byte  (λ [(Ref Uint16)] Byte))
+  (register int32-to-byte  (λ [(Ref Uint32)] Byte))
+  (register int64-to-byte  (λ [(Ref Uint64)] Byte))
   (register system-endianness-internal (λ [] Int))
 
   (doc system-endianness
@@ -102,6 +105,16 @@
       (Order.BigEndian)
         (Maybe.zip &to-int16 (Array.nth bytes 1) (Array.nth bytes 0))))
 
+  (doc int16->bytes
+    "Converts a Uint16 to a sequence of bytes representing the value using the provided `order`")
+  (sig int16->bytes (Fn [Order Uint16] (Array Byte)))
+  (defn int16->bytes [order i]
+    (match order
+      (Order.LittleEndian)
+        (Array.copy-map &int16-to-byte &[i (Uint16.bit-shift-right i (Uint16.from-long 8l))])
+      (Order.BigEndian)
+        (Array.copy-map &int16-to-byte &[(Uint16.bit-shift-right i (Uint16.from-long 8l)) i])))
+
   (doc bytes->int32-unsafe
     "Interprets the first four bytes in a byte sequence as an Uint32 value.
      **This operation is unsafe.**")
@@ -129,6 +142,19 @@
       (Order.BigEndian)
         (Maybe.zip4 &to-int32 (Array.nth bs 3) (Array.nth bs 2)
                               (Array.nth bs 1) (Array.nth bs 0))))
+
+  (doc int32->bytes
+    "Converts a Uint32 to a sequence of bytes representing the value using the provided `order`")
+  (sig int32->bytes (Fn [Order Uint32] (Array Byte)))
+  (defn int32->bytes [order i]
+    (let [shift (fn [lng] (Uint32.bit-shift-right i (Uint32.from-long lng)))]
+      (match order
+        (Order.LittleEndian)
+          (Array.copy-map &int32-to-byte
+            &[i (shift 8l) (shift 16l) (shift 24l)])
+        (Order.BigEndian)
+          (Array.copy-map &int32-to-byte
+            &[(shift 24l) (shift 16l) (shift 8l) i]))))
 
   (doc bytes->int64-unsafe
     "Interprets the first eight bytes in a byte sequence as an Uint64 value.
@@ -165,6 +191,22 @@
                               (Array.nth bs 5) (Array.nth bs 4)
                               (Array.nth bs 3) (Array.nth bs 2)
                               (Array.nth bs 1) (Array.nth bs 0))))
+  (doc int64->bytes
+    "Converts a Uint64 to a sequence of bytes representing the value using the provided `order`")
+  (sig int64->bytes (Fn [Order Uint64] (Array Byte)))
+  (defn int64->bytes [order i]
+    (let [shift (fn [lng] (Uint64.bit-shift-right i (Uint64.from-long lng)))]
+      (match order
+        (Order.LittleEndian)
+          (Array.copy-map &int64-to-byte
+            &[i (shift 8l) (shift 16l)
+                (shift 24l) (shift 32l)
+                (shift 40l) (shift 48l) (shift 56l)])
+        (Order.BigEndian)
+          (Array.copy-map &int64-to-byte
+            &[(shift 56l) (shift 48l)
+              (shift 40l) (shift 32l)
+              (shift 24l) (shift 16l) (shift 8l) i]))))
 
   (doc bytes->int16-seq-unsafe
     "Interprets a sequence of bytes as a sequence of Uint16 values.
--- a/core/Binary.carp
+++ b/core/Binary.carp
@@ -14,6 +14,14 @@
   (register to-int16 (λ [Byte Byte] Int))
   (register to-int32 (λ [Byte Byte Byte Byte] Int))
   (register to-int64 (λ [Byte Byte Byte Byte Byte Byte Byte Byte] Int))
+  (register system-endianness-internal (λ [] Int))
+
+  (doc system-endianness "Returns the endianness of the host system.")
+  (sig system-endianness (λ [] Order))
+  (defn system-endianness [] 
+    (if (= (system-endianness-internal) 1)
+        (Order.LittleEndian)
+        (Order.BigEndian)))
 
   (doc bytes->int16-unsafe
     "Interprets the first two bytes in a byte sequence as an int16 value.
--- a/core/Binary.carp
+++ b/core/Binary.carp
@@ -2,8 +2,8 @@
 (load "StdInt.carp")
 
 (defmodule Binary
-  (doc Order 
-    "The type of byte orders. 
+  (doc Order
+    "The type of byte orders.
 
     LittleEndian designates the little endian ordering, and indicates the least
     significant byte appears first in a given byte sequence.
@@ -17,10 +17,10 @@
   (register to-int64 (λ [Byte Byte Byte Byte Byte Byte Byte Byte] Uint64))
   (register system-endianness-internal (λ [] Int))
 
-  (doc system-endianness 
+  (doc system-endianness
     "Returns the endianness of the host system.")
   (sig system-endianness (λ [] Order))
-  (defn system-endianness [] 
+  (defn system-endianness []
     (if (= (system-endianness-internal) 1)
         (Order.LittleEndian)
         (Order.BigEndian)))
@@ -30,7 +30,7 @@
      **This operation is unsafe.**")
   (sig bytes->int16-unsafe (Fn [Order (Ref (Array Byte) a)] Uint16))
   (defn bytes->int16-unsafe [order bs]
-    (match order 
+    (match order
       (Order.LittleEndian)
         (to-int16 @(Array.unsafe-nth bs 0) @(Array.unsafe-nth bs 1))
       (Order.BigEndian)
@@ -41,11 +41,11 @@
      **This operation is unsafe.**")
   (sig bytes->int32-unsafe (Fn [Order (Ref (Array Byte))] Uint32))
   (defn bytes->int32-unsafe [order bs]
-    (match order 
-      (Order.LittleEndian) 
-        (to-int32 @(Array.unsafe-nth bs 0) @(Array.unsafe-nth bs 1) 
-                  @(Array.unsafe-nth bs 2) @(Array.unsafe-nth bs 3)) 
-      (Order.BigEndian) 
+    (match order
+      (Order.LittleEndian)
+        (to-int32 @(Array.unsafe-nth bs 0) @(Array.unsafe-nth bs 1)
+                  @(Array.unsafe-nth bs 2) @(Array.unsafe-nth bs 3))
+      (Order.BigEndian)
         (to-int32 @(Array.unsafe-nth bs 3) @(Array.unsafe-nth bs 2)
                   @(Array.unsafe-nth bs 1) @(Array.unsafe-nth bs 0))))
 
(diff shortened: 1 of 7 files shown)


scolsen:
@eriksvedang @hellerve, a draft for you to take a peek at. I definitely think this is functionality we should support in core, but I do have a few questions about where exactly it should live (see the PR description).

Let me know what you think! Hoping to get this into a baseline mergeable state by the end of the weekend.

hellerve:
This looks quite useful. Maybe it would also make sense to think about adding [stdint.carp](https://github.com/carpentry-org/stdint.carp) into core, since it seems like there would be a lot of synergies with this code—and being able to express the width of your types in a safe manner seems desirable?

hellerve:
If we do this, I propose we add stdint first and build this module on top of that.

core/Binary.carp
```diff
@@ -0,0 +1,101 @@
...
+                  @(Array.unsafe-nth bs 3) @(Array.unsafe-nth bs 2)
+                  @(Array.unsafe-nth bs 1) @(Array.unsafe-nth bs 0))))
+ 
+  (doc byte-seq->int8-seq-unsafe 
+    "Interprets a sequence of bytes as a sequence of int8 values.
+     **This operation is unsafe.**")
+  (sig byte-seq->int8-seq (Fn [(Ref (Array Byte) a)] (Array Int)))
+  (defn byte-seq->int8-seq [bs]
+    (let [f (fn [b] (Byte.to-int @b))]
+    (Array.copy-map &f bs)))
```
hellerve: I suggest indenting here.

core/Binary.carp
```diff
@@ -0,0 +1,101 @@
...
+(defmodule Binary
+  (doc Order 
+    "The type of byte orders. 
+
+    LittleEndian designates the little endian ordering, and indicates the least
+    significant byte appears first in a given byte sequence.
+
+    BigEndian designates the big endian ordering, and indicates the most
+    significant byte occurs first in a given byte sequence.")
+  (deftype Order LittleEndian BigEndian)
```
hellerve: How about adding a function that returns the host system’s endianness?

eriksvedang:
Looks good to me, very nice work @scolsen !

I'm definitely open to merge:ing the stdint module first @hellerve – let's do it!

core/Binary.carp
scolsen: Great idea, thanks!

core/Binary.carp
```diff
@@ -0,0 +1,111 @@
...
+                  @(Array.unsafe-nth bs 2) @(Array.unsafe-nth bs 3)
+                  @(Array.unsafe-nth bs 4) @(Array.unsafe-nth bs 5)
+                  @(Array.unsafe-nth bs 6) @(Array.unsafe-nth bs 7))
+      (Order.BigEndian) 
+        (to-int64 @(Array.unsafe-nth bs 7) @(Array.unsafe-nth bs 6)
+                  @(Array.unsafe-nth bs 5) @(Array.unsafe-nth bs 4)
+                  @(Array.unsafe-nth bs 3) @(Array.unsafe-nth bs 2)
+                  @(Array.unsafe-nth bs 1) @(Array.unsafe-nth bs 0))))
+ 
+  (doc byte-seq->int8-seq-unsafe
```
hellerve: It seems like the doc uses the `-unsafe` prefix but the function doesn’t here. Is this deliberate?

core/Binary.carp
```diff
@@ -0,0 +1,111 @@
...
+
+  (doc system-endianness 
+    "Returns the endianness of the host system.")
+  (sig system-endianness (λ [] Order))
+  (defn system-endianness [] 
+    (if (= (system-endianness-internal) 1)
+        (Order.LittleEndian)
+        (Order.BigEndian)))
+
+  (doc bytes->int16-unsafe
```
hellerve: Should we also introduce safe variants of those functions?

hellerve: Sorry, didn’t remember that this is actually already in the todo list.

core/Binary.carp
scolsen: Yep, my mistake--thanks for catching this!

core/Binary.carp
scolsen: No prob! They're in :)

scolsen:
So, this wound up being pretty beefy! I think it's at a decent place for now--I do still want to implement a way to get an arbitrary number of bits from a byte sequence, but I think we can hold off till later.

Let me know what you think!

core/Binary.carp
scolsen: Removed this particular function since it's covered by StdInt!

eriksvedang:
Looks great. Are you happy too @hellerve ?

test/binary.carp
```diff
@@ -0,0 +1,145 @@
...
+                &(bytes->int16 (Binary.Order.LittleEndian) &[3b])
+                "bytes->int16 returns Nothing on insufficient data.")
+
+  (assert-equal test
+                &[(Uint16.from-long 2051l) (Uint16.from-long 776l)]
+                &(unsafe-bytes->int16-seq (Binary.Order.LittleEndian) &[3b 8b 8b 3b])
+                "Unsafe little endian bytes->int16-seq works as expected.")
+
+  (assert-equal test
+                &[(Maybe.Just (Uint16.from-long 776l)) (Maybe.Just (Uint16.from-long 2051l)) (Maybe.Nothing)]
```
hellerve: I’m not entirely sure this API is useful as is (since in order to check whether everything was decoded, you’ll have to check whether the last element is `Maybe.Nothing`). Maybe making the entire thing a `Maybe` makes more sense?

hellerve:
I added one comment, but other than that I’m pretty happy with this!

test/binary.carp
scolsen: A pretty typical idiom is to return the read value(s) + number of bytes read--or, alternatively, the number of excess bytes not read-- perhaps that works? Then we'd have:

`bytes->intN-seq (Fn [Order (Array Byte)] (Pair (Array UintN) Int))`

I think your suggestion works too. I'm open to any of the following signatures:

`(Fn [Order (Array Byte)] (Pair (Array UintN) Int))` -- some + number of bytes read or not read.
`(Fn [Order (Array Byte)] (Maybe (Array UintN))` -- all or nothing.
`(Fn [Order (Array Byte)] (Result (Array Byte))` -- all or nothing just like the maybe variant, but return the number of excess bytes for the failure case '

@hellerve @eriksvedang any preference? Of course, we could just support all variants--I could see each being useful under different circumstances.

hellerve: The number of bytes read is somewhat redundant, since we have a length property on the array, I think. I think I personally like the last version most; it seems like a local optimum for both safety and information content.

scolsen: > The number of bytes read is somewhat redundant, since we have a length property on the array, I think. I think I personally like the last version most; it seems like a local optimum for both safety and information content.

Fair point! 

I think returning the number read is a convenience thing, since the length of the resulting int array != the length of the number of read bytes. e.g. reading bytes into an array of 2 int16s requires an input of 4 bytes, reading into 2 int32s requires 8 bytes--the calculation is trivial, but it is convenient to just get it back automatically. It's more useful for cases where you have some N bytes and just want to read say, as many int16s as you can--if you have 5 bytes, you can only read 2 int16s, which is easy enough to figure out...but it's more convenient to just get the number from the function.

Still, I think bytes remaining (not read) is more useful information anyway for things like scanning files or working with streams--after some thought, I think both the `Pair` and `Result` variants are useful for different use cases. When you're scanning or streaming it's kind of painful to have to get your numbers exactly correct, and you often want to know how many bytes remain in order to determine how to/if you can continue processing the data (use `Pair`). In other circumstances, being unable to process some precise sequence of bytes is in fact catastrophic, and we can't really continue with our task (use `Result`).

For example, I'm going to use this to read terminfo files--the first section of those files are called a header, which is made up of 6 littleEndian int16s (12 bytes). 

Using `(Fn [Order (Array Byte)] (Pair (Array UintN) Int))` where the Int returned is bytes remaining, it'd be possible to read the entire terminfo file into a single byte array, read the first 12 bytes, and use `Pair.b` to see how many bytes remain in the file, and process those accordingly. This is convenient for a "greedy" approach in which we eagerly read in a file.

Contrarily, using `(Fn [Order (Array Byte)] (Result (Array Byte) Int)` still works, but instead of working with the entirety of the file, we need to be specific about the bytes we need and process each piece separately--if we read the whole file into a single byte array and passed it to this function, it'd result in a failure if it was longer than just the 12byte header. This is convenient for a stream based or "incremental" approach to processing the file.

I feel like the latter `Result` based function is appropriate when failed reads are not recoverable--for instance, in my example, if terminfo files *must* have a header, it makes sense to use the `Result` variant and stop if we fail.

On the other hand, the `Pair` variant is good for cases in which a failure is not catastrophic and you want to just keep chugging along processing all the data you can, or do something with the data remaining.

So, I'm a fan of both. I think the `Pair` version could be defined in terms of the `Result` version--so that may be the only one we need.

Sorry for the long post, wdyt?

hellerve: I think having both makes a lot of sense. I follow your reasoning completely: sometimes failed reads are recoverable, sometimes they aren’t, and having a specialized function for either work makes sense.

I’m not sure how the `Pair` version could be defined in terms of the other version; shouldn’t it be the other way around?

scolsen: I believe either can be taken as the primary definition because sum and product types are dual, and these types model sum and product types (pair being product and result being sum) (in categorical mumbo jumbo, pair is the product of some category C while Result is the coproduct).

...as to what actually makes the most sense in practice, I have no idea--I'll try to find the clearest solution. I'll update this PR later today!

Thanks for your comments! 

Thinking further about this, it would also be helpful to have a function that just returned the rest of the byte array (the unprocessed bytes `Fn [Order (Array Byte)] (Pair (Array Uint) (Array Byte))`--this is the shape of a lot of stream based scanning/parsing techniques, where parsing the next N items in the stream returns the parsed value + the remainder of the stream). Functions that let you decide how many bytes you want to process in a given byte array should also be useful--but I'll add these in later PRs.

scolsen: You were right, @hellerve! Defining `Results` in terms of `Pairs` is simpler

scolsen:
Alrighty, I've changed the sigs of the sequence conversion functions--mind taking one more look? 🍻

test/binary.carp
hellerve: Yay! Thanks for trying it out!

test/binary.carp
```diff
@@ -0,0 +1,163 @@
...
+  (assert-equal test
+                &(Pair.init [(Uint16.from-long 776l) (Uint16.from-long 2051l)] 1)
+                &(bytes->int16-seq (Binary.Order.BigEndian) &[3b 8b 8b 3b 2b])
+                "Big endian bytes->int16-seq works as expected.")
+
+  ;; We unwrap the error here since it's simpler than defining equality over
+  ;; arrays of Uint values.
+  ;; TODO: Define equality for arrays of non-ref values.
+  (assert-equal test
+                &(Result.unsafe-from-error (the (Result (Array Uint16) Int) (Result.Error 1)))
```
hellerve: This is sadly a common necessity in tests atm. Maybe assertions for `Maybe` and `Result` might be a good idea as well, like `assert-just` or `assert-success`; that’s out of scope for this PR though.

core/Binary.carp
```diff
@@ -0,0 +1,363 @@
...
+     If the conversion fails, returns a `Result.Error` containing the byte array passed as an argument.")
+  (defn byte-converter [f order]
+     (fn [bs]
+       (match (~f order bs)
+              (Maybe.Nothing) (Result.Error @bs)
+              (Maybe.Just i)  (Result.Success i))))
+  (doc interpreted
+    "Returns the interpreted value from a sequence of byte-converion results")
+  (private interpreted)
+  (defn interpreted [results]
```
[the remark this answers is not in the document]
scolsen: Ahh yes! good ol' fusion! Thanks for the tip!

Actually, it'd be really sweet if the compiler did automatic fusion for us...is that something you're interested in investigating :) I have several papers on fusion that I could pass along (not sure precisely whether or not any of them discuss actually implementing it in a compiler though)

core/StdInt.carp
```diff
@@ -179,6 +179,7 @@
   (register copy (Fn [&Uint16] Uint16))
 
   (defn prn [a] (Uint16.str a))
+  (defn zero [] (from-long 0l))
```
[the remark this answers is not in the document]
scolsen: Will do! In general, I try to implement the absolute minimum necessary to avoid unwitting commit/PR scope creep....but following this principle blindly does lead to some obvious omissions like in this case!

core/Binary.carp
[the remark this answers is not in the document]
scolsen: Soooo, I'm going to hold off on this until a later PR. There seems to be a slight bug in reduce whereby it's interpreting the function argument array as a literal array I'm getting:

```
Arrays can’t contain references: `x` at line 94, column 15 in '/Users/scottolsen/dev/Carp/core/Binary.carp'.

You’ll have to make a copy using `@`.
```

With the definition
```
(defn interpreted [results] 
  (let [f (fn [x y] (if (Result.success? y)
                                (Array.push-back x y)
                                x)]
    (Array.reduce &f [] results))
```

Where the function definition in the let is at line 94. There's also a good chance I'm just writing this incorrectly.

scolsen:
One last addition! Added zero for the other StdInt types. I'll add fusing in a follow-up once I can figure out what's goin on with Array.reduce or if I'm using it incorrectly.

test/binary.carp
scolsen: Thanks for the insight! Added #689 to track.

core/StdInt.carp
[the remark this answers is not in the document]
scolsen: See commit 32b61fb!

eriksvedang:
Thanks for this!

core/Binary.carp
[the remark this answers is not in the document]
hellerve: I think the problem is the way you use `reduce` and `push-back` in combination.

Their respective type signatures are:
```
Array.reduce : (λ [(Ref (λ [a, (Ref b c)] a d) e), a, (Ref (Array b) c)] a)
Array.push-back : (λ [(Array a), a] (Array a))
```

The `reduce` signature is a bit hard to read, but the main problem is the function that is passed in, so let’s zoom in on that:

```
(λ [a, (Ref b c)] a d)
```

Let’s remove the lifetime params, because they look a little messy:

```
(λ [a, (Ref b)] a)
```

The array element is passed in by reference! That makes sense, `Result.success?` takes a reference. But let’s look at `Array.push-back` again. It doesn’t take a reference, it takes the argument! Thus, `x` is inferred to be a `(Array (Ref (Result a b)))` (because we push references in), which is invalid!

Those are tricky to figure out, but basically you’ll have to add a copy in `push-back`, like so: `(Array.push-back x @y)`.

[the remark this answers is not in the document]
scolsen: Thanks for the detailed explanation! I was trying a couple of different variants, but didn't think of this one. I'll give it a shot!

Maybe we can better distinguish lifetimes from type params? I also find types that have some combination of both difficult to read at times.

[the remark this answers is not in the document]
eriksvedang: Yeah, any ideas for how to make them stand out (like the tick in Rust) would be greatly appreciated. I just haven't come up with any good ideas...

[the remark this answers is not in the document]
scolsen: It's tricky! 

I kind of like alligator brackets, it takes up a little extra space, but it makes them really obvious:

`(λ [(Ref (λ [a, (Ref b <c>)] a <d>) <e>), a, (Ref (Array b) <c>)] a)`

I think `{}` could be nice too, but it's harder to read than `<>`:

`(λ [(Ref (λ [a, (Ref b {c})] a {d}) {e}), a, (Ref (Array b) {c})] a)`

For single character options, I think stuff that's equally obvious is best:

`(λ [(Ref (λ [a, (Ref b *c)] a *d) *e), a, (Ref (Array b) *c)] a)`
`(λ [(Ref (λ [a, (Ref b #c)] a #d) #e), a, (Ref (Array b) #c)] a)`

Or you could overload the `&` or `@` which would make it clear that it's specific to references:

`(λ [(Ref (λ [a, (Ref b &c)] a &d) &e), a, (Ref (Array b) &c)] a)`
`(λ [(Ref (λ [a, (Ref b @c)] a @d) @e), a, (Ref (Array b) @c)] a)`

But that could be confusing, too.

My personal preference is for `<a>`-- not sure if we're using that to mean something else already, though.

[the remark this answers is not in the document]
hellerve: Yeah, alligator brackets (I always called them angle brackets, but this name is way cooler, thanks @scolsen) look good to me!

PR #691, opened 2020-02-21
Add new test primitives and ref=

### LINKED ISSUE #689: [Feature Request] Better assertions in Test
There are currently some limitations in the Test module that make it slightly tedious to write some tests. Namely, having to use refs in a lot of places, and lack of support for core Sumtypes such as `Maybe` and `Result`.

`Maybe` we can improve the support for core types?

hellerve:
This PR fixes #689 by introducing:

- `ref=` for reference equality (currently only used in `Maybe` and `Success`, but maybe we should make this a more general idiom?),
- `Test.assert-just` for testing if something is a `Maybe.Just`,
- `Test.assert-nothing` for testing if something is a `Maybe.Nothing`,
- `Test.assert-success` for testing if something is a `Result.Success`,
- `Test.assert-error` for testing if something is a `Result.Error`, and
- `Test.assert-ref-equal` for testing if two things are `ref=`.

This already simplifies some code in our tests, and should make things simpler in the future generally.

Cheers

### DIFF
--- a/core/Result.carp
+++ b/core/Result.carp
@@ -113,7 +113,7 @@ but `(Success 1)` and `(Success 2)` are not.")
 constructor and the contained value, i.e. `(Success [1])` and
 `(Success [1])` are the same, but `(Success [1])` and `(Success [2])` are
 not.")
-  (defn = [a b]
+  (defn ref= [a b]
     (match @a
       (Success x)
         (match @b
--- a/core/Result.carp
+++ b/core/Result.carp
@@ -95,7 +95,9 @@ It is the inverse of [`success?`](#success?).")
       (Error _) true
       (Success _) false))
 
-  (doc = "equality is defined as the equality of both the constructor and the contained value, i.e. `(Success 1)` and `(Success 1)` are the same, but `(Success 1)` and `(Success 2)` are not.")
+  (doc = "equality is defined as the equality of both the constructor
+and the contained value, i.e. `(Success 1)` and `(Success 1)` are the same,
+but `(Success 1)` and `(Success 2)` are not.")
   (defn = [a b]
     (match @a
       (Success x)
@@ -106,6 +108,21 @@ It is the inverse of [`success?`](#success?).")
         (match @b
           (Success _) false
           (Error y) (= x y))))
+
+  (doc ref= "reference equality is defined as the equality of both the
+constructor and the contained value, i.e. `(Success [1])` and
+`(Success [1])` are the same, but `(Success [1])` and `(Success [2])` are
+not.")
+  (defn = [a b]
+    (match @a
+      (Success x)
+        (match @b
+          (Error _) false
+          (Success y) (= &x &y))
+      (Error x)
+        (match @b
+          (Success _) false
+          (Error y) (= &x &y))))
 )
 
 (defmodule Maybe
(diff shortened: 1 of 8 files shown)


core/Result.carp
```diff
@@ -106,6 +108,21 @@ It is the inverse of [`success?`](#success?).")
...
+  (defn = [a b]
+    (match @a
+      (Success x)
+        (match @b
+          (Error _) false
+          (Success y) (= &x &y))
+      (Error x)
+        (match @b
+          (Success _) false
+          (Error y) (= &x &y))))
```
scolsen: Should these and other ref comparisons use `ref=` now? And should we implement the interface for other types?

hellerve: I guess this is the question I was asking myself while doing this (and the world at large when writing the PR description). The problem is that equality is on references by default for strings and it would probably be a big breaking change to change this to `ref=`, but I think that would probably be the best way to go in the future? Implementing `=` twice in different modules for the same type has always been kind of annoying for me.

core/Result.carp
```diff
@@ -106,6 +108,21 @@ It is the inverse of [`success?`](#success?).")
         (match @b
           (Success _) false
           (Error y) (= x y))))
+
+  (doc ref= "reference equality is defined as the equality of both the
+constructor and the contained value, i.e. `(Success [1])` and
+`(Success [1])` are the same, but `(Success [1])` and `(Success [2])` are
+not.")
+  (defn = [a b]
```
jacereda: Shouldn't this one read `ref=`?

hellerve: It should, and now it does! Good catch!

eriksvedang:
Sorry for not merge:ing this yet. I'm a little wary of making `ref=` an interface, it seems pretty ad-hoc to me. But I guess it must be to make this useful in the testing context? 

What I'm most afraid of is making people accidentally start using equality operators that performs copying. Forcing people to use refs for equality in all cases so far has been a very nice design, I think. 

I guess what I'm asking is what it would mean to implement `ref=` for other types? Perhaps what this really is about is the lack of `match`:ing on refs..? (which I could prioritize to get done soon)

hellerve:
I’m going to respond only for the question at the end for now: matching on non-refs wouldn’t make much of a difference, because we need to check the equality of member types by reference, and those cannot be references as of now. Basically the problem boils down to this:

```clojure
(= &(Maybe.Just 1) &(Maybe.Just 2)) ; works fine, since equality is defined as (Fn [Int Int] Int)
(= &(Maybe.Just [1]) &(Maybe.Just [2])) ; does not work, since equality is on references
```

This has lead to a lot of head-scratching on my end, and it makes it impossible to write some tests without the `<Type>.from` escape hatch, which I’d like to avoid.

eriksvedang:
OK that clarifies it a bit, thanks. So what types would benefit from having the interface implemented for them? (Only "containers", right?)

hellerve:
I think it’d generally be nice to know by the function of equality whether it is a ref or not (though I presume the default should be the other way around: `=` works on equality, `copy=` works on non-refs).

The main problem that I see is that some standard types (numerics) have different definitions of equality than the rest of the world. I guess having equality work only on references canonically could also work.

eriksvedang:
How is this PR affected by the new `match-ref` construct? @scolsen

hellerve:
I will close this for now. I’ll pull out the new `assert-*` constructs, though, because they seem useful to me.

eriksvedang:
Sounds good!

Issue #673, opened 2020-02-10
defdynamic in modules lead to errors in documentation generation


hellerve:
When generating documentation for modules that internally use `foreach` (which as of #656 emits `gensym`s) we encounter errors related to some internal macro code. The error message always reads as follows:

```
carp: Unhandled binder: XObj {obj = Lst [XObj {obj = DefDynamic, info = Nothing, ty = Nothing},XObj {obj = Sym RESP.*gensym-counter* Symbol, info = Nothing, ty = Nothing},XObj {obj = Num Int 1001.0, info = Just (Info {infoLine = 0, infoColumn = 0, infoFile = "dummy-file", infoDelete = fromList [], infoIdentifier = -1}), ty = Just Int}], info = Just (Info {infoLine = 361, infoColumn = 34, infoFile = "/Users/veitheller/Documents/Code/Github/carp/carp/core/Macros.carp", infoDelete = fromList [], infoIdentifier = 0}), ty = Just Dynamic}
CallStack (from HasCallStack):
  error, called at src/Obj.hs:205:26 in CarpHask-0.3.0.0-HIw6hfOW0ih6jzjLFsz5xF:Obj
```

The module of course changes. What’s weird about this is that the dynamic `*gensym-counter` seems to be inlined into the module; I didn’t know that happened. This can be fixed by patching the code that fails to handle `defdynamic` objects as well, and the `defdynamic` will be visible in the generated documentation. The symbol should not however be inlined in the module in the first place.

I'm not quite sure how to best resolve this since I don’t know where the dynamic gets inlined into the enclosing module but I assume it happens somewhere in the macro code in `Eval.hs`.

Cheers

eriksvedang:
Interesting..!

eriksvedang:
Is this solved now @hellerve ?

hellerve:
Yeah, this should be fixed.

Issue #659, opened 2020-01-30
Local variables can't share names with commands/macros/functions in the dynamic context.


sdilts:
When trying to use a name for a local variable that has the same name a command, macro, or dynamic function in the dynamic context, Carp doesn't overshadow that global variable or provide warnings. Changing the variable name to something else fixes the problem.

For example, you cannot use `c` as a variable name, as it is a function.
```  clojure
(let [c (car (list 1 2 3))]
     (list c))

= > invalid args to 'c' command
```
Issue #555 suggests we wrap the result in `(quote ...)` while using the REPL, which results in:
```  clojure
(quote (let [c (car (list 1 2 3))]
     (list c)))

= > ((command c))
```
If we use a variable name that is not a function, everything works fine:
```  clojure
(quote (let [foo (car 0 (list 1 2 3))]
     (list foo)))
=> (1)
```

This is the case for code run in the REPL and while compiling a file. If you place the following code into a file, then load it and look at `test-val`'s value, you will see that the same thing occurs.
``` clojure
(defdynamic test-val (let [c (car (list 1 2 3))]
                        c))
```
You can try this with `car`, `e`, `nthcdr`, `nthcar`, etc. and a similar result occurs.

sdilts:
Possibly related to #597

hellerve:
I suspect this is also the root cause of the weird issues we’re seeing in #460.

As discussed in #555, we’ll have to rewrite the evaluator ina  more principled manner; hopefully that will fix a bunch of these issues all at once.

eriksvedang:
Nice find @sdilts! But yes, what @hellerve said. I'll leave this open though, as a reminder.

eriksvedang:
Is this fixed now because of #695 being merged?

eriksvedang:
OK, it's not a giant problem but I'll keep the issue open then.

scolsen:
As @hellerve stated, this ultimately comes down to our lookup orders. 

Ideally, preferring local environments before the Dynamic environment wouldn't cause any problems and could fix this issue. I'll give it a shot.

scolsen:
I have a fix for this...but oddly it seems to make `postwalk`, and co. unacceptable performance wise. It doesn't seem to have any other effect, though.

scolsen:
In particular, it makes the performance of walk-replace prohibitve and breaks the behavior of postwalk. Prewalk is strangely unaffected as are all other functions and test thus far—so the solution doesn’t account for an edge

[...]

Issue #555, opened 2019-09-12
Dynamic evaluator evaluates too much


eriksvedang:
There is something wrong with the implementation of the dynamic evaluation of Carp code. Expressions returning things that can be evaluated (i.e. lists) are evaluated an "extra" time, which is often wrong. This extra step needs to be removed, and the reason it was added there investigated.

One simple example of how this bug manifests is that if you enter a quoted expression in the repl it will evaluate:
```
> '(+ 2 3)
=> 5
```

scolsen:
So, straight up removing the hack and returning the evaluated form as is has the following effects:

1. When starting the repl, a bunch of meta-sets are echoed:

```clojure
-> (meta-set! print-doc "doc" "Print the documentation for a binding.")
-> (meta-set! sig "doc" "Annotate a binding with the desired signature.")
-> (meta-set! print-sig "doc" "Print the annotated signature for a binding.")
-> (meta-set! hide "doc" "Mark a binding as hidden, this will make it not print with the 'info' command.")
-> (meta-set! private "doc" "Mark a binding as private, this will make it inaccessible from other modules.")
-> (meta-set! todo "doc" "sets the todo property for a binding.")
-> (meta-set! private? "doc" "Is this binding private?")
-> (meta-set! private? "todo" "This is buggy, will report true when meta is set to 'false'!")
-> (meta-set! hidden? "doc" "Is this binding hidden?")
-> (meta-set! hidden? "todo" "This is buggy, will report true when meta is set to 'false'!")
-> (meta-set! annotate "doc" "Add an annotation to this binding.")
-> (meta-set! defproject "doc" "Define a project configuration.")
-> (meta-set! const-assert "doc" "asserts that the expression `expr` is true at compile time.\nOtherwise it will fail with the message `msg`.\n\nThe expression must be evaluable at compile time.")
...many more meta-set! forms.
```
2. The vast majority of cases are unproblematic. Also, one no longer needs to quote certain forms that one needed to quote before:

```clojure
(list 'foo)
; => (foo)

;; contrasted with carp as it stands right now, executing 
(list 'foo)
;; results in 
; I couldn’t find the symbol 'foo' at line 1, column 8 in 'REPL'.

Maybe you wanted one of the following?
    Bool
    doc
    fmt
    for
    not
    or
    todo 
;; the extra eval we're trying to fix here.
```

[...]

PR #695, opened 2020-02-29
New evaluator

hellerve:
This PR is meant to track the progress of rewriting the evaluator. Currently this is very much a work in progress and pretty messy, but it will get better I’m sure.

### Goals

The goals of this evaluator are to move away from an ad-hoc, not very well defined evaluator to a well defined, simple, but extensible version. The evaluator should be specified using a design document.

While we’re at it, we should also resolve some long-standing issues with the old evaluator.

We want to fix the following issues at least (let me know if I missed any):
- #555 
- #673
- #659

### Implementation

This implementation adds multiple changes to the IR as well as to the evaluator.

#### IR

We add a `Closure` type that wraps dynamic functions et al with their environment to add lexical scoping. This solves the weird dynamic-scoping problems that we occasionally had before.

#### Evaluator

The evaluator moves towards extensibility by adding primitives as well as commands. One gets evaluated arguments, while the other one does not. This means that special commands such as `def*`, `type`, `info`, etc. can be expressed as functions instead of having to be part of the evaluator loop. This makes it easier to extend the set of commands and reduces the amount of  duplicated work.

We also move away from double evaluation in the REPL (which was extremely error-prone) to recursive macro expansion (still a WIP), which is a little more standard and well-understood.

### State

Currently, I estimate that about 2/3 of the production work are done, after which we’ll have to clean up the code, reorganize it, and finally test it.

An itemized todo list:
- [x] Refactor evluator loop
- [x] Add primitives
- [x] Add `Closure`
  - [x] Make sure closures are unwrapped properly everywhere (especially in the emitter and qualifier)
- [x] Get to feature parity of evaluator loop
- [x] Add argument capability to commands
- [x] Recursively expand macros
- [x] Add  evaluator tracebacks
- [x] Make REPL autocomplete update with every loop
- [x] Clean up
- [ ] Validate

Cheers


eriksvedang:
First of all – great job! This is truly awesome and so much better than what was before.
Here's some feedback after trying it out (I've lettered them to make it easier to respond to a specific one:

- [A] When the top-level result is Unit (empty list) it shouldn't be printed.

- [B] Pasting a whole code block (with multiple top-level forms) into the repl doesn't work (only the first one is evaluated). I prefer if they all print their results (except if it's Unit).

- [C] Entering a symbol at the top level should (probably) just echo the name and it's type (using 'showBinderIndented' or something like that).

- [D] Situation 3 does not give the expected result, calling 'foo' should evaluate to () and define 'a' to be 3.5 but instead it returns (if true (def a 3.5) (def a 10)). Wrapping foo in an extra set of parens leads to infinite loop in the repl, it seems.

- [E] Situation 4 did not do what I expected, defining f leads to no output while calling it with (f) gives both outputs. Perhaps this is correct though, I'm not sure...

- [F] Situation 5 didn't do what I expected, (m true 'AHA 'BOO) returns (if x b a), but perhaps that's exactly what it SHOULD do?

- [G] Would be nice to also be able to use :i as a shortcut for (info <symbol>), now that :t works!

- [H] This is more of an open-ended question: How hard is it do get back the behaviour (but in a principled way) of being able to call static functions from the repl? Right now calling (defn f [] 123) at the repl like so: (f) returns ((defn f [] 123)) which is bewildering. There is (or were) a Haskell function called 'executeFunctionAsMain' for doing this call more or less intelligently (redefining main to call the function, execute the program, and then restore main to what it was before). Ideally this would work for global variables too. It's a whole can of worms but I think it's something everyone wants and I had started to rely a bit on it already, despite it's buginess in the old Eval.

hellerve:
Thank you for the kind words and the feedback @eriksvedang, much appreciated! I’ll address the easy concerns first:

[A], [C], [G], and [H] should now be fixed. Commit 60e4d337 showcases how easy adding a new REPL shortcut is now.

[B] will be fixed soon.

I’ll go through [D-F], which are concerned with the examples you defined in #697 (thanks again for that) in a follow-up comment.

hellerve:
D) and E) boil down to the same question: should dynamic functions expand macros on definition or should they “interoperate”? Last we talked I think we decided the only difference between dynamic functions and macros should be that macros do not evaluate their arguments. We can change that to macros expanding in-place in dynamic functions, which would enable a different kind of interoperation. Let’s talk about this in more depth next time we meet?

About [F]: I think this should be fixed, because this is definitely not how other macro systems work. I’ll look at it sometime this week.

eriksvedang:
Sweet! Yes, adding `:i` was quite easy indeed :)

hellerve:
Since some time has passed since the last update in here, I decided it might be time to write up what has happened.

I decided to take on #696 as part of this work as well, which has proved harder than expected. I’ve found the general problem—`do` reused the old context for all statements inside it, and `StateT` was too lazy for us. I fixed the first problem, and am mostly done with rewriting from a `StateT` to explicit context-passing. So far, so good.

However, it turns out that fixing `StateT` required me to think about how to fix [F], and I realized that there is a fundamental flaw in how we think about macro expansion. Let me elaborate.

Consider this macro from `core/Macros.carp`:

```clojure
(defmacro and [x y]
  (list 'if x y false))
```

Now imagine we use it like this:

```clojure
(defn between [x l u]
  (and (>= x l) (<= x u)))
```

This will first expand to:

```clojure
(defn between [x l u]
  (if (>= x l) (<= x u) false))
```

but the evaluator cannot determine whether it is done, because `if` exists both at compile time and at run time. It will thus try to expand it once again, and fail.

Why does it have to evaluate again? Because some macros generate macros, and we have to account for those. We have to recurse. It turns out that the double evaluation was a problem from the get-go, but now that I’m rewriting the evaluator for the—second? third?—time and doing it properly I realize that its repercussions run deeper than I first understood.

The “obvious” solution would be to separate between the evaluator and the "meta langauge" (defining types etc.), but this would require yet another rewrite, and U’m not quite sure whether this would lead me anywhere where I would care to be.

If anyone has any input, that’d be great. I’m pretty stumped right now (and I dont understand why this has not lead to problems earlier). My current working theory is that there is a fundamental incapability to express what we want to express in the last few examples of #697 without a more principled separation of concerns at compile time.

Maybe I’m also just talking nonsense, I’ve been staring at macro expansion for a very long time.

My interim work for rewriting state handling can be inspected in [my branch `where-is-this-going`](https://github.com/hellerve/Carp/tree/where-is-this-going), which is an apt name for my current state of mind.

Cheers

hellerve:
I’ve already shared this information in the Gitter channel, but the rewrite seems to work now! We are ready for a code review while I will go ahead and write some documentation. This will probably still take a minute, since there is a lot that could possibly be written about! I’ll try to keep it manageable though.

eriksvedang:
Awesome work.

PR #703, opened 2020-04-02
Fix args test

hellerve:
This PR fixes a bug discovered by @TimDeve. One of the tests in `test/args.carp` constructed the path to the executable erroneously when there was no trailing path separator.

Cheers

### DIFF
--- a/test/args.carp
+++ b/test/args.carp
@@ -2,7 +2,10 @@
 (use Test)
 
 (defndynamic _make-exe-path [pth]
-  (String.join (array (Project.get-config "output-directory") pth)))
+  (let [out (Project.get-config "output-directory")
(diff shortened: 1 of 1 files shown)


hellerve:
I haven’t tested it on Windows, but it should work. Additional tests would be appreciated.

eriksvedang:
I'll test on Windows at some point...

TimDeve:
Fails for me as `(os)` returns `mingw32`, I think line 6 should be:
`sep (if (or (= (os) "windows") (= (os) "mingw32")) "\\" "/")`

hellerve:
Ah, I should add that, thanks for the feedback!

hellerve:
Filed #705 to fix.

PR #704, opened 2020-04-03
Doesn't compile assert-exit and assert-signal on windows

TimDeve:
Context: https://github.com/carp-lang/Carp/pull/244#issuecomment-391382849

As there is no script to run the test suite in windows, is it alright to remove these assert until there is a window implementation (if that is possible)?

### DIFF
--- a/core/Test.carp
+++ b/core/Test.carp
@@ -78,7 +78,7 @@
   (defn assert-exit [state exit-code f descr]
     (assert-equal state exit-code (run-child f) descr))
 
-  (doc assert-exit "Assert that function f aborts with OS signal signal.")
+  (doc assert-signal "Assert that function f aborts with OS signal signal.")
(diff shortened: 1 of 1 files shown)


TimDeve:
Should there be a `(not-on-windows)` on each function or just is fine?

hellerve:
I  think one is fine, but maybe we could add something like this:

```clojure
(windows-only
  (defndynamic assert-exit [state exit-code f descr]
      (macro-error "assert-exit is not implemented on Windows."))

  (defndynamic assert-signal [state signal x descr]
    (macro-error "assert-signal is not implemented on Windows."))
)
```

to get a more useful error message?

TimDeve:
I have ammended my commit with your suggestion.

eriksvedang:
Cool solution replacing the defn:s with error messages, I like it!

PR #706, opened 2020-04-04
Adds Github Actions CI to repository

TimDeve:
This PR adds Github Actions to the repo as a mean to automatically build the Carp compiler and run tests.

Right now it only runs `./run_carp_tests.sh` as `stack test` is failling, it also skips the sdl tests as I wasn't sure what dependencies are required.

I also had to make changes the to the tests as the load call to StdInt.carp was failing on case-sensitive systems like the CI.

The CI will run on push to master and when someone raise a PR against master, unsure if the second part is alright...

### DIFF
--- a/.github/workflows/carp.yml
+++ b/.github/workflows/carp.yml
@@ -15,11 +15,15 @@ jobs:
 
     - uses: actions/setup-haskell@v1
 
+    - run:   echo ::set-env name=GITHUB_SHA::$GITHUB_SHA
+      shell: bash
+
     - uses: actions/cache@v1
       name: Cache ~/.stack
       with:
         path: ~/.stack
-        key: ${{ runner.os }}-stack
+        key: ${{ runner.os }}-stack-${{ env.GITHUB_SHA }}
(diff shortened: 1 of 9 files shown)


.github/workflows/carp.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-haskell@v1
+
+    - uses: actions/cache@v1
+      name: Cache ~/.stack
+      with:
+        path: ~/.stack
+        key: ${{ runner.os }}-stack
```
bjorn3, line 22: Caches are immutable meaning that future changes to.`~/.stack` will not be stored.

TimDeve: Ah, thanks. I'll look into that.

TimDeve: @bjorn3 This should fix it:
https://github.com/carp-lang/Carp/commit/f081992830618565c2523767d9ad0abd9fa07141

PR #707, opened 2020-04-04
Create CI for Windows

TimDeve:
This PR creates a Powershell script so that the test suite can be run on Windows.

It's just a draft for now to get some feedback.

Aside from the issues in https://github.com/carp-lang/Carp/pull/705 and https://github.com/carp-lang/Carp/pull/704 the only failing tests seems related to int64 and uint64 probably because of https://github.com/carp-lang/Carp/pull/577, should these tests be skipped on Windows for now?

The script also doesn't include the output checks for errors and the examples, I'm looking at adding this later.

My goal with this script is to hopefully get Windows CI running similar to https://github.com/carp-lang/Carp/pull/706 so that regressions on Windows can be detected earlier.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,55 @@
+name: Windows CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    steps:
+      - name: Check out
+        uses: actions/checkout@v2
+
+      - name: Save git commit SHA 
+        run: echo ::set-env name=GITHUB_SHA::%GITHUB_SHA%
+        shell: cmd
+
+      - name: Install Scoop
+        run: |
+          iwr -useb get.scoop.sh | iex
+          echo "::add-path::~\scoop\shims"
+          echo "::add-path::C:\ProgramData\scoop\shims"
+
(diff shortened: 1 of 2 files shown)


eriksvedang:
This is great, thanks so much for getting the ball rolling!

eriksvedang:
When trying to run this I get an error from stack about the args after `--` though:

```
Invalid option `-x'

Usage: stack.exe exec CMD [-- ARGS (e.g. stack exec -- ghc-pkg describe base)]
                      ([--plain] | [--[no-]ghc-package-path] [--[no-]stack-exe]
                      [--package ARG] [--rts-options RTSFLAG] [--cwd DIR])
                      [--help]
  Execute a command
``` 

Passing args to stack doesn´t seem to work for me, is there something I need to do?

eriksvedang:
Just passing a path seems to work, it's the `-x` flag that is sent to stack even though it's after the `--` it seems...

TimDeve:
Interesting... What's your environment? Windows 10 and Powershell?

eriksvedang:
Updating Stack solved it.

TimDeve:
Added the CI as part of this PR. Main blocker seems to be https://github.com/carp-lang/Carp/pull/577. I tried to disable the tests using `(not-on-windows)` but it seems to interact badly `(deftest)`.

The examples and error check are still missing but I see values in merging without them.

PR #710, opened 2020-04-09
Add functions to: turn Ptr into value & prevent destructor from being run 

TimDeve:
Adds two functions:

Pointer.to-value turns a Ptr into an owned value

Unsafe.leak consumes a value and prevents it from being freed by Carp

This is a draft because there isn't any tests yet and I'm not sure this is something that people want.

Combining the two allows to extract values out of a struct without copy.

Example:
```clojure
(defn to-ptr [r] (the (Ptr a) (Unsafe.coerce (the (Ref a) r))))

(defn unsafe-value [a] (Pointer.to-value (to-ptr a)))

(defn main []
  (let [p (Pair.init @"Hello" @"Sup")]
    (let [a (unsafe-value (Pair.a &p))
          b (unsafe-value (Pair.b &p))
          _ (Unsafe.leak p)]

      (println* (String.concat &[a b])))))
```

### DIFF
--- a/docs/core/Pointer.html
+++ b/docs/core/Pointer.html
@@ -330,6 +330,26 @@
                 </p>
             </div>
             <div class="binder">
+                <a class="anchor" href="#to-value">
+                    <h3 id="to-value">
+                        to-value
+                    </h3>
+                </a>
+                <div class="description">
+                    template
+                </div>
+                <p class="sig">
+                    (λ [(Ptr a)] a)
+                </p>
+                <span>
+                    
+                </span>
+                <p class="doc">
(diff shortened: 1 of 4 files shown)


TimDeve:
Does this need tests?

TimDeve:
Added a test for `Pointer.to-value`, I don't know how `Unsafe.leak` could be tested.

Is the description correct for `Unsafe.leak`, I don't if the term 'destructor' is correct in particular.

TimDeve:
Added test for `Unsafe.leak`

eriksvedang:
Awesome, looks great! Thanks @TimDeve !!!

PR #711, opened 2020-04-12
Add immutable macro, update Emit to handle const

scolsen:
Generally speaking, this change enables the annotate macro for Def
forms.

This commit adds a new macro that's short hand for a "const" annotation
in C. This effectively marks a variable as immutable (subsequent set)
calls will produce invalid C. There's no checking of the immutability
within carp itself, Carp will emit the C as is, which will fail to
compile if a def declared immutable is mutated at some point in the
program.

I've also had to update Emit.hs to handle `const` defs as a special
case. Carp's current strategy is to declare the types of names before
actually initializing their values, and to assign values later in a
special init function e.g.

```c
int foo;
// ...
carp_init_globals() {
  // ...
  foo = 4;
}
```

This is problematic for const qualified types, since C will initialize
them to a default value, then consider the subsequent initialization
emitted by Carp to be an error:

```c
int const foo;
// ...
carp_init_globals() {
  // ...
  foo = 4; // ERROR!
}
```

So, we have to handle const explicitly, and change the emitters behavior
for these cases. This commit handles that, and the emitter will emit the
  following for const annotated defs:

```c
int const foo = 4;
```

At the moment, there is a duplication issue--we still emit the
assignment in the init function, and I'd welcome solutions to resolving
this. So, what we really emit for consts, as of this commit is:

```c
int const foo = 4;
// ...
carp_init_globals() {
  // ...
int const foo = 4; // OK
}
```

This emulates the desired behavior whereby subsequent `set` calls will
result in invalid c, i.e.

```
(def foo 4)

(immutable foo)

(defn main []
  (set! foo 5)
)
```

Results in:

```c
int const foo = 4;

carp_globals_init() {
  // ...
  int const foo = 4;
}

main() {
  foo = 5; // ERROR!
}
```

So, the duplication in the init function is a-ok, at least according to
my compiler. Contrarily adding the qualifier later doesn't work and has
no actual effect.

```c
int foo;

carp_init_globals() {
  // ...
int const foo = 4;
```

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -43,6 +43,16 @@
 (defmacro annotate [name annotation]
   (list 'meta-set! name "annotations" (annotate-helper name annotation)))
 
+(doc immutable
+  "Annotate a binding as immutable. Emits the correpsonding c code for the
+  binding qualified with `const`.
+
+   Carp itself performs no checks against such bindings, rebinding a value
+   annotated as immutable is valid in Carp, but it will result in invalid
+   reassignment in the emitted C, which won't compile.")
+(defmacro immutable [name]
+  (list 'meta-set! name "annotations" (annotate-helper name "const")))
+
 (defmodule Dynamic
   (defndynamic caar [pair] (car (car pair)))
   (defndynamic cadr [pair] (car (cdr pair)))
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -8,7 +8,7 @@ module Emit (toC,
              wrapInInitFunction
             ) where
 
-import Data.List (intercalate, sortOn)
+import Data.List (intercalate, sortOn, isInfixOf)
 import Control.Monad.State
 import Control.Monad (when, zipWithM_)
 import qualified Data.Map as Map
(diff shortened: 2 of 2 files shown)


scolsen:
@eriksvedang @hellerve turns out adding support for const qualifiers was slightly more complicated than I expected. I wanted to be able to mark defs as immutable variables, so I enabled annotate for Def forms, but it turns out we need special handling since C doesn't allow a naked const declaration followed by assignment -- or at least my compiler doesn't. It considers subsequent assignment an error

lmk if you need more info! Hopefully the pr op is detailed enough.

Also looking for cleaner ways of doing this--I just made the first possible alterations to the Emitter that I could make work, which results in some hairy duplication.

hellerve:
This is an interesting change and it’s very cool to see how we could implement it. In the long run I’d like Carp to be able to check for this itself, since this will likely result in better error messages, especially when it comes to source locations.

My main question is therefore: should we add this and expect it to be an interim solution? Or should this be the final solution?

scolsen:
> This is an interesting change and it’s very cool to see how we could implement it. In the long run I’d like Carp to be able to check for this itself, since this will likely result in better error messages, especially when it comes to source locations.
> 
> My main question is therefore: should we add this and expect it to be an interim solution? Or should this be the final solution?

I'm in full agreement! If we do merge this, I'd expect it to be temporary until we implement a Carp 'native' solution for the reasons you mentioned.

Alternatively, we could wait until your Eval changes are in and add `(immutable)` as a primitive and handle it that way (which would let us do the assignment checking and error reporting on the Carp side).

hellerve:
I think having it as a primitive sounds appealing, but we’ll probably still have to have it embedded in other parts of the compiler as well, to make sure the variable doesn’t get reassigned later (think for instance the REPL).

eriksvedang:
This looks solid, but yes, it seems weird to let the C compiler handle it. At the very least someone should prototype how it could be done in the Carp compiler and see how complicated it is, before potentially merge:ing this PR as a temporary solution.

eriksvedang:
I think we should try adding information regarding immutability on a deeper level in the compiler so that we can give proper error messages and the like.

PR #712, opened 2020-04-17
Add inner workings to macro docs

hellerve:
This PR completes the documentation work started in #695 by adding a (fairly extensive) section on the inner workings of the new evaluator. It doesn’t aim to document so much how it works but how to read it.

Cheers


docs/Macros.md
```diff
@@ -98,4 +98,103 @@ the resulting code. `macro-log` is useful for tracing your macro, a form of
...
+```
+
+Thus, to understand it, we’ll have to understand at least `Context`, `XObj`,
+and `EvalError`. The types `IO` and `Either` are part of the Haskell standard
+library and will not be covered extensively—please refer to your favorite
+tool for Haskell documentation (we recommend [Stackage](https://stackage.org))
+to find out more about them.
+
+All data structures that are discussed here are defined in
+[`src/Obj.hs`](/src/Eval.hs).
```
scolsen, line 133: Looks like the link here is pointing to Eval instead of Obj!

docs/Macros.md
```diff
@@ -98,4 +98,103 @@ the resulting code. `macro-log` is useful for tracing your macro, a form of
...
+illustrates the flow of the compiler.
+
+![The compiler passes](./compiler-passes.svg)
+
+The dynamic evaluator is arguably one of the most central pieces of the Carp
+compiler. It orchestrates macro expansion, borrow checking, and type inference,
+as it encounters forms that have requirements for these services, such as
+function definitions, variables, or `let` bindings.
+
+As such, the evaluator serves many purposes. While you might start out trying
```
scolsen, line 111: Might be able to pare this down to just "Understanding the evaluator will give you a lot of insight into how Carp works."

I think the rest mostly restates the points of the previous paragraph.

docs/Macros.md
```diff
@@ -98,4 +98,103 @@ the resulting code. `macro-log` is useful for tracing your macro, a form of
...
+library and will not be covered extensively—please refer to your favorite
+tool for Haskell documentation (we recommend [Stackage](https://stackage.org))
+to find out more about them.
+
+All data structures that are discussed here are defined in
+[`src/Obj.hs`](/src/Eval.hs).
+
+#### `XObj`
+
+`XObj` is short for “`Obj` with eXtras”. `Obj` is the type for AST nodes in
```
scolsen, line 137: Amazing. This sort of information is just what I need, for example, to add proper immutability checking :)

docs/Macros.md
hellerve: I condensed it, thank you for pointing this out!

docs/Macros.md
hellerve: Whoops, this should be fixed now!

hellerve:
@scolsen thanks for reading through it and for your kind words! I’ve incorporated your feedback!

PR #716, opened 2020-04-18
Guard against calling main directly

### LINKED ISSUE #713: Eval buffer in REPL does not emit main(), fails to compile
This issue is seen as of SHA 84403988

With reference to `examples/arrays.carp`

Eval buffer in REPL (bound to `C-c C-l` in emacs) fails to generate main() in examples/out/main.c causing the clang compile step to fail. 
* This is independently reproducible and has been confirmed by @hellerve
* I can call (ex1) and (ex2) from the REPL and they work
* However, calling (main) from the REPL fails and the inferior carp process exits
* The `examples/out/main.c` file shows ex1() and ex2() generated along with the structs and includes, but does not have main()

I can further confirm that:
* `carp -x examples/arrays.carp` works and produces the expected output.
* Build-and-run (bound to `C-c C-c` in emacs) works and produces the expected output, in the inferior Carp process. 
* The fat binary generated by build-and-run at `example/out/Untitled` also successfully runs standalone and produces the expected output.

hellerve:
This PR fixes #713 by adding a check to calling a function at the REPL: if that function is `main`, we emit an error.

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -409,6 +409,9 @@ executeCommand ctx@(Context env _ _ _ _ _ _ _) xobj =
          reportExecutionError newCtx (show e)
          return (xobj, newCtx)
        -- special case: calling a function at the repl
+       Right (XObj (Lst (XObj (Lst (XObj (Defn _) _ _:(XObj (Sym (SymPath [] "main") _) _ _):_)) _ _:_)) _ _) -> do
+        reportExecutionError newCtx "`main` should not be called from the REPL. Use `(build)` and `(run)` instead."
+        return (xobj, newCtx)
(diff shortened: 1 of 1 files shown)


eriksvedang:
Maybe calling `(main)` could mean just `(do (build) (run))` then?

hellerve:
That could also work, though I’m not sure which one is better UX.

eriksvedang:
Most people would want the main function to be called rather than getting an error, I'm sure!

hellerve:
Fair enough! I’ll update this.

hellerve:
Done!

eriksvedang:
Very nice, I like the feel of this.

adityaathalye:
I tried it out, nice fix. Thank you!

PR #717, opened 2020-04-19
docs: enumerate special forms

hellerve:
This PR enumerates all special forms in the documentation, since they are kind of “magical”.

Cheers


hellerve:
This PR also adds documentation for #718 and #719.

PR #724, opened 2020-04-22
Add more functional dynamic functions

scolsen:
I took some time today to restrict the functionality of `map` to mapping only, since we have `zip`.

While I was at it, I added a few other utility functions too. I'll add tests tomorrow if I get a chance, but I'm opening this now for feedback!


eriksvedang:
Neat!

PR #725, opened 2020-04-22
Add bit-width command

hellerve:
This PR adds the command `bit-width`, which  gives you the integer/pointer width of the platform (usually 32/64 bits on modern systems).

This could be a dynamic global variable, but since we currently do not have a good mechanism for adding variables like that I added this as a command. I’d also be open to change this to a variable and build some plumbing for that.

Cheers


eriksvedang:
Cool!

PR #726, opened 2020-04-22
Long type to ensure longs are actually 64 bits

hellerve:
This MR is a rebased verison of #577; it should be functionally equivalent. Any problems with it are probably introduced by me.

Cheers


jacereda:
@hellerve thanks, I'm quite busy these days.

hellerve:
@jacereda no problem at all; I didn’t want to offend you or anything.

hellerve:
It seems like the Windows problem persists, and the other build breaks. That’s weird, because it works for me locally. I’ll dig into it.

hellerve:
I’ve investigated and fixed as many of the problems as I could find. Windows CI still quits after the 30th test (no matter which one it is) and I’m not quite sure why. Apparently it works on Windows desktops.

eriksvedang:
This does not compile on my Windows machine (Windows 10, clang-cl.exe).
Here's the full error: https://gist.github.com/eriksvedang/3be5a457cdb8c3224b0519b738d3e7c0

hellerve:
Interesting. It seems like safe division generates invalid code on some Windows machines. I’d just stub that out for Windows for the moment, then.

eriksvedang:
Yup, sounds good.

eriksvedang:
The test for safe arithmetic fails now, otherwise we're good.

hellerve:
The last commit should disable that test on Windows.

eriksvedang:
The output from the script is somewhat noisy (due to all the complaints about "format specifies type 'ssize_t' (aka 'int') but the argument has type 'ssize_t' (aka 'long')" in CHK_INDEX) but now all tests run!

There is actually an additional issue on my machine with the "no core" example, because my local config uses a function (defdynamic-once) that is undefined when using the --no-core flag, but this is a separate thing that we can solve at some future point. It won't happen on the CI at least. And I'm surprised it hasn't happened before, should be an issue on Mac and Linux aswell.

eriksvedang:
I've pushed a fix for the format specifiers.

Issue #729, opened 2020-04-26
Incorrect parenthesis placement results in infinite loop


sdilts:
The following code produces an infinite loop in the compiler:
``` clojure
(defn test [)
   (prn "Infinite loop!"))
```
Notice that the error is that the argument array is closed with a parenthesis and not a bracket.

hellerve:
Hey, good catch!

Actually, it seems like this does not produce an infinite loop in files. What’s happening is that the bracket/parens matcher that determines whether you’re doing multiline or single-line input is getting tricked into thinking the input will continue. It seems as if typing `]` doesn’t resolve it either.

hellerve:
Ah, because of the way the balance parser works, you’ll have to first close the bracket, then the paren. Typing `])` should work. This kind of behavior makes sense, but a guide might help here.

eriksvedang:
Will close since this isn't a bug as much as a usability problem (which is greatly eased by enabling the fix in #730, perhaps it should be on by default though?).

eriksvedang:
Oh, it is on by default! Nevermind then :)

PR #730, opened 2020-04-27
Proposal show unbalanced elements in the prompt

hellerve:
This PR is an effort to clearer communicate what needs to be closed in multiline input (which led to e.g. #729 being opened).

We already match unclosed brackets, parens, and quotes. All we need to do is display them. A simple way to do that would be something like what I added:

new prompt

What do y’all think? (Pinging especially @sdilts, since he got confused by this)

Cheers

### DIFF
--- a/src/Repl.hs
+++ b/src/Repl.hs
@@ -24,6 +24,8 @@ import Path
 import Lookup
 import Parsing (balance)
 
+import Debug.Trace
+
 instance MonadState s m => MonadState s (InputT m) where
     get = lift get
     put = lift . put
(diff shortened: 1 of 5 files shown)


src/Repl.hs
```diff
@@ -124,23 +126,24 @@ treatSpecialInput (':':rest) =
...
   do context <- get
-     let prompt = strWithColor Yellow (if null readSoFar then projectPrompt (contextProj context) else "     ")
-     input <- getInputLine prompt
+     input <- getInputLine (strWithColor Yellow prompt)
      case input of
         Nothing -> do
           liftIO exitSuccess
           return ()
         Just i -> do
           let concat = readSoFar ++ i ++ "\n"
```
scolsen, line 138: Not related to this PR, but maybe we should rename this binding? 

`concat` is a function in the Prelude, so this can throw people off if they're reading the code quickly.

scolsen:
What a neat idea! I love it! Code looks good.

I do wonder if we should stick this behind some kind of switch, though. While I'm a huge fan of these hints, I'm sure there's a contingent of people that would find them noisy.

hellerve:
Fair enough. I was thinking about making the second prompt customizable, but this pretty much has to be part of the compiler I think. But it could be part of some kind of project flag, for sure!

Naming preferences?

scolsen:
hmm, `paren-suggest`? `paren-hints`? `balance-hints`? `balance-suggestions`? something like that?

hellerve:
I think `hints` captures it better than `suggestions`, and while `balance` is less concrete than `paren`, it’s also more accurate (as we’re also tracking quotes and brackets)!

So my vote, out of those 4, goes for `balance-hints`.

bjorn3:
`paren-balance-hints`?

eriksvedang:
Looks great! But yes, has to be possible to turn off both in settings and as a command line flag since it will trip up external tools (read Emacs).

sdilts:
Good idea! Although when I saw the pull request title, I was expecting something more like the way Emacs handles opening and closing brackets:
```
Whenever you type a self-inserting character that is a closing delimiter, Emacs briefly indicates the location of the matching opening delimiter, provided that is on the screen. 
```
Source: https://www.gnu.org/software/emacs/manual/html_node/emacs/Matching.html

After trying it out, I'm happy with this solution. Carp already has a pretty luxurious REPL, and this makes it even better!

src/Repl.hs
hellerve: This was renamed to `concatenated`!

hellerve:
I finally got around to adding the switch for this! I ended up going with @bjorn3’s suggestion of `paren-balance-hints`.

eriksvedang:
Very cool feature @hellerve !!!

PR #732, opened 2020-04-29
Extend inner environments with lhs case bindings

### LINKED ISSUE #731: Regression after recent commits.
After pulling from the latest master, I can't use lambda's as member types:

```
out/main.c:10088:36: error: initializing 'void *' with an expression of incompatible type 'Lambda'
        Lambda _13 = { .callback = f, .env = NULL, .delete = NULL, .copy = NULL }; //Sym f Symbol
```

I've confirmed that there's no issue as of commit de8104eb--the problem arose someplace after that commit.

scolsen:
Previously, we'd stopped doing this, but it leads to issues with matches
on functions, e.g.

```
(deftype A (F [(Fn [] Int)]))

(defn m [a]
  (match a
    (A.F f) (f)))

(defn main []
  (println* (m (A.F (fn [] 123)))))
```

Will result in a runtime error if the environment isn't extended.

Fixes #731


scolsen:
@eriksvedang ready for a review whenever!

eriksvedang:
Lovely, thanks for fixing the bug and adding some much needed tests!

PR #733, opened 2020-04-29
Add TCO

hellerve:
This PR will add tail-call optimization/tail-call elimination (or TCO for short) to Carp. It’s currently in a prototyping phase, and I added this PR as a draft so everyone who wants to can follow along. This code is alpha quality, and as of now, was borne from a one-beer evening session. Don’t put too much hope in it just yet.

Conceptually, what we are doing to eliminate tail calls is adding a label `tco` to the beginning of every function and using `goto tco` to jump back to it when we want to make a recursive call. Before we can do that, though, we will have to reset the arguments to the new values (easy, we would pass them to the call anyway), and call all the deleters for things whose lifetime expires (easy, we have the instruments to do that).

There are complications, though, and they are mostly due to the fact that anything is an expression in Carp, and a lot of the emitter code will have to be changed to accomodate the special case of stopping an expression for a tail call. Consider `if`, for instance:

```clojure
(defn step [n]
  (if (= 0 n)
     0
     (step (- n 1)))
```

This is well and good, except that the `if` will try to emit something like this (I simplified it for readability):

```c
int s(int n) {
  int result;
  if (n == 0) {
     result = 0;
  } else {
      result = step(n - 1);
  }
  return result;
}
```

If we just replaced the tail call with a `goto`, this would become:

```c
int s(int n) {
tco: ;
  int result;
  if (n == 0) {
     result = 0;
  } else {
      n = n - 1;
      result = goto tco;
  }
  return result;
}
```

Which obviously does not work, because a `goto` is not an expression. (N.B. in the actual emitter things get a little more hairy, but this more or less describes the issue)

Now what we need to figure out is how to tell the emitter to handle a branch that will, by definition, never return and just go back to the beginning of the function. This is not a small undertaking, but should be fairly mechanical.

TODO:
- [x] Emit label and `goto`
- [ ] Ensure we check for tail position correctly
- [x] Update arguments before jumping
- [ ] Run deleters before jumping
- [ ] Update emitter to handle non-expression cases
- [ ] Verify

Cheers

### DIFF
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -91,6 +91,15 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
                            Functions -> 0
                            Globals -> 4
                            All -> 0
+        currentArgs =
+          case obj root of
+            Lst [XObj (Defn _) _ _, _, XObj (Arr args) _ _, _] -> map getSymName args
+            _ -> []
+        currentPath = getFuncName root
+        getFuncName (XObj (Lst [XObj (Defn _) _ _, XObj (Sym path _) _ _, _, _]) _ _) = path
+        getFuncName _ = SymPath [] ""
+        getSymName (XObj (Sym path _) _ _) = path
+        getSymName _ = SymPath [] ""
         visit :: Int -> XObj -> State EmitterState String
         visit indent xobj =
           case obj xobj of
@@ -194,6 +203,7 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
                      appendToSrc (defnDecl ++ " {\n")
                      when (name == "main") $
                        appendToSrc (addIndent innerIndent ++ "carp_init_globals(argc, argv);\n")
+                     appendToSrc "tco: ;\n"
                      ret <- visit innerIndent body
                      delete innerIndent i
                      when (retTy /= UnitTy) $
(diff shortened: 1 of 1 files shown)


eriksvedang:
Nice writeup! In the example above, it should be fine to just not assign to `result`, right? So when in tail call position, use `goto` instead of assigning...?

hellerve:
Yes, though the problem currently is that the assignment is emitted by the parent node, `if`. So the parent node needs to have a way to realize that its child nodes interrupted the expression flow.

jacereda:
Would it be easy to generate this instead?
```
int s(int n) {
tco:{
  int result;
  if (n == 0) {
     result = 0;
  } else {
      n = n - 1;
      result = 0;
      goto tco;
  }
  return result;}
}
```

hellerve:
Something like this is indeed what I’m currently doing, but the problem here is that this needs to work for all types, not just `int`s. This means that we need to know how to assign a dummy value for all cases, and I don’t think that’s a good or better solution than to reformulate the emitter.

But maybe there is another way?

eriksvedang:
I agree that a dummy value sounds like a bad idea. 

If the form knows that it's in tail-call position, shouldn't the calling form be able to know that too? That would make it possible to avoid emitting the assignment completely.

hellerve:
The problem is that this tail call might be arbitrarily deep in the stack. Consider a `do` form inside the `if`, for instance.

eriksvedang:
Right, good point...

jacereda:
> But maybe there is another way?

result=result?

hellerve:
That’s a fun one, and that might actually work. I’ll try it out! Thanks for the suggestion!

eriksvedang:
Would love to get this in! Perhaps something to pair on? Will close for now to keep things tidy.

PR #734, opened 2020-04-30
Change maybe and result to match on refs if possible

hellerve:
This PR is for the new reference matching capability. They change the matching functions that work on refs and get rid of the copy.

Cheers


eriksvedang:
Very nice!

PR #736, opened 2020-05-01
Make Dynamic.or and Dynamic.and primitives

### LINKED ISSUE #718: `Dynamic.or` and `Dynamic.and` should be primitives
Historically, `Dynamic.or` and `Dynamic.and` have been special forms, since they do not evaluate their arguments. There is no reason for this anymore, since primitives behave in exactly that way since #695.

Making these forms primitives would allow us to use them in higher-order functions.

It would require a bit of work inside `makePrim` to make this work, since it does not expect a module path now. This should be a fairly minor change, though.

I’d be happy to take this on.

Cheers

hellerve:
This PR fixes #718 by making `Dynamic.or` and `Dynamic.and` primitives.

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -72,37 +72,6 @@ eval ctx xobj@(XObj o i t) =
     resolveDef x = x
     eval' form =
       case form of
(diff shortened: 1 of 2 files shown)


eriksvedang:
Yay!

PR #740, opened 2020-05-03
Adds functions to operate on StaticArray

TimDeve:
This PR tries to get StaticArray to parity with Array function wise where possible. Functions that creates a new array were skipped. Also adds test where they were missing in `test/array.carp`.

Adds these functions:
- `Array.from-static`
- `StaticArray.reverse!`

Copies the following functions from Array to ArrayStatic:
- `empty?`
- `any?`
- `all?`
- `find`
- `find-index`
- `first`
- `unsafe-first`
- `last`
- `unsafe-last`
- `minimum`
- `maximum`
- `sum`
- `index-of`
- `element-count`
- `predicate-count`
- `aupdate!`
- `swap!`
- `nth`
- `contains?`

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -106,7 +106,7 @@ Returns `Nothing` if the array is empty.")
               (break))))
         eq)))
 
-    (doc maximum "gets the maximum in an array (elements must support `<`) and wraps it in a `Just`.
+  (doc maximum "gets the maximum in an array (elements must support `<`) and wraps it in a `Just`.
 
 If the array is empty, it returns `Nothing`.")
   (defn maximum [xs]
@@ -127,7 +127,7 @@ If the array is empty, returns `Nothing`")
     (if (empty? xs)
       (Maybe.Nothing)
       (let-do [result (unsafe-nth xs 0)
-             n (length xs)]
+               n (length xs)]
         (for [i 1 n]
           (let [x (unsafe-nth xs i)]
             (when (> result x)
@@ -323,9 +323,9 @@ The trailing elements of the longer array will be discarded.")
 
 If the `index` is out of bounds, return `Maybe.Nothing`")
   (defn nth [xs index]
-  (if (and (>= index 0) (< index (length xs)))
-    (Maybe.Just @(unsafe-nth xs index)) ; the copy will go away with lifetimes
-    (Maybe.Nothing)))
+    (if (and (>= index 0) (< index (length xs)))
+       (Maybe.Just @(unsafe-nth xs index)) ; the copy will go away with lifetimes
+       (Maybe.Nothing)))
 
   (doc remove "removes all occurrences of the element `el` in the array `arr`, in place.")
   (defn remove [el arr]
@@ -341,7 +341,7 @@ If the `index` is out of bounds, return `Maybe.Nothing`")
         (aset! &arr j @(unsafe-nth &arr (inc j))))
       (pop-back arr)))
 
-   (doc copy-filter "filters the elements in an array.
+  (doc copy-filter "filters the elements in an array.
 
 It will create a copy. If you want to avoid that, consider using [`endo-filter`](#endo-filter) instead.")
   (defn copy-filter [f a] (endo-filter f @a))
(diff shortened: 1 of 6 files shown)


core/Array.carp
```diff
@@ -323,9 +323,9 @@ The trailing elements of the longer array will be discarded.")
 
 If the `index` is out of bounds, return `Maybe.Nothing`")
   (defn nth [xs index]
-  (if (and (>= index 0) (< index (length xs)))
-    (Maybe.Just @(unsafe-nth xs index)) ; the copy will go away with lifetimes
-    (Maybe.Nothing)))
+   (if (and (>= index 0) (< index (length xs)))
```
hellerve, line 326: While we’re at it, I think we could even indent by one more space here.

core/Array.carp
```diff
@@ -390,4 +390,11 @@ It will create a copy. If you want to avoid that, consider using [`endo-filter`]
             (set! a (push-back a (Array.slice arr x y)))
             (set! x y)))
         a))
-)
+
+  (doc from-static "Turns a StaticArray into an Array. Copies elements.")
```
hellerve, line 394: If you want you can use Markdown syntax here for the types, like this:

```
(doc from-static "Turns a `StaticArray` into an `Array`. Copies elements.")
```

core/StaticArray.carp
```diff
@@ -1,5 +1,23 @@
 (defmodule StaticArray
 
+  (doc map! "Maps a function over the static array `xs`, mutating it in place. The difference to Array.endo-map (which does the same thing internally) is that this function takes a ref (since you can never have static arrays as values) and that it returns ().")
+  (defn map! [xs f]
```
hellerve, line 4: For convenience, could we return `xs` here anyway?

eriksvedang: I don't think that's very convenient since you'll have to `(ignore ...)` that often, no?

TimDeve: Don't believe so as every static array literals are (Ref StaticArray)

hellerve: Hm, okay! I just thought that a `map!` that doesn’t return anything looks a little weird, but I guess that makes sense.

eriksvedang: Since all other `!` function returns `()` I think this one should too.

hellerve: Sure, this can be marked as resolved from my end :)

TimDeve:
Made the requested edits, sorry about the random indentation changes, that would be Parinfer acting up.

core/StaticArray.carp
```diff
@@ -1,5 +1,23 @@
...
+             j (Int.dec (length a))]
+      (while (Int.< i j)
+        (let-do [tmp @(unsafe-nth a i)]
+          (aset! a i @(unsafe-nth a j))
+          (set! i (Int.inc i))
+          (aset! a j tmp)
+          (set! j (Int.dec j))))))
+
+  ;; NOTE: Following function declarations are copy of the ones in Array
+  ;; could we share inmplementation between Array and StaticArray?
```
eriksvedang, line 20: typo "inmplementation"

eriksvedang: I guess that whole second line can be removed since we've kinda resolved that in chat, for now?

TimDeve: Ok, will do.

eriksvedang:
The address sanitizer finds an error with the first test case (for `map!`), it's a free after use error since the ref `arr` is returned and the static array lives inside the `let`. This is actually detected when I paste the code into a separate function (only on the most recent commit on `master` though, this was not detected properly until today). 

For some reason it's still not detected in the test case though, so I need to fix that bug with the reference checker.

Please go through the test cases an look if this error pops up in other places. You might have to rearrange the code a bit for it to work, I'm afraid. Also, do you have the Clang address sanitizer? It will help you find these errors locally.

TimDeve:
For some reason I thought that it was fair game to return from a let but I guess it's a block {} in c so the automatic memory gets freed.

eriksvedang:
Yeah, sorry about the compiler not catching this (I'm on it!).

TimDeve:
Fixed the sanitiser errors

eriksvedang:
Hoorray! Thanks for your hard work, this will be super useful. Looking forward to seeing what you can do with it on the raspi!

PR #742, opened 2020-05-03
Make primitives proper AST objects

### LINKED ISSUE #719: Primitives should be usable in higher-order functions
Currently, primitives cannot be used in higher-order functions, because they cannot be passed around by value. We should fix that to make their usability better and the design a little cleaner.

Cheers

hellerve:
This PR fixes #719 by making primitives less magical. Because of current limitations of name resolution, this also means that dynamic `or` and `and` have to be renamed, which is a bummer. I settled on `dynor` and `dynand`, but I’m not really happy with those, so if you have suggestions, let me know!

Cheers

### DIFF
--- a/core/Core.carp
+++ b/core/Core.carp
@@ -47,7 +47,7 @@
 (load "Binary.carp")
 (load "Control.carp")
 
-(if (not (dynor (= "windows" (os)) (= "mingw32" (os))))
+(if (not (dynamic-or (= "windows" (os)) (= "mingw32" (os))))
   (do
     (system-include "sys/wait.h")
     (system-include "unistd.h"))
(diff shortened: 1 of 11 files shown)


eriksvedang:
Looks good and I can merge this as-is. But since you asked for it, here are some other possible names:
* `dynamic-or`, very long but definitely clearer
* `dyn-or`, at least a tiny bit clearer
* `OR`, kinda silly, but definitely short

scolsen:
I like `dynamic-or` even though it's long, I've found a lack of clarity is a usually bigger cost than a couple of keystrokes.

Someone who's new and can't intuitively puzzle out what `dyn` stands for will possibly never figure it out and give up (they could read the docs, but there's issue with capturing meta for dynamics right now, right? ). Someone who sees `dynamic-or` is less likely to give up just because the name is a bit verbose--at least there's really no ambiguity.

Plus, users can alias it locally to a shorter name, right?

eriksvedang:
I like that reasoning!

Issue #741, opened 2020-05-03
Allow aliases for registered types


hellerve:
We are currently able to alias C functions to make them more manageable in Carp:

```clojure
; given this C function
(inline-c "
int my_fun_function() { return 1; }
")
; we can do
(register my-fun-function (Fn [] Int) "my_fun_function")
```

This ensures that we keep consistent naming convention across language boundaries.

This is currently not possible with types, and it’s a bit of an annoyance. Consider a type like `godot_string`, mentioned int #737:

```clojure
; i can register it
(register-type godot_string)
; but i cant use it
(deftype MyWrapper [
  inner godot_string
])
; this will be read as a type variable and blow up!
```

We could relax our naming conventions, but I don’t think this is a good solution. Instead, in the spirit of what we can do for types, we should be able to do this:

```clojure
(register-type GodotString "godot_string") ; with optional members before

(deftype MyWrapper [
  inner GodotString
])
```

Since we already have a mechanism for this, this shouldn’t be _too big_ of a change. I’d be open to contributing it!

cc: @linkpy because he reminded me that this is an issue, and because it also gives some context to the problem he’s experiencing.

Cheers

eriksvedang:
Sounds like a great first step towards solving the problems described in #737 !

linkpy:
I started working on implementing C-type aliases : https://github.com/linkpy/Carp/tree/c-type-aliases

But to be honest I'm not entirely sure what I'm doing (I dont know much about Haskell). What I did is basically handle `ExternalType` the same way as `External` is handled : `ExternalType (Maybe String)`. For now the project compiles, and I think when emitting C code, a `ExternalType (Just String)` uses `LookupGlobalOverride` to emit the right type name.

But I encountered some issues (which I think is the reader for the `register-type` syntax ?) : https://github.com/linkpy/Carp/blob/e73534dede586cfd68ad57d9313181807450364e/src/Primitives.hs#L208
This line generates the following error and I'm not sure why exactly. Futhermore I'm not even sure it's the right place or the right way to do it.

~~~~~~
/home/linkie-pi/Documents/Carp/src/Primitives.hs:208:1: warning: [-Woverlapping-patterns]
    Pattern match is redundant
    In an equation for ‘primitiveRegisterType’:
        primitiveRegisterType _ ctx [XObj (Sym (SymPath [] t) _) _ _, ty, XObj (Str override) _ _] = ...
    |
[...]

PR #743, opened 2020-05-04
Registered type aliasing

linkpy:
Implements #741.

As I'm very new to Haskell, I assume most of the things could be done in a better way but I really want to contribute.

To do:

- [x] Make the `ExternalType` object to have an optional string (like `External`) ;
- [x] ~~Make a `ExternalType (Just String)` emit the override instead of the type name (I think it's done, but I'm not completely sure about this, so if someone could check that as I dont really know how the emitting process works, I based my code on how it is done for `External`)~~ Make `ExternalType (Just String)` to emit a `typedef` declaration ;
- [x] Adds the reader handler for `(register-type Type "Alias")`
- [x] Adds the reader handler for `(register-type Type "Alias" [fields])`

### DIFF
--- a/src/Obj.hs
+++ b/src/Obj.hs
@@ -358,7 +358,7 @@ prettyUpTo max xobj =
             Instantiate _ -> ""
             External Nothing -> ""
             External (Just override) -> ")"
-            ExternalType Nothing -> ")"
+            ExternalType Nothing -> ""
             ExternalType (Just override) -> ")"
             DocStub -> ""
             Defalias _ -> ""
--- a/src/Obj.hs
+++ b/src/Obj.hs
@@ -71,7 +71,7 @@ data Obj = Sym SymPath SymbolMode
          | DefSumtype Ty
          | With
          | External (Maybe String)
-         | ExternalType
+         | ExternalType (Maybe String)
          | DocStub
          | Deftemplate TemplateCreator
          | Instantiate Template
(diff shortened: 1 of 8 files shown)


eriksvedang:
Looks good so far!

src/Obj.hs
```diff
@@ -358,7 +358,8 @@ prettyUpTo max xobj =
             Instantiate _ -> ""
             External Nothing -> ""
             External (Just override) -> ")"
-            ExternalType -> ""
+            ExternalType Nothing -> ")"
```
eriksvedang, line 361: You can just match on `_` here.

eriksvedang: (to condense the two cases into one)

linkpy: The two cases doesn't return the same value. Is it fine if both returns `")"` ?

linkpy: Oh actually I made a mistake. I wanted to make that part just like `External` so  `ExternalType Nothing` should return "".

src/Primitives.hs
```diff
@@ -183,10 +183,33 @@ primitiveRegisterType _ ctx [XObj (Sym (SymPath [] t) _) _ _] = do
   let pathStrings = contextPath ctx
       typeEnv = contextTypeEnv ctx
       path = SymPath pathStrings t
-      typeDefinition = XObj (Lst [XObj ExternalType Nothing Nothing, XObj (Sym path Symbol) Nothing Nothing]) Nothing (Just TypeTy)
+      typeDefinition = XObj (Lst [XObj (ExternalType Nothing) Nothing Nothing, XObj (Sym path Symbol) Nothing Nothing]) Nothing (Just TypeTy)
   return (ctx { contextTypeEnv = TypeEnv (extendEnv (getTypeEnv typeEnv) t typeDefinition) }, dynamicNil)
 primitiveRegisterType _ ctx [x] =
   return (evalError ctx ("`register-type` takes a symbol, but it got " ++ pretty x) (info x))
+primitiveRegisterType _ ctx [XObj (Sym (SymPath [] t) _) _ _, XObj (Str override) _ _] = do
```
eriksvedang, line 190: Do you think you could remove some of the code duplication here by extracting out a common helper function?

linkpy: Done. I created two new functions, one when there is members and one when there is none.

src/Obj.hs
eriksvedang: Lovely!

src/Primitives.hs
eriksvedang: Lovely!

eriksvedang:
Thanks a lot for this!

PR #746, opened 2020-05-04
Polymorphic Constructors

scolsen:
Adds support for polymorphic type constructors, e.g. 

`(definterface fmap (Fn [(Fn [a] b) (f a)] (f b)))`

Can be properly resolved to concrete constructors and type inhabitants, e.g.:

`(sig Array.fmap (Fn [(Fn [a] b) (Array a)] (Array b)))`

Currently it all compiles, but there's already a runtime error I have to track down.


scolsen:
> Should we also add compiler tests with type variables in the struct name to `test/Spec.hs`?

Sounds good to me! 

I think there’s still an outstanding design consideration too, right now, it’s possible to do the following:

```
(definterface fmap (Fn [(Fn [a] b) (f a)] (f b)))
(defmodule Int (defn fmap [f x] (f x)))
```

Then Int.fmap gets the type `(Fn [a b] b)` and is a valid implementation of fmap. 

In other words the constructor gets “erased”. We can either decide this is ok, or we can implement some extra checks for Fn types to make sure that polymorphic constructors are matched with concrete constructors and not non-constructor types

eriksvedang:
I think we should implement proper kinds checking to help the programmer from confusing themselves too much. I get very confused just looking at that erasure :)

Issue #735, opened 2020-04-30
Make (reload) not reload the core libraries


eriksvedang:
To make the repl snappier, we could have a mechanism for marking loaded files as "frozen" so they are not reloaded when calling `(reload)` / `:r` since that shouldn't be necessary (unless you're working on the core libraries... which I guess most Carp users are so this should be possible to turn off also).

eriksvedang:
Done!

PR #748, opened 2020-05-05
Do not reload core libs

eriksvedang:
I implemented a solution to #735 which made the repl reloading way faster, so I think it's a win. 

Since it's hard to test in a realistic way, if anyone have the time to run it on their own projects and report back, that would be much appreciated!


eriksvedang:
This will probably be very annoying for Core library developers, so I will add a way to turn it off...

hellerve:
This looks great, though, and `load-once` is potentially useful for other people as well :) Great work!

eriksvedang:
You can now configure this with `(Project.config "force-reload" true)`.

PR #750, opened 2020-05-05
Add kinds checking

### LINKED ISSUE #561: Kind-checking of type constructors is missing
Right now it's possible to use type constructors like `Vector3` without applying them to a type, for example:

```
(deftype Player
    [pos Vector3
     hp Int])
```

This leads to warnings about missing deleters, etc. but that misses the main point which is that the `Vector3` type must take an argument:

```
(deftype Player
    [pos (Vector3 Float)
     hp Int])
```

scolsen:
This commit adds basic kinds checking on interfaces, such that an
interface that takes a type constructor (f a) can't be implemented by a
function using a concrete type like `Int` in that position.

It also adds checks on deftypes to ensure uninhabited type constructors can't be set as members:

```
(deftype (Foo a) [pos Maybe])
Invalid type definition for 'Foo':

Can't use an uninhabited type constructor (Maybe a) as a member type at line 1, column 23 in 'REPL'. The type constructor (Maybe a) expects 1 arguments but got 0

Traceback:
(deftype (Foo a) [pos Maybe]) at REPL:1:1.
```

Fixes #561

### DIFF
--- a/core/Int.carp
+++ b/core/Int.carp
@@ -62,11 +62,11 @@
 
   (doc to-int "acts as the identity function to implement the interface.")
   (sig to-int (Fn [Int] Int))
-  (defn to-int [a] a)
+  (defn to-int [a] (the Int a))
 
   (doc from-int "acts as the identity function to implement the interface.")
   (sig from-int (Fn [Int] Int))
-  (defn from-int [a] a)
+  (defn from-int [a] (the Int a))
 )
 
 (defmodule IntRef
--- a/core/Int.carp
+++ b/core/Int.carp
@@ -62,11 +62,11 @@
 
   (doc to-int "acts as the identity function to implement the interface.")
   (sig to-int (Fn [Int] Int))
-  (defn to-int [a] (the Int a))
+  (defn to-int [a] a)
 
   (doc from-int "acts as the identity function to implement the interface.")
   (sig from-int (Fn [Int] Int))
-  (defn from-int [a] (the Int a))
+  (defn from-int [a] a)
 )
 
 (defmodule IntRef
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -122,7 +122,7 @@ registerInInterfaceIfNeeded ctx path@(SymPath _ name) definitionSignature =
                      in  return $ ctx { contextTypeEnv = TypeEnv (extendEnv typeEnv name updatedInterface) }
                 else Left ("[INTERFACE ERROR] " ++ show path ++ " : " ++ show definitionSignature ++
                            " doesn't match the interface signature " ++ show interfaceSignature)
-           else Left ("[INTERFACE ERROR] " ++ show path ++ ":" ++ show definitionSignature ++ " kind mismatch")
+           else Left ("[INTERFACE ERROR] " ++ show path ++ ":" ++ " One or more types in the interface implementation " ++ show definitionSignature ++ " have kinds that do not match the kinds of the types in the interface signature " ++ show interfaceSignature ++ "\n" ++ "Types of the form (f a) must be matched by constructor types such as (Maybe a)")
        Just (_, Binder _ x) ->
          error ("A non-interface named '" ++ name ++ "' was found in the type environment: " ++ show x)
        Nothing -> return ctx
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -116,11 +116,13 @@ registerInInterfaceIfNeeded ctx path@(SymPath _ name) definitionSignature =
   let typeEnv = getTypeEnv (contextTypeEnv ctx)
   in case lookupInEnv (SymPath [] name) typeEnv of
        Just (_, Binder _ (XObj (Lst [XObj (Interface interfaceSignature paths) ii it, isym]) i t)) ->
-         if areUnifiable interfaceSignature definitionSignature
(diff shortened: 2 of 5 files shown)


scolsen:
Oops, I opened this for review when I meant for it to be a draft! It's not quite ready for review yet

scolsen:
@eriksvedang @hellerve this should be ready for review now!

core/Int.carp
```diff
@@ -62,11 +62,11 @@
 
   (doc to-int "acts as the identity function to implement the interface.")
   (sig to-int (Fn [Int] Int))
-  (defn to-int [a] a)
+  (defn to-int [a] (the Int a))
```
scolsen, line 65: Shouldn't be necessary anymore--I snuck this in when I initially was trying out the rule that type variables can't replace variables (this rule is too strict).

src/Primitives.hs
```diff
@@ -116,11 +116,13 @@ registerInInterfaceIfNeeded ctx path@(SymPath _ name) definitionSignature =
...
-              in  return $ ctx { contextTypeEnv = TypeEnv (extendEnv typeEnv name updatedInterface) }
-         else Left ("[INTERFACE ERROR] " ++ show path ++ " : " ++ show definitionSignature ++
-                    " doesn't match the interface signature " ++ show interfaceSignature)
+         if checkKinds interfaceSignature definitionSignature
+           then if areUnifiable interfaceSignature definitionSignature
+                then let updatedInterface = XObj (Lst [XObj (Interface interfaceSignature (addIfNotPresent path paths)) ii it, isym]) i t
+                     in  return $ ctx { contextTypeEnv = TypeEnv (extendEnv typeEnv name updatedInterface) }
+                else Left ("[INTERFACE ERROR] " ++ show path ++ " : " ++ show definitionSignature ++
+                           " doesn't match the interface signature " ++ show interfaceSignature)
+           else Left ("[INTERFACE ERROR] " ++ show path ++ ":" ++ show definitionSignature ++ " kind mismatch")
```
scolsen, line 125: Better message?

src/Types.hs
```diff
@@ -285,6 +301,27 @@ areUnifiable FuncTy{} _ = False
...
+-- Checks whether or not the kindedness of types match
+-- Kinds are polymorphic constructors such as (f a)
+-- Note that this disagrees with the notion of unifiablitity in areUnifiable
+checkKinds :: Ty -> Ty -> Bool
+-- Base < Higher
+checkKinds (FuncTy argTysA retTyA _) (FuncTy argTysB retTyB _) =
+  let argKinds = zipWith (<=) (map tyToKind argTysA) (map tyToKind argTysB)
+      retKinds = tyToKind retTyA <= tyToKind retTyB
+  in all (== True) (retKinds : argKinds)
+--checkKinds (StructTy _ argTysA) (StructTy _ argTysB) =
```
scolsen, line 313: All the commented code can be deleted--what we have is sufficient for checking interfaces, I'm not sure that any other checks are needed.

src/Validate.hs
```diff
@@ -52,13 +52,35 @@ canBeUsedAsMemberType typeEnv typeVariables t xobj =
...
                           canBeUsedAsMemberType typeEnv typeVariables t xobj
-    StructTy name tyvars -> return ()
+    s@(StructTy name tyvar) ->
+      if isExternalType typeEnv s
+      then return ()
+      else case name of
+             (ConcreteNameTy n) ->
+               case lookupInEnv (SymPath [] n) (getTypeEnv typeEnv) of
+                 Just (_, binder@(Binder _ xo@(XObj (Lst (XObj (Deftype t') _ _ : _))_ _))) ->
+                     checkInhabitants t'
```
scolsen, line 71: fix spacing.

scolsen: Done 56d3b02

src/Types.hs
scolsen: Done 56d3b02

src/Primitives.hs
scolsen: Done fc045f6

core/Int.carp
scolsen: Done 
56d3b02

scolsen:
I've tried to make the error messages a bit less mysterious! @hellerve @eriksvedang

src/Types.hs
```diff
@@ -285,6 +301,17 @@ areUnifiable FuncTy{} _ = False
...
+-- Checks whether or not the kindedness of types match
+-- Kinds are polymorphic constructors such as (f a)
+-- Note that this disagrees with the notion of unifiablitity in areUnifiable
+checkKinds :: Ty -> Ty -> Bool
+-- Base < Higher
+checkKinds (FuncTy argTysA retTyA _) (FuncTy argTysB retTyB _) =
+  let argKinds = zipWith (<=) (map tyToKind argTysA) (map tyToKind argTysB)
+      retKinds = tyToKind retTyA <= tyToKind retTyB
+  in all (== True) (retKinds : argKinds)
+checkKinds t t' = tyToKind t == tyToKind t'
```
scolsen, line 313: Maybe this should match the logic above? Instead of equals: `t <= t`. As of right now though, this function is only called against `Fn` types.

scolsen: Actually, if we define it this way we can call the function recursively instead of mapping `tyToKind`

scolsen: I went ahead and made this change.

eriksvedang:
Bring out the champagne, this is so great!!!

PR #751, opened 2020-05-05
Inlined C command

linkpy:
Adds the `inline-c` command, which is a simple pass-through for custom C code.

~~~~lisp
(system-include "stdio.h")

(inline-c 
  "void my_c_function() {\n"
  "  printf(\"hello from C\\n\");\n"
  "}")

(register my-c-function (Fn [] ()) "my_c_function")

(defn main []
  (my-c-function))
~~~~

The `inline-c` commands accepts numbers and strings. It can be used with dynamic functions for complex custom C code generation.


t.carp
```diff
@@ -0,0 +1,6 @@
+(defmodule Foo
```
hellerve, line 1: Should this file be part of the PR?

app/Main.hs
```diff
@@ -86,7 +87,7 @@ main = do setLocaleEncoding utf8
                       else return context
           finalContext <- loadFiles context' argFilesToLoad
           case execMode of
-            Repl -> do putStrLn "Welcome to Carp 0.3.0"
+            Repl -> do putStrLn "Welcome to Carp 0.3.1WIP"
```
hellerve, line 90: I don’t think that the version string should be changed here. That should happen in master or on release.

'
```diff
@@ -0,0 +1,32 @@
+;; Generic tests on sumtypes.
```
hellerve, line 1: Are those tests pulled in by @scolsen’s commit?

scolsen: Yeah, that was a goof on my part that I removed. (I somehow copied and saved that test file as `'`).

I removed that in another commit, so pulling in/merging master to this branch should fix this.

t.carp
scolsen: Also some muck from my old commits, geez-- I stepped these back via another commit, but it looks like I should have squashed and amended, oof.

Pulling master into this branch should fix this.

linkpy: No, it must come from @scolsen commit that got into my branch. File removed.

app/Main.hs
linkpy: I think I have an issue with my git repo, or that i'm not using it write. This change was even before my first PR (a commit named WIP i squashed). I have no idea how it got there, maybe during a rebase. It wasn't intended. Version reverted.

'
linkpy: File removed.

eriksvedang:
Unfortunately this implementation has a big problem, since each call to `(inline-c <code>)` will add `<code>` to the string to be emitted. So this breaks reloading of files...

src/Commands.hs
```diff
@@ -850,3 +853,27 @@ saveDocs :: Context -> [(SymPath, Binder)] -> IO (Context, Either a XObj)
...
+    checkArg _ (XObj (Str _) _ _) = Nothing
+    checkArg ctx x = Just (evalError ctx ("Invalid argument to inline-c (expected Num, Str, Sym, InlinedC) : " ++ pretty x) (info x))
+
+    stringify :: XObj -> String
+    --stringify (XObj (Sym p _) _ _) = show p
+    stringify x@(XObj (Num _ _) _ _) = pretty x
+    stringify (XObj (Str s) _ _) = unescapeString s
+
+commandInlineC ctx _ =
+  return (evalError ctx "inline-c expects a Symbol then some Strings and/or Numbers." (Just dummyInfo))
```
hellerve, line 879: I would consider matching on the arguments here and changing the error message to:

```
"The first argument to `inline-c` must be a symbol, got `"++ <matchvar> ++ "`"
```

src/StringUtils.hs
```diff
@@ -0,0 +1,29 @@
+
```
hellerve, line 1: While I recognize that unescaping the string is important here, I’m not sure I like having another Parsec parser just for that. Maybe we can think of a more lightweight way of doing this?

eriksvedang:
I think this command is a good idea and your examples in the chat show that there's a lot of potential to it.

This implementation looks good for what it does, the code is clean, no complaints. But like I said (also in chat) this is a pretty severe code duplication, since the existing templating system can already do all this (and more). There's already a proliferation of `Obj` cases, so I'm very wary of adding more unless strictly necessary. I hope you agree.

I think the best way forward might be to implement access to the templates as a command, and then you can add the `inline-c` as a macro on top of that. It will have the same interface that you have designed here, but the internal additions will be smaller. Since the templates are one of the trickiest things in the compiler (and they're more or less undocumented) I can volunteer to implement the `deftemplate` command.

app/Main.hs
```diff
@@ -4,6 +4,8 @@ import qualified System.Environment as SystemEnvironment
 import System.Console.Haskeline (runInputT)
 import GHC.IO.Encoding
 
+import qualified Data.Map as Map
```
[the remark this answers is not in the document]
linkpy: Yes indeed, it was when the inlined C was added to a map.

src/StringUtils.hs
linkpy: I actually spent a while trying to figure out something but wasn't able to find anything unfortunately. The options I found used external packages not included in Carp, and I didn't want to add a package on my own. 

For example, [Text.Replace](http://hackage.haskell.org/package/text-replace-0.0.0.6/docs/Text-Replace.html) could do the work.

linkpy: Actually I dont think it is required since rawstrings are merged now.

eriksvedang:
I really like it now, seems super powerful!
As far as I can tell, the implementation is solid to.

PR #752, opened 2020-05-06
Solve for polymorphic constructor constraints

scolsen:
Looks like I forgot a tiny but important change for the polymorphic
constructor support--constraint solving!

I only found this out when refactoring an applicative <*> or 'sequence'
definition-- the definition has a polymorphic constructor with a
function argument: `(f (Fn [a] b))`.

When carp attempted to resolve this type against a type returned by lift
which has the type `[a] (f a)`, a generically lifted value, including
functions, it complained:

`can't match (r1 r2)` with type `(r3 (Fn [r4] r5))`. (numbers are
arbitrary here and just for illustration).

Sure enough, we weren't solving for such a constraint. The constraint
resolution now works as expected.


eriksvedang:
Excellent!

PR #754, opened 2020-05-06
Fix non-exhaustive patterns in concretize

### LINKED ISSUE #464: Non-exhaustive patterns in Concretize
When I attempt to use the following definitions, I get an error about non-exhaustive patterns in `concretize.hs`

```clojure
(def scheduled-propagators (the (Array (Fn [] ())) (zero)))                 
    ;; all propagators ever scheduled during the scheduler's lifetime.          
(def history (the (Array (Fn [] ())) (zero)))  

(defn run-alerted []                                                        
        (do                                                                     
          (clear-schedule)                                                      
          (foreach [propagator &scheduled-propagators]                          
            (~propagator))                                                      
          (if (propagators-scheduled?)                                          
              (run-alerted)                                                     
              (State.Done))))    
```

My guess is that concretize doesn't yet account for the `()` type? either that or there's something wrong in my carp code. 

Happy to look into this!

scolsen:
This change supports more flexible Sumtype declarations.

We used to allow a sumtype declaration such as:

`(deftype (State a) Done (Value [a]))`

Which the compiler accepts. However, concretization for this type
failed, since we didn't handle the solitary Sym case `Done` when
concretizing the type.

This commit fixes that issue, so that now the following works:

```
(deftype (State a) Done (Value [a]))
;; Omitting the type annotation here would result in the obvious
;; unresolved generic type error.
(defn emit-done [] (the (State Int) (State.Done)))
(emit-done)
;; => State.Done
```

fixes #464

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -449,6 +449,10 @@ instantiateGenericSumtype typeEnv originalStructTy@(StructTy _ originalTyVars) g
                                             : concat okDeps
                     Left err -> Left err
 
+-- Resolves dependencies for sumtype cases.
+-- NOTE: This function only accepts cases that are in "canonical form"
+-- (Just [x]) aka XObj (Lst (Sym...) (Arr members))
+-- On other cases it will return an error.
 depsForCase :: TypeEnv -> XObj -> Either TypeError [XObj]
 depsForCase typeEnv x@(XObj (Lst [_, XObj (Arr members) _ _]) _ _) =
   concat <$>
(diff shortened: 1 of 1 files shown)


src/Concretize.hs
```diff
@@ -449,19 +449,31 @@ instantiateGenericSumtype typeEnv originalStructTy@(StructTy _ originalTyVars) g
...
 replaceGenericTypeSymbolsOnCase mappings singleCase@(XObj (Lst (caseName : caseMembers)) i t) =
   XObj (Lst (caseName : map replacer caseMembers)) i t
   where replacer memberXObj =
           replaceGenericTypeSymbols mappings memberXObj
+-- Handle cases like `(State a) Done (Value [a]))`
+-- `Done` is a Sym, not a Lst. DepsForCase, like this function
+-- expects and only matches on a Lst, so we convert the problematic cases to a
+-- canonical form. (see `depsForCase` above
+replaceGenericTypeSymbolsOnCase mappings nakedCase@(XObj (Sym (SymPath _ _) _) i t) =
+  XObj (Lst [nakedCase, XObj (Arr []) i t]) i t
```
eriksvedang, line 475: I know there is some other pass that does this transformation too. It obviously doesn't hurt to do it here too, since this must be earler in the process. But I might add a comment about that in the code too. Will merge though.

scolsen: Essentially the same thing happens in `SumtypeCases.hs` from what I can tell--I think that's what permits the initial definition to come through--but somehow the translation doesn't make it over to `Concretize`

Maybe we should abstract it out into a function or type synonym? 

`canoncialSumtypeCase` or `validSumtypeCase` or `concreteSumtypeCase` or something? `XObj -> XObj`

eriksvedang:
Sounds like a great idea!

PR #761, opened 2020-05-08
Dynamic `while`

linkpy:
Make the `while` control structure to be evaluated in dynamic code.

```
> (defndynamic test [] (for [i 0 4] (macro-log "test")))
> (test)
test
test
test
test
```


hellerve:
The test here is waiting for the fix in #763, then it should hopefully work!

Issue #708, opened 2020-04-05
Unicode characters generate invalid c


scolsen:
When I build the following:

`(def foo \😀)`

It generates invalid c:

```
 out/main.c:2899:15: error: character too large for enclosing character literal type
        foo = '😀';
              ^
1 error generated.
carp: callCommand: clang -fPIC -lm -g out/main.c -shared -o "out/Untitled"  -I/Users/scottolsen/dev/Carp//core/   (exit 1): failed
```

...I fuzzily recall that this used to work correctly? Everything works right in the repl as well, e.g.

`(= \😀\😀)` evaluates to `true` without a problem.

I believe this is due to the fact that we translate carp `Chars` to C's `char` type which is not large enough for unicode values. This [document](https://www.cprogramming.com/tutorial/unicode.html) recommends using `uint32` to represent characters in order to provide full unicode support.

Alternatively, we can keep Carp's types closer to their direct C equivalents and solve this in Carp itself...this is probably/possibly the right way to go for portability reasons? We would need to define a new types/functions etc. that translate Char arguments to the appropriate UInt32 type.

jacereda:
I would stay away from uint32 and just use UTF8.

jacereda:
Oh, never mind... you mean for single characters. A single character should probably be an uint, yes.

jacereda:
It's probably a matter of prefixing characters with `U`. This builds for me with clang:
```
uint32_t foo = U'😀';
```

scolsen:
> It's probably a matter of prefixing characters with `U`. This builds for me with clang:
> 
> ```
> uint32_t foo = U'😀';
> ```

Wide characters `wchar_t` work similarly--only the prefix is `L'\u2665'`--shouldn't be too tricky to add a new type for the compiler. 

What does confuse me a little, however, is that the repl can evaluate the equality comparison `=` just fine; meanwhile several other functions break on this input (e.g. prn or str).

jacereda:
`wchar_t` can be 16 or 32 bits. Some compilers have flags to select the size.

jacereda:
I guess the right type is `wint_t`. That's what things like `iswalnum()` and friends use.

scolsen:
> `wchar_t` can be 16 or 32 bits. Some compilers have flags to select the size.

Right, sorry--wasn't suggesting it as an alternative, just mentioning that it works similarly. (though I actually do need to support wchar_t for ncurses bindings and having the L'' literals in carp would be nice)!

[...]

PR #768, opened 2020-05-09
Linux fixes

jacereda:
This fixes the problems I encountered when compiling under NixOS while preparing a fix for #708 .

### DIFF
--- a/benchmarks.sh
+++ b/benchmarks.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e; # will make the script stop if there are any errors
 
--- a/core/StdInt.carp
+++ b/core/StdInt.carp
@@ -1,4 +1,4 @@
-(relative-include "carp_stdint.h")
+(system-include "carp_stdint.h")
 
 (register-type Uint8)
 (register-type Uint16)
--- a/core/System.carp
+++ b/core/System.carp
@@ -15,7 +15,7 @@
   (doc sleep-seconds "Sleeps for a specified number of microseconds.")
   (register sleep-micros (Fn [Int] ()))
   (doc system "Performs a system command.")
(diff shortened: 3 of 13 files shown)


hellerve:
This looks good to me, although it is mildly flabbergasting that adding `U` to our characters, i.e. emitting `char32_t` didn’t lead to more errors. One thing that definitely breaks is `chars` and `from-chars`:

```clojure
  (assert-equal test
                &[\s \v \e \d \ä \n \g]
                &(chars "svedäng")
                "chars works as expected")
  (assert-equal test
                "svedäng"
                &(from-chars &[\s \v \e \d \ä \n \g])
                "from-chars works as expected")
```

Both of these test cases would currently fail.

(BTW: I had to look up what type the `U` character literal emits and decided to take a look at the C standard; it seems that C99 only had `L` character literals and `u` and `U` were added in C11.)

jacereda:
Yes, those tests would be handled in the next PR (unless those were already passing).
As for `U` being C11, AFAIK (correct me if I'm wrong) it's quite widespread even in pre-C11 compilers, but we can always replace those by the numeric value at some point if it's problematic.

hellerve:
I think it should be fine, but I’m personally also lucky enough to have a quite simple setup (modern compilers, almost no constraints).

I just made up those tests because I was interested in whether it would be somehow handled magically, expecting them to fail. When we make the switch to `U`, though, I guess they should be made to pass.

Was the `U` change a first stab at #708 or is that part of making it work on Linux?

jacereda:
The `U` change was just the simple side of #708, mainly making `str` work on wide characters.

hellerve:
Cool. I’m happy with keeping it in this PR, just wanted to clarify.

eriksvedang:
Is this PR surpassed now? @jacereda

jacereda:
Sure.

PR #769, opened 2020-05-09
Add an `implements` primitive, update core

### LINKED ISSUE #758: [Question] Explicitly Implemented or Scoped Interfaces?
Right now, any function that shares a name with an interface is considered a potential implementation of that interface. This can lead to some clashes. Consider:

```
(doc empty                                                                                                                                                                                  
  "The identity of choice.")                                                                                                                                                                
(definterface empty (Fn [] (f a)))
```

When we load this interface, the following error occurs:

```
carp: [INTERFACE ERROR] Bucket.empty : (λ [] (Bucket a b)) doesn't match the interface signature (λ [] (f a))
  error, called at src/Primitives.hs:413:17 in CarpHask-0.3.0.0-5ZhYmqzW9hmAQ94u8eHNWx:Primitives
```

Because `Bucket` happens to already define a function `empty` that has the same name as our interface.

In large programs, the possibility of this occurring increases. I'm wondering if we should consider adding a way to make interface implementations more explicit to avoid these kinds of problems? 

One possibility: instead of taking the declaration of the binding itself to be an interface based on name alone, we require an additional call:

```
(defn empty [] ...fn-body)
(implements empty empty)
;; this would also grant us naming flexibility
(defn foobar [] ...fn-body)
(implements empty foobar)
```

This is how most languages do it with keywords like `override` or e.g. Haskell's `instance where`

What does everyone think?

scolsen:
This change adds a new primitive Implements which changes interface
implementations from being implicit to being explicit. Going forward,
users will have to declare (implements <interface> <implementation>) to
explicitly add a function to an interface. This provides two benefits:

- Prevents unwitting name clashes. Previously, if one defined a function
  that happened to have the same name as an interface, it was
  automatically assumed the function implemented that interface when this
  is not always the intention, especially in large programs.
- Name flexibility. One can now implement an interface with a function
  that has a different name than the interface, which allows for greater
  flexibility.

I've updated core to make the necessary calls to the new primitive.

Since str and copy are derived automatically for types, we treat these
functions as a special case and auto-implement the interfaces.

Fixes #758

TODO:
- [x] Make it possible to call `implements` before the implementing function is defined.


eriksvedang:
Very cool, looks nice! I'll read it through properly tomorrow or Monday.

scolsen:
Cool! Looks like I'll have to add a few more implements declarations before the tests pass.

There are also two remaining things that I can't seem to get to work:

- retroactively adding definitions to interfaces if the definitions are defined first.
- We can support calling `implements` before providing a definition now, but one still has to qualify them with the full module path--it'd be nice to make this relative to the module instead.

scolsen:
It looks like using alternative names isn't reflected in the ouput right now either

I think it's because we use multiSym to lookup and print the interface, which only takes the single name -- we have to change this to look for all functions with `implements` metadata that equals the interface name instead.

src/StartingEnv.hs
```diff
@@ -360,6 +360,7 @@ dynamicModule = Env { envBindings = bindings
                     , makePrim "eval" 1 "evaluates a list." "(eval mycode)" primitiveEval
                     , makePrim "defined?" 1 "checks whether a symbol is defined." "(defined? mysymbol)" primitiveDefined
                     , makePrim "deftemplate" 4 "defines a new C template." "(deftemplate symbol Type declString defString)" primitiveDeftemplate
+                    , makePrim "implements" 2 "designates a function as an implementation of an interface" "(implements zero Maybe.zero)" primitiveImplements
```
hellerve, line 363: Nitpick: there should be a period at the end of the documentation sentence.

src/Primitives.hs
```diff
@@ -401,18 +437,26 @@ primitiveMetaSet _ ctx [XObj (Sym _ _) _ _, key, _] =
...
   argumentErr ctx "meta-set!" "a symbol" "first" target
 
-retroactivelyRegisterInterfaceFunctions :: Context -> String -> Ty -> IO Context
-retroactivelyRegisterInterfaceFunctions ctx name t = do
+retroactivelyRegisterInterfaceFunctions :: Context -> String -> SymPath -> IO Context
+retroactivelyRegisterInterfaceFunctions ctx name interface@(SymPath _ inter) = do
   let env = contextGlobalEnv ctx
-      found = multiLookupALL name env
-      binders = map snd found
+      bindings = map snd $ Map.toList (envBindings env)
```
hellerve, line 443: Won’t this only look in the current environment, and shouldn’t it also look in all sub-environments?

scolsen: Yes! Thank you! I was misunderstanding precisely how contexts were set up so this part was confusing me greatly. Thanks for clearing it up!

scolsen:
Welp, I can't seem to figure out why, but for some reason functions with different names aren't being accepted as implementations--yet when you call info on the function, it repots them as implementatations:

```
鲤 (definterface goo (Fn [] Int))
鲤 (defn bar [] 3)
鲤 (implements goo bar)
鲤 :i goo
goo : (λ [] Int) = {
    
}
Defined at line 1, column 15 in 'REPL'
鲤 :i bar
goo : (λ [] Int) = {
    bar
}
Defined at line 1, column 15 in 'REPL'
bar : (λ [] Int)
鲤 (defn baly [] (goo))
I can’t find any implementation for the interface `goo` of type (λ [] Int) at line 1, column 16 in 'REPL'.

None of the possibilities have the correct signature:
     at REPL:1:1.
鲤 
```

So right now only functions with the same name work...

```
鲤 (defn goo [] 3)
鲤 (implements goo goo)
鲤 :i goo
goo : (λ [] Int) = {
    goo
}
Defined at line 1, column 15 in 'REPL'
goo : (λ [] Int)
鲤 (defn boo [] (goo)) 
鲤 (boo)
```

But I can't seem to figure out where this restriction comes into play.

Another thing you might notice from the above sample, this mechanism makes it possible for us to provide default implementations for interfaces :D (treating the global fn with the same name as the default).

I think? This has something to do with contexts/envs but I'm not sure. I'm lacking sufficient context on how exactly the relationship between envs works and where exactly they come into play. There are no remaining obvious places (at least not to me) where the interface implementations still rely on having the same name, and yet the behavior persists.

src/Primitives.hs
scolsen: This particular case is fixed now! But I think some environment problems are still persisting when it comes to using implementation with different names than the interface. :/

src/StartingEnv.hs
scolsen: Done, nice catch, thanks!

scolsen:
Aaaaaaand now I'm getting runtime errors for the tests--help! I'm going to sleep now 😴

scolsen:
Yeah, I'm stumped. The behavior of this primitive seems wildly inconsistent! I must've gotten something wrong

scolsen:
Looked into this a little bit. So, weirdly, `Bool.=` *is* added to the `=` interface:

```
interlook: Just = : (λ [a, a] Bool) = {
    Bool.=
}
```

Also, looks like some piece of the compiler is not honoring the new primitive!

```
interlook: Just > : (λ [a, a] Bool) = {
    Byte.>
    ByteRef.>
    Int.>
    Long.>
    Double.>
    Float.>
    PairRef.=
}
```

`>` shouldn't be a member here.

scolsen:
I was passing the wrong variable to the type environment .... -.-'

scolsen:
Alrighty, we're almost there. There's just a few more spots I need to add `implements` calls to, then all the tests should pass.

scolsen:
Ok, everything is passing now and I added some new tests for basic assumptions about the way this should behave.

@eriksvedang, @hellerve it's ready for a look!

Since this is a pretty major change to the way interfaces work from a user perspective, we should probably do sue diligence and figure out some way to soften the blow on users before merging this, if we decide to merge it, whether that's:

1. Waiting until a fresh release?
2. Sending out some notice?
3. Opening courtesy PRs against repos that add calls to the new primitive where needed.

Also note that the behavior for declaring an implementation before a defintion isn't super smooth. It works the same way as doc insofar as this:

```
(defmodule Boo
  (implements foo Boo.baz)
  (defn baz ....)
)
```

results in a meta-set error about `Boo.baz` not being found, but unqualified is fine (just like `doc`):

```
(defmodule Boo
  (implements foo baz)
  (defn baz ....)
)
```

For `implements` this slight UX thorn is perhaps slightly thornier than for doc since this also works:

```
(defmodule Boo
  (defn baz ....)
  (implements foo Boo.baz)
)
```

Which might lead one to believe leaving the qualification as is should work in the `implements` preceding the definition case, but it currently doesn't.

Otherwise, I think everything is pretty much good to go and ready for a review.

src/Primitives.hs
scolsen: All good. I had a really dumb typo that broke nearly everything -.-

hellerve:
I can’t say that I scrutinized the entire PR meticulously, but the compiler parts look fine (my comment above notwithstanding) and the tests run, so I believe this works as expected!

scolsen:
> Currently, calls to `implements` where the interface doesn’t exist do not seem to generate any kind of feedback and just silently do nothing (this is in line with the old behavior of `registerDefnOrDefInInterfaceIfNeeded`, where this was necessary).
> 
> I would argue that we should at least emit a warning about that.

Right, this was a bit of a question for me--while I was implementing this, I ultimately decided to retain the retroactive interface registration (`retroactivelyRegisterInterfaceFunctions`) to have behavioral parity. But you make a great point! With the new setup, where things are explicit, it kind of makes more sense to at least warn about this.

scolsen:
@hellerve 446972f emits warnings like you suggested! Thanks for taking a look!

src/Lookup.hs
```diff
@@ -90,6 +93,51 @@ multiLookupInternal allowLookupInAllModules name rootEnv = recursiveLookup rootE
           in --(trace $ "multiLookupInternal '" ++ name ++ "' " ++ show (envModuleName env) ++ ", spine: " ++ show (fmap snd spine) ++ ", leaves: " ++ show (fmap snd leaves) ++ ", above: " ++ show (fmap snd above))
             spine ++ leaves ++ above
 
+binderToEnv :: Binder -> Maybe Env
```
eriksvedang, line 96: The new function in this module look nice and general, but since there are no removed code I'm a bit wary of all the code duplication. Many of them seem to be slightly different versions of the ones already there, like `multiLookupALL`, etc. Some more general and some less so. Since you have this fresh in your mind @scolsen, do you agree with my sentiment?

scolsen: Yep that's correct--nearly all of it is just lifted out of the body of `multiLookup`

I agree we should remove the dupes....I didn't remove them in this set of commits since I felt that'd be a potentially tricky change to get right (since we need to make sure the new approach doesn't break `multiLookup`!) So I opted for leaving the duplication in here to keep this set of changes focused on the `implements` stuff only.

I could add a TODO here if that makes sense, unless you think it's OK to go ahead with a complete refactor of Lookups based on the more generic functions.

scolsen:
> Like you said @scolsen, using meta data for this is a slight stretch but I guess that part can be changed later if it turns out to be too messy. Most of the logic still belongs in the interface declaration itself, which is reassuring. Am I correct in thinking that the meta data is only really necessary for the forward declarations?

Yep, that's right! Turns out the `implements` Meta is only used to allow the forward declarations. It basically just flag guards a call to the interface logic in `define`.

src/Lookup.hs
eriksvedang: Yeah, I think some TODO:s would be a good solution.

eriksvedang:
> > Like you said @scolsen, using meta data for this is a slight stretch but I guess that part can be changed later if it turns out to be too messy. Most of the logic still belongs in the interface declaration itself, which is reassuring. Am I correct in thinking that the meta data is only really necessary for the forward declarations?
> 
> Yep, that's right! Turns out the `implements` Meta is only used to allow the forward declarations. It basically just flag guards a call to the interface logic in `define`.

OK, perfect.

src/Lookup.hs
scolsen: added!

PR #770, opened 2020-05-09
Fix unicode handling

### LINKED ISSUE #708: Unicode characters generate invalid c

jacereda:
Should fix #708. Builds on top of #768.


hellerve:
I suppose the signature of `String.cstr` should also be changed to `Ptr CChar` then, to enable C interop?

jacereda:
> I suppose the signature of `String.cstr` should also be changed to `Ptr CChar` then, to enable C interop?

What do you mean? Unless I messed something it should return `Ptr CChar` now...

jacereda:
The last commit should allow things like:

```clojure
(def 😀 \😀)
```

jacereda:
Works on clang, gcc has accepted the patch quite recently and won't support it until v10:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67224

hellerve:
Oh man, I don’t know why I keep commenting today, I definitely need to go to bed! I must’ve scrolled past your change to `core/String.carp` repeatedly...

Looks good to me!

src/Parsing.hs
```diff
@@ -223,7 +224,8 @@ symbolSegment :: Parsec.Parsec String ParseState String
 symbolSegment = do sym <- Parsec.many1 validInSymbol
                    incColumn (length sym)
                    return sym
-  where validInSymbol = Parsec.choice [Parsec.letter, Parsec.digit, Parsec.oneOf validCharacters]
+  where validInSymbol = Parsec.choice [Parsec.letter, Parsec.digit, Parsec.oneOf validCharacters, highCharacters]
+        highCharacters = Parsec.satisfy ((> 127) . ord)
```
hellerve, line 228: You probably shouldn’t listen to me tonight, because I clearly am not on top of my game, but I suppose this will also allow Unicode whitespace?

hellerve: Fun fact: we allow this now(I tried it with an Em Quad), but Clang does not:
```
> (def   1)
>  
  : Int
> (defn main []  )
> :bx
/Users/veitheller/.carp/out/main.c:1424:5: warning: treating Unicode character as whitespace
      [-Wunicode-whitespace]
int  ;
    ^
/Users/veitheller/.carp/out/main.c:1424:1: warning: declaration does not declare anything
      [-Wmissing-declarations]
int  ;
^~~
/Users/veitheller/.carp/out/main.c:2903:9: warning: treating Unicode character as whitespace
      [-Wunicode-whitespace]
          = 1;
        ^
/Users/veitheller/.carp/out/main.c:2903:13: error: expected expression
          = 1;
          ^
/Users/veitheller/.carp/out/main.c:8758:12: warning: treating Unicode character as whitespace
      [-Wunicode-whitespace]
    return  ;
           ^
/Users/veitheller/.carp/out/main.c:8758:5: error: non-void function 'main' should return a value
      [-Wreturn-type]
    return  ;
    ^
4 warnings and 2 errors generated.
carp: callCommand: clang -fPIC -lm -g /Users/veitheller/.carp/out/main.c -o "/Users/veitheller/.carp/out/Untitled"  -I/Users/veitheller/Documents/Code/Github/carp/Carp//core/   (exit 1): failed
```

eriksvedang:
This and all the later PRs look great! I'm a little wary of merging without testing on Windows first, so I'll try to run the test suite on my Windows machine tomorrow.

eriksvedang:
Works really nice on my Mac. A bit more problematic on Windows, as feared.

First of all there are lots of warnings about Char_copy, like this:

```bash
out\main.c(7068,25): warning: incompatible pointer types passing 'char *' to parameter of type 'const Char *' (aka 'const unsigned int *') [-Wincompatible-pointer-types]
    char _8 = Char_copy(a);
```` 

And then clang-cl.exe stops at this:
```bash
out\main.c(11661,31): error: character too large for enclosing character literal type
        String _37 = Char_str('ñ');
                              ^
out\main.c(11670,31): error: character too large for enclosing character literal type
        String _52 = Char_str('😲');
```

I'm all for getting this merged quickly, so not running these tests on Windows is an acceptable solution in my mind, but perhaps there is a fix we can try? @jacereda

hellerve:
That’s weird, as I understand it the second two errors should not be there since they characters should be prefixed by `U`. Did you run this with the version of Carp from this PR?

eriksvedang:
Will double check...

eriksvedang:
Oh, I missed that the build script on Windows doesn't rebuild the Haskell project. Now I get other errors...

eriksvedang:
This is what I get now:

```bash
Test 'str works with ASCII' passed
Test 'str works with Latin1' failed:
        Expected value: 'Ã±', actual value: 'ñ'
Assertion failed: sz > 0, file C:\Users\erik\Documents\Carp/core\carp_string.h, line 196
```

jacereda:
Can you check if the format of both the .carp and .c files is UTF-8?

eriksvedang:
I think they are, this is what Emacs tells me about them:
```
U -- utf-8-dos (alias: mule-utf-8-dos cp65001-dos)

UTF-8 (no signature (BOM))
Type: utf-8 (UTF-8: Emacs internal multibyte form)
EOL type: CRLF
This coding system encodes the following charsets:
  unicode
```

eriksvedang:
Or wait, the `main.c` is something else:
```
1 -- iso-latin-1-dos (alias: iso-8859-1-dos latin-1-dos)

ISO 2022 based 8-bit encoding for Latin-1 (MIME:ISO-8859-1).
Type: charset (charset)
EOL type: CRLF
This coding system encodes the following charsets:
  iso-8859-1
```

jacereda:
That's the problem then, the result should be encoded as UTF-8.

eriksvedang:
OK, I'll fix that.

jacereda:
If you find you can't print unicode characters on Windows, try with

```
SetConsoleOutputCP(CP_UTF8);
```

eriksvedang:
Saving as UTF-8 fixed that error, thanks @jacereda !
But there's a new one...

```
C:\Users\erik\Documents\Carp/core\carp_string.h(194,48): warning: format specifies type 'wint_t' (aka 'unsigned short') but the argument has type 'int' [-Wformat]
    int sz = snprintf(buf, sizeof(buf), "%lc", c);
                                         ~~~   ^
                                         %c
1 warning generated.
Test 'str works with ASCII' passed
Test 'str works with Latin1' failed:
        Expected value: '', actual value: 'Assertion failed: sz > 0, file C:\Users\erik\Documents\Carp/core\carp_string.h, line 197
```

jacereda:
Does something like `"%lc", (wint_t)c` do the trick?

eriksvedang:
It silenced the warning, but the assert still fails.

jacereda:
maybe the `setlocale` should use `LC_ALL`?

eriksvedang:
Same error.

eriksvedang:
Now the Windows tests are run on push, so that might help in debugging this, I hope...

eriksvedang:
Awesome work, thanks so much @jacereda !!!

src/Parsing.hs
hellerve: This was not fixed, was it? Should I take this on or do you want to, @jacereda?

jacereda: I think I fixed it in some other place (`cross` or `tcc` PRs).

hellerve: Ah cool, nevermind then! (And thank you)

jacereda: Nope, looks like it should be in a PC where I did some experiments compiling with Visual C. I think the change was just emitting via something like:

```
show (ord c) ++ "/*" ++ show c ++ "*/"
```

hellerve: Oh cool!

Maybe we should disallow it in the parser, though, using something like `isSpace` from `Data.Char` (that should work on Unicode whitespace)? Whitespace variables don’t quite feel right to me.

jacereda: Ah, I thought the problem was related to values. No, I haven't worked on that, but sounds like `isSpace` could do the trick.

PR #772, opened 2020-05-10
Flamegraph

jacereda:
Script to generate a clickable flamegraph svg.

This one also builds on top of #768.

Sample output:

[out.zip](https://github.com/carp-lang/Carp/files/4605566/out.zip)

### DIFF
--- a/flamegraph.sh
+++ b/flamegraph.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+# To enable the required permissions, execute the following sequence as root:
+# $ chown root:wheel /proc/modules
+# $ chown root:wheel /proc/kallsyms
+# $ echo 0 > /proc/sys/kernel/kptr_restrict
+# $ echo -1 > /proc/sys/kernel/perf_event_paranoid
+#
+# Usage:
+#
+# Profile an executable
+# ./flamegraph.sh <executable>
+#
+# To change the frequency:
+# HZ=100 ./flamegraph.sh <executable>
+#
+# To use the maximum frequency
+# HZ=max ./flamegraph.sh <executable>
+#
+# This can be useful to see how much time is spent in carp vs clang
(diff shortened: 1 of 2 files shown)


flamegraph.sh
```diff
@@ -0,0 +1,28 @@
...
+# This can be useful to see how much time is spent in carp vs clang
+# ./flamegraph.sh ./carp.sh -b <file.carp>
+
+HZ=${HZ:-1000}
+prg=`basename $1`
+svg=`mktemp XXXXXX.svg`
+perf record -F $HZ -a -g -- $*
+perf script | stackcollapse-perf.pl | flamegraph.pl > $svg
+xdg-open $svg
+rm $svg
```
bjorn3, line 28: This will likely remove the svg before the view application had a change to open it.

jacereda: Oh, right. It seems to work for me though. 

Is there some way to view the svg sinchronously?

bjorn3: Maybe use `read -p "Press enter to delete the temporary svg file" some_var`?

hellerve: Does `xdg-open` exit before reading the file?

jacereda: Not that I like it, but perhaps a `sleep 5` will be less intrusive than prompting...

bjorn3: > Does xdg-open exit before reading the file?

`xdg-open` runs the program to open a file and then it exits. It doesn't wait for the program to exit.

eriksvedang:
The diff on this one is confusing. Very cool tool though, will merge once rebased.

eriksvedang:
(sorry about the docs, they changed now when we replaced lambda with Fn when printing signature)

jacereda:
Rebased.

eriksvedang:
Thanks!

PR #780, opened 2020-05-11
Fix the `inline-c` macro to avoid using a reserved keyword as parameter.

linkpy:
The title says it all. Tiny fix for `inline-c`.


eriksvedang:
Thanks!

PR #781, opened 2020-05-11
Makes test bash script running on MSYS/Git Bash

TimDeve:
It all works except some for some logging from the C compiler/linker `Creating library out\Untitled.lib and object out\Untitled.exp` seems to be related to [this](https://stackoverflow.com/questions/2820694/why-does-my-visual-c-exe-project-build-create-lib-and-exp-files) but I'm not sure how to disable the logging.

It uses the dos2unix binary to turn the output of the run so it matches the expected ouput. I think it comes by default with MSYS but will need to check, sed could be used instead.

Not sure why this check is different as well:
```
<     A.some-interface : (Fn [Int] Bool)
---
>     A.some-interface : (λ [Int] Bool)
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -2,8 +2,9 @@ name: Windows CI
 
 on:
   push:
-    branches-ignore:
-      - '**'
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
   build:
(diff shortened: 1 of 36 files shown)


TimDeve:
Diff has a command to ingnore line endings so I'm gonna use that instead.

TimDeve:
The PR also reenables the Windows CI

eriksvedang:
Wohoo!

Issue #776, opened 2020-05-10
Array.range should guard against nonsensical input


hellerve:
The `range` function currently will happily accept input like `(range 0 10 -1)` or even `(range 0 0 0)`, the latter of which crashes.

It should probably guard against these kinds of inputs and return a `Result` type. This will change the signature of `Array.enumerated` as well, which depends on `range`.

Cheers

PR #782, opened 2020-05-11
Makes enumerated work with empty arrays

TimDeve:
This is a work around for https://github.com/carp-lang/Carp/issues/776 the proper fix would be to allow range to take `(range 0 0 1)` inputs but the decision about what the correct behavior should be hasn't been made.

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -315,9 +315,11 @@ The trailing elements of the longer array will be discarded.")
 
   (doc enumerated "creates a new array of `Pair`s where the first position is the index and the second position is the element from the original array `xs`.")
   (defn enumerated [xs]
-    (zip &Pair.init-from-refs
-         &(range 0 (length xs) 1) ; Inefficient, creates a temporary array.
-         xs))
+    (if (Array.empty? xs) ; Only there because range will crash with a length of 0
+      []
+      (zip &Pair.init-from-refs
+           &(range 0 (length xs) 1) ; Inefficient, creates a temporary array.
+           xs)))
 
   (doc nth "gets a reference to the `n`th element from an array `arr` wrapped on a `Maybe`.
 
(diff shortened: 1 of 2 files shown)


test/array.carp
```diff
@@ -191,6 +191,10 @@
                 "[(Pair 0 @\"a\") (Pair 1 @\"b\") (Pair 2 @\"c\")]"
                 &(str &(Array.enumerated &[@"a" @"b" @"c"]))
                 "enumerated works as expected")
+  (assert-equal test
+                "[]"
+                &(str &(Array.enumerated &(the (Array Int) [])))
```
hellerve, line 196: You should be able remove stringification here and compare the arrays directly by ref.

core/Array.carp
```diff
@@ -315,9 +315,11 @@ The trailing elements of the longer array will be discarded.")
 
   (doc enumerated "creates a new array of `Pair`s where the first position is the index and the second position is the element from the original array `xs`.")
   (defn enumerated [xs]
-    (zip &Pair.init-from-refs
-         &(range 0 (length xs) 1) ; Inefficient, creates a temporary array.
-         xs))
+    (if (Array.empty? xs) ; Only there because range will crash with a length of 0
```
hellerve, line 318: We could also get rid of the intermediate array and this special case if we transform this to an explicit `for` loop. This is not necessarily nicer, but it would solve both problems at once.

TimDeve: Yeah a more procedural approach was what I had in mind to remove range and not have it's new signature "contaminate" enumerated. Went with the quickest here. I'll try to update the PR later.

test/array.carp
TimDeve: Just copy pasted the test above but you're right.

core/Array.carp
hellerve: Yeah, no worries!

TimDeve: Updated.

hellerve:
Looks great now! Thanks for the updates!

PR #786, opened 2020-05-12
Move templates out of the compiler into deftemplates

hellerve:
This PR moves as many functions as possible from compiler internals into `deftemplate` forms (thank you, @linkpy, this is great!). This means that we can get rid of a lot of code, and make the compiler even smaller.

A few comments:
- Moving the `copy` and `delete` functions outside led to weird results, so I opted not to do that for now. Same for everything concerning functions.
- The (static) array templates almost all have dependencies, which we can’t represent yet. They stay in the compiler as well.
- Some of the lifetimes were static before and are now variables. I do not know whether this could lead to problems, but the test suite still passes.

Cheers


eriksvedang:
Very cool!

How do you feel about just moving the calls to `deftemplate` from `Templates.carp` into their corresponding files, I feel like it will be a lot easier to find them that way (you won't have to know whether they are templates or just defns, etc).

hellerve:
Generally I think this is a good idea, it will just require a little bit of reshuffling of `core/Core.carp` because those functions were expected to be there from the beginning and are required all over the place (`Pointer`, for instance, is used from within `Generics`).

PR #788, opened 2020-05-12
Add str on refs for numeric types

hellerve:
This PR adds `str` for references on numeric types to make things like `(Double.+ 2.0 3.0)` work on the REPL, which implicitly call `str` on the `ref` of the result.

Cheers

### DIFF
--- a/test/execute.sh
+++ b/test/execute.sh
@@ -2,7 +2,7 @@
 
 # Runs the executable and compares its output to the .expected file
 ./carp.sh $1 --log-memory -b && \
-  ./out/Untitled > test/output/$1.output.actual 2>&1
+  ~/.carp/out/Untitled > test/output/$1.output.actual 2>&1
 echo $1
 
 if ! diff --strip-trailing-cr test/output/$1.output.actual test/output/$1.output.expected; then
--- a/test/execute.sh
+++ b/test/execute.sh
@@ -2,7 +2,7 @@
 
 # Runs the executable and compares its output to the .expected file
 ./carp.sh $1 --log-memory -b && \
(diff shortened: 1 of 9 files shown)


hellerve:
Please note that this probably collides with #769, which also works on interfaces. Depending on the order of the merges, either @scolsen or I will have to do a little more work to make this work.

test/execute.sh
```diff
@@ -2,7 +2,7 @@
 
 # Runs the executable and compares its output to the .expected file
 ./carp.sh $1 --log-memory -b && \
-  ./out/Untitled > test/output/$1.output.actual 2>&1
+  ~/.carp/out/Untitled > test/output/$1.output.actual 2>&1
```
TimDeve, line 5: That's gonna break peoples/the CI run if they don't have their out folder setup like yours. Using `-x` like it use to be would solve that but I'm not sure how to not have the compiler output in the diff.

hellerve: Oh yeah, I didn’t want to commit that, sorry. I just need to do that to make the tests work on my machine atm.

TimDeve: Maybe we should hardcode the out folder in the Carp tests for now.

hellerve: Right, but this script is also used for the examples. I’m just going to work on the CLI flag tonight I think.

eriksvedang:
Your prediction was right @hellerve, needs some fixing now.

hellerve:
Updated. I’d wait for the tests to pass before merging this, though.

eriksvedang:
Seems fine 💯

PR #789, opened 2020-05-12
Add mini-guide on implementing primitives

scolsen:
Just a small walkthrough on the common steps for implementing a new primitive.

### DIFF
--- a/docs/hacking.md
+++ b/docs/hacking.md
@@ -307,13 +307,19 @@ calling `set!` on it.
     return $ evalError ctx ("`immutable` expected a single symbol argument, but got" ++ show xobjs) (info xobj)
   ```
 
-And that wraps up the core logic of our primitive. To make it available, we just need to register it in `StartingEnv.hs`:
+And that wraps up the core logic of our primitive. To make it available, we
+just need to register it in `StartingEnv.hs`.
+
+
+#### 1. Add your primitive to the starting environment
+
+To add a primitive to the starting environment, call `makePrim`:
 
 ```
 , makePrim "immutable" 1 "annotates a variable as immutable" "(immutable my-var)" primitiveImmutable
+```
 
 That's about it. Note that this implementation just adds special metadata to
 bindings--to actually prevent users from calling `set!` on an immutable `def`
 we'd need to update `set!`'s logic to check for the presence of the `immutable`
 metadata.
-```
--- a/docs/hacking.md
+++ b/docs/hacking.md
@@ -21,7 +21,7 @@ The sections below briefly describe the purpose of each stage and list
 important source files. You can use these sections to get a rough idea of what
 files you might need to edit in order to alter the functionality of a
 particular phase.
- 
+
 > Note: Some sources contain definitions that are important or used in pretty
 > much every phase of the compiler, in result some files may appear more than
 > once in the sections below.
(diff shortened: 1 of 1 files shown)


docs/hacking.md
```diff
@@ -176,3 +176,150 @@ files that serve different purposes in the compiler:
...
+high-level steps. The following sections provide some guidance on making such
+changes.
+
+### Adding a new Primitive
+
+If it doesn't require anything fancy or out of the ordinary, adding a new
+primitive to the compiler entails the following:
+
+1. Define your new primitive in `Primitives.hs`
+2. Add your primitive to the staring environment using `makePrim` in
```
hellerve, line 191: Typo: `staring`->`starting`.

docs/hacking.md
```diff
@@ -176,3 +176,150 @@ files that serve different purposes in the compiler:
...
+
+Primitives are functions of the `Primitive` type:
+
+```
+type Primitive = XObj -> Context -> [XObj] -> IO (Context, Either EvalError XObj)
+```
+
+Every primitive takes an xobj, the form that represents the primitive, a
+compiler context, and a list of XObjs the primitive form's arguments.
+Primitives return either a new `Context` updated based on the logic they
```
hellerve, line 204: Clarification: Primitives always return a context, and either an `XObj` or an `EvalError`.

docs/hacking.md
```diff
@@ -176,3 +176,150 @@ files that serve different purposes in the compiler:
...
+
+For example, here's how the `defmodule` primitive maps to the `Primitive` type:
+
+```
+(defmodule Foo (defn bar [] 1))
+ |         |-----------------|
+ XObj      [XObj] (arguments)
+```
+
+> The `Context` argument is a compiler concept and doesn't have a corresponding
```
hellerve, line 215: Maybe add that it stands for the state of the compiler?

docs/hacking.md
```diff
@@ -176,3 +176,150 @@ files that serve different purposes in the compiler:
...
+            return $ evalError ctx ("`immutable` expects a variable as an argument") (info xobj)
+  primitiveImmutable _ _ xobjs = -- any other number or types of xobj arguments are incorrect! Let's error.
+    return $ evalError ctx ("`immutable` expected a single symbol argument, but got" ++ show xobjs) (info xobj)
+  ```
+
+And that wraps up the core logic of our primitive. To make it available, we
+just need to register it in `StartingEnv.hs`.
+
+
+#### 1. Add your primitive to the starting environment
```
hellerve, line 314: Why is this heading numbered?There do not seem to be any other numbers anywhere?

scolsen: There is another numbered heading above, but I think it makes sense just to drop the numbers entirely (I numbered this one wrong anyways).

docs/hacking.md
scolsen: Thanks, done!

docs/hacking.md
scolsen: Ah good catch! Thank you!

docs/hacking.md
scolsen: Thanks!

scolsen:
Thanks for taking a look, @hellerve 🙇

hellerve:
Looks really great now, thank you!

eriksvedang:
Such a nice read @scolsen . Very pedagogical!

PR #790, opened 2020-05-12
Make Num.from-string better

hellerve:
This PR changes `from-string` for all the number types to actually fail when the string is not a valid number. This means that all of them return a `Maybe` now. Under the hood this is achieved by checking whether the `endptr` argument to `strto*` is `\0`, which means that we reached the end of the string after parsing.

Of note: these functions will still consume any leading whitespace, which is the behavior of `strto*`.

Cheers


eriksvedang:
Oh, that's really nice!

eriksvedang:
Needs some love.

Issue #791, opened 2020-05-12
`set!` erases type info and causes programs to hang


scolsen:
Currently, calling `set!` on a global variable erases the type information from the initial variable and, if the type variable is used, causes carp to hang:

```
;; erasure
鲤 (def foo 3)
鲤 :i foo
foo : Int
Defined at line 1, column 1 in 'REPL'
鲤 (set! foo 4)
鲤 :i foo
(defdynamic foo 4) : Dynamic
Defined at line 1, column 11 in 'REPL'
鲤 
``` 

This program will cause a hang, removing the call to `set!` makes everything run fine:

```
(def foo 3)                                                                                                                                                                                                                                                     
(set! foo 4)                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                
(defn main []                                                                                                                                                                                                                                                   
  (+ 1 foo)                                                                                                                                                                                                                                                     
)
```

Luckily, I think the fix for this might be pretty simple. Currently, `specialCommandSet` reassigns a `defs` type to `Dynamic` whenever `set!` is called. My hope is that simply retaining the type instead does the trick, though we might have to distinguish between `def` cases and `defdynamic` cases.

hellerve:
As discussed in the chat, the problem currently is that Carp has to do some “heuristics” to figure out whether this `set!` is dynamic or static. It looks like it currently infers that it is static whenever it’s global; it should probably be a little smarter than that! We’ll have to match on the thing that is passed in.

scolsen:
#793 gets us part of the way there by adding a basic check for `Defs`--but it doesn't seem to actually resolve the issue that dynamic eval code is called on every set-bang. Seems like the emission level setbang code doesn't run.

PR #793, opened 2020-05-12
`Immutable` primitive

scolsen:
This PR adds a new primitive `immutable` that marks variables as immutable. Calling `set!` on an immutable variable will throw an error.

Example use:

```
鲤 (def foo 4)
鲤 (immutable foo)
鲤 (set! foo 3)
Tried to set foo but it was declared immutable. at REPL:1:7.
鲤 
```

Counts toward #791


src/Emit.hs
```diff
@@ -432,19 +432,22 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
...
-                     Just varInfo = info variable
-                 --appendToSrc (addIndent indent ++ "// " ++ show (length (infoDelete varInfo)) ++ " deleters for " ++ properVariableName ++ ":\n")
-                 delete indent varInfo
-                 appendToSrc (addIndent indent ++ properVariableName ++ " = " ++ valueVar ++ "; "
-                              ++ " // " ++ show (fromMaybe (VarTy "?") (ty variable)) ++ " = " ++ show (fromMaybe (VarTy "?") (ty value))
-                              ++ "\n")
-                 return ""
+              let immutable = metaIsTrue meta "immutable"
+              in  if immutable
+                  then error (show (CannotSet variable))
```
hellerve, line 437: Should we add this reason to the `CannotSet` type error!

eriksvedang:
Very cool, we really need something like this!

Unfortunately, one thing I don't like about this implementation is that it won't work on local variables right now, since they can't carry meta data. That could probably be fixed though, and is more of a limitation of the rest of the compiler. It might be good to look into first though, since having the ability to only make certain variables immutable seems like a strange thing to have to explain to people.

Also, can't the `immutable` primitive be implemented as a macro, using `meta-set!`?

scolsen:
> Very cool, we really need something like this!
> 
> Unfortunately, one thing I don't like about this implementation is that it won't work on local variables right now, since they can't carry meta data. That could probably be fixed though, and is more of a limitation of the rest of the compiler. It might be good to look into first though, since having the ability to only make certain variables immutable seems like a strange thing to have to explain to people.
> 
> Also, can't the `immutable` primitive be implemented as a macro, using `meta-set!`?

Ah, good point. And you're right that the primitive could be implemented as a macro!

Regarding the let bindings, I'm not sure if it's better to improve let bindings by allowing them to take metadata or if it's better to instead introduce a new XObj for immutable variables instead. Part of me feels like using the distinct XObj is slightly cleaner and would be a bit easier to handle/maintain in the long term than relying on metadata...but then we're increasing the number of compiler changes and XObjs which isn't great.

eriksvedang:
Should we close this for now?

eriksvedang:
(until we've got meta on local bindings, I guess..?)

scolsen:
Yeah, I think we can close this for now!—I have to think about the implementation a bit more.

PR #795, opened 2020-05-12
Flamegraph for the haskell side.

jacereda:
Generate flamegraphs for the compiler.
```
$ nix-shell --arg profiling true
$ ./haskell-flamegraph.sh -b test/carp.sh
```
Generates the attached carp.svg.
[carp.svg.zip](https://github.com/carp-lang/Carp/files/4618746/carp.svg.zip)

### DIFF
--- a/carp.sh
+++ b/carp.sh
@@ -7,4 +7,4 @@ then
 	CARP="cabal -v0 run carp"
     fi
 fi
-$CARP "--" $*
+$CARP $BUILD_OPTS $"--" $*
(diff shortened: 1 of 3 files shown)


hellerve:
Do you have an idea why this build failed? It doesn’t look like this has anything to do with your PR.

jacereda:
No idea, can the build be triggered again without pushing more commits?

hellerve:
I don’t think so. If so, Erik is probably the only person with the power to do that!

eriksvedang:
Re-running.

hellerve:
Looks good now!

PR #796, opened 2020-05-13
Add FreeBSD as a supported platform

dbalan:
This commit adds FreeBSD as a supported platform. I understand that this may not be as easy as just adding it to the list, but if can point me at things that I should test, I can do that.

Tests (on FreeBSD 12.1-RELEASE, amd64)
- [x] Compiler builds and runs on FreeBSD
- [x] Compiles and runs the example SDL program
- [x] Builds and runs reptile.carp from examples.
- [x] Test 'exp works as expected' test

### DIFF
--- a/core/Core.carp
+++ b/core/Core.carp
@@ -51,3 +51,5 @@
 (not-on-windows
   (system-include "sys/wait.h")
   (system-include "unistd.h"))
+(freebsd-only
+  (system-include "signal.h"))
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -548,6 +548,11 @@
 (defmacro update! [x f]
   (list 'set! x (list f x)))
 
+(defmacro freebsd-only [:rest forms]
+  (if (= "freebsd" (os))
+    (eval (cons (quote do) forms))
+    ()))
+
 (defmacro mac-only [:rest forms]
   (if (= "darwin" (os))
     (eval (cons (quote do) forms))
--- a/src/Util.hs
+++ b/src/Util.hs
@@ -61,7 +61,7 @@ enumerate 5 = "sixth"
 enumerate 6 = "seventh"
 enumerate n = show n ++ ":th"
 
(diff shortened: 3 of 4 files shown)


hellerve:
Hey @dbalan, thanks for your contribution! You can try running the test suite using `sh run_carp_tests..sh`. If that works I’d say we’re good to go!

hellerve:
Sadly[it seems as if FreeBSD is not supported for GitHub workflows](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on), otherwise I’d suggest that we add a CI target for it.

dbalan:
I usually use <https://builds.sr.ht/> for FreeBSD CI. I was running a test job for this PR there and got this error
```
Test 'exp works as expected' failed:
	Expected value: '2.71828', actual value: '2.71828'
```

Full log: https://builds.sr.ht/~dbalan/job/207840#task-cryptopals-548

I will check whats going on, but maybe you have some insights (since those two numbers look same)

hellerve:
Thanks for posting the logs! I’m suspecting that the issue is that the doubles aren’t quite the same bit by bit, but the printing precision doesn’t show the divergence. Since these are hardware doubles, it’s often a little hard to get exact precision.

For these cases there is `Double.approx`. You can try switching the test case to use that function like so:

```clojure
  (assert-op test
             Double.e
             (exp 1.0)
             "exp works as expected"
             Double.approx)
```

dbalan:
@hellerve That fixed my test cases!

hellerve:
Yay! Is there any way we can integrate https://builds.sr.ht builds for FreeBSD into the CI pipeline here? Sadly I don’t really know anything about it and I’m not feeling like learning _yet another_ CI tool right now. Is this something you could/would want to champion?

dbalan:
I can help setup the CI, but I think its better kept to a separate PR. What do you think?

Also here is the build log for current PR-head (status: success :partying_face:)

hellerve:
I agree re: separate PR!

PR #799, opened 2020-05-13
Add String.to-bytes

### LINKED ISSUE #794: Add `String.to-bytes`
Since we now no longer live under the assumption that a string is a collection of bytes (kudos to us, especially @jacereda), we should probably add an escape hatch to get back to a simpler time where they were. This is useful when implementing encodings, even simple ones.

Carpentry’s UTF8 library in particular builds on that notion.

Cheers

hellerve:
This PR fixes #794 by adding `String.to-bytes`.

Test cases are included.

Cheers


hellerve:
It seems as if a fix for `Pointer.to-value` snuck into this PR.

PR #803, opened 2020-05-14
Added --compile-fast to compile with tcc.

jacereda:
Not sure if this is the best approach. Maybe we should just read a `CC` environment variable?

Timings on some very old hardware:

- `CARP_OPTS=--compile-fast ./run_carp_tests.sh --no_sdl`: 2m:51
- `./run_carp_tests.hs --no_sdl`: 8m58

### DIFF
--- a/core/carp_long.h
+++ b/core/carp_long.h
@@ -24,7 +24,7 @@ bool Long_safe_MINUS_sub(Long x, Long y, Long* res) {
 bool Long_safe_MINUS_mul(Long x, Long y, Long* res) {
     Long r = x * y;
     *res = r;
-    return (r / y) != x;
+    return y == 0 || (r / y) != x;
 }
 #else
 bool Long_safe_MINUS_add(Long x, Long y, Long* res) {
--- a/core/carp_long.h
+++ b/core/carp_long.h
@@ -10,7 +10,23 @@ Long Long__MUL_(Long x, Long y) {
 Long Long__DIV_(Long x, Long y) {
     return x / y;
 }
-#ifndef _WIN32
+#if defined _WIN32 || defined __TINYC__
+bool Long_safe_MINUS_add(Long x, Long y, Long* res) {
+    Long r = x + y;
+    *res = r;
+    return (y > 0) && (x > (INT64_MAX - y)) || (y < 0) && (x < (INT64_MIN - y));
+}
+bool Long_safe_MINUS_sub(Long x, Long y, Long* res) {
+    Long r = x - y;
+    *res = r;
+    return (y > 0 && x < (INT64_MIN + y)) || (y < 0 && x > (INT64_MAX + y));
+}
+bool Long_safe_MINUS_mul(Long x, Long y, Long* res) {
+    Long r = x * y;
+    *res = r;
+    return (r / y) != x;
+}
+#else
 bool Long_safe_MINUS_add(Long x, Long y, Long* res) {
     return __builtin_add_overflow(x, y, res);
 }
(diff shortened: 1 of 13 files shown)


jacereda:
Forgot to say that this also fixes non-standard code generated for `(deftype Simple [])`. C doesn't allow empty structs, even if some compilers allow it. It just generates a `char __dummy` field. Should also help with Visual C.

core/carp_long.h
```diff
@@ -10,7 +10,23 @@ Long Long__MUL_(Long x, Long y) {
...
+}
+bool Long_safe_MINUS_sub(Long x, Long y, Long* res) {
+    Long r = x - y;
+    *res = r;
+    return (y > 0 && x < (INT64_MIN + y)) || (y < 0 && x > (INT64_MAX + y));
+}
+bool Long_safe_MINUS_mul(Long x, Long y, Long* res) {
+    Long r = x * y;
+    *res = r;
+    return (r / y) != x;
```
hellerve, line 27: Won’t this explode if `y` is `0`?

jacereda: Doh! Fixed, thanks!

I also fixed the timings in the description, they were reversed.

hellerve:
This seems a bit like a “miscellaneous" PR. Are all of the features still relevant? I think there was some discussion about changing the TCC bit in #774?

jacereda:
Well, besides adding the tcc stuff it also increases the chances it will compile in non-clang compilers. I'll send another PR to add the `--cc` flag but this one should probably be rebased on top to provide the empty struct fix and the safe ops fallback.

jacereda:
@eriksvedang I think the windows build might be a problem with the CI machine. I just built on Windows with that same resolver.

GrayJack:
I do think it would be a better idea to have a `--cc` or read the `CC` environment variable

Or maybe both? Default to `CC` but if `--cc` specified, use what was specified

hellerve:
Yes, we do. I think it might make sense to also give an official precedence/semantics to potentially over-lapping configurations like this. I’d propose:

```
; right overrides left
Global config file -> CLI flags -> code
```

What do y’all think?

GrayJack:
I think that would be the best option :smile:

jacereda:
IMHO CLI flags should have preference. Say I want to cross-compile or use tcc. I don't want to edit the code or config file. It would be better to just pass `-cc tcc`.

Having said that, maybe something like this would be quite flexible:

```
carp --eval-postload '(Project.config "compiler" "tcc")' -x whatever.carp
```

jacereda:
I mean, we could also have `--eval-preload` and you would have:
```
Config file -> preload code -> code -> postload code
```

scolsen:
I agree that CLI flags should have the highest precedence. pre and post eval is cool, but I think somewhat verbose/complicated for this particular feature—it feels more natural to me to pass a compiler choice as a a simple flag value—the pre-post-eval mechanism might be very useful generally speaking, though.

jacereda:
You're right, it feels a bit verbose, but maybe it's not so common to deserve a shorthand? You can always  create, say, a `fast-carp.sh` that does:

```
./carp.sh --eval-preload '(Project.config "compiler" "tcc")' "$@"
```

The part that worries me is that you might want to override in the command line any of the current or future Project settings and we could end up having tons of switches.

scolsen:
That’s a fair point! Sounds good—are we going to make it just a general mechanism that evals some arbitrary form, or are we going to restrict it to evaluating only `Project.config` forms?

PR #810, opened 2020-05-16
Fix Debug.trace for new evaluator

hellerve:
This PR fixes `Debug.trace` for the new evaluator. In the old evaluator `file`, `line`, and `column` were commands that evaluated their argument before it was passed in. Now they are primitives, and passing in a macro symbol will result in the information on the macro symbol being returned. We often do not want that, and that’s why we have to wrap it in an `eval` to make sure the symbol gets resolved first before passing it in.

Cheers


eriksvedang:
Phew!

PR #815, opened 2020-05-18
Make Array.range safe

### LINKED ISSUE #776: Array.range should guard against nonsensical input

hellerve:
This PR makes `Array.range` safe. I first wanted to make the error case return a `String`, but `String.copy` is not yet available in the `Array` module. It fixes #776 

Cheers

### DIFF
--- a/core/ArrayExt.carp
+++ b/core/ArrayExt.carp
@@ -24,8 +24,8 @@ the input given was wrong, containing an error message.")
             (set! e (+ e step))))
         (Result.Success x))))
 
-  (doc range-or-default "is an unsafe version of [`range](#range). It returns
-an empty array on failure.")
+  (doc range-or-default "is a version of [`range`](#range) that returns an empty
+array on failure.")
   (defn range-or-default [start end step]
     (Result.from-success (range start end step) []))
 )
--- a/core/ArrayExt.carp
+++ b/core/ArrayExt.carp
@@ -24,8 +24,8 @@ the input given was wrong, containing an error message.")
             (set! e (+ e step))))
         (Result.Success x))))
 
-  (doc unsafe-range "is an unsafe version of [`range](#range). It returns an
-empty array on failure.")
-  (defn unsafe-range [start end step]
+  (doc range-or-default "is an unsafe version of [`range](#range). It returns
+an empty array on failure.")
+  (defn range-or-default [start end step]
     (Result.from-success (range start end step) []))
 )
--- a/core/ArrayExt.carp
+++ b/core/ArrayExt.carp
@@ -1,0 +1,31 @@
+; we extend the array module, now that we have strings
+(defmodule Array
+  (doc range "creates an array from `start` to `end` with `step` between them
+(the elements must support `<`, `<=`, `>=`, and `to-int`).
+
+It returns a `Result.Success` if the input was right, and a `Result.Error` if
+the input given was wrong, containing an error message.")
+  (defn range [start end step]
+    (cond
+      (= step (zero))
+        (Result.Error @"`Array.range` cannot be called with step size `0`.")
+      (and (< start end) (< step (zero)))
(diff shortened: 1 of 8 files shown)


eriksvedang:
while the safety of this is obviously great, the usability seems to *really* take a hit. Now you both need to pull out the range and match on it and then understand some error codes. Doesn’t seem worth it to me... I kinda missed the discussion on this one in the chat though, so I might be missing some important things to consider.

hellerve:
There are two cases especially that can make your program crash, setting the `step` to `0` and going backwards when you should be going forwards.

A step size of `0` will lead to an infinite loop, going in the wrong direction will leave you with uninitialized memory. Both of those are unacceptable and dangerous IMO.

jacereda:
Why choose? I'd have `range` call `unsafe-range` after validation...

hellerve:
Right, I actually came to this conclusion too and mentioned it in the chat; I should’ve probably crossposted here. @eriksvedang would that make you less concerned about the UX of this PR?

eriksvedang:
Personally I'd be happy with just returning an empty array in those problematic cases (I totally agree that the current behaviour is really bad and confusing). But everyone else thinks the more type heavy approach is better, I'm fine with it. In that case maybe define the function in a new file later on so that it's possible to get String error messages?

hellerve:
I’m more or less fine with either, but I think just returning an empty array can lead to cascading failure cases that are potentially hard to debug for the users. We could define `unsafe-range` to be:

```
(defn unsafe-range [start end step]
   (Result.from-success (range start end step) []))
```

to get both behaviors: `unsafe-range` also doesn’t crash then, and `range` is the more “heavy-handed”, but safe approach. What do y’all think?

eriksvedang:
I think adding both safe/unsafe is a good solution then. But can we put it into something like `Array_ext.carp` to be able to get a string error message? Or maybe we could have a specific error sumtype with the different cases that can go wrong for `range`? You can always stringyfying that if you don't want to handle each case.

hellerve:
Sounds good. I’ll make the changes tomorrow :)

hellerve:
I updated this :)

GrayJack:
I think that is `unsafe-range` the way it is now if not unsafe.

I think would be better naming if  `range` returned empty array in fail cases and a `try-range` that returns a `Result`.

But since I think I'm long past the naming discussion, maybe the function that return empty array in fail cases should be called `range-or-default` or `range-default`, or any other thing other than `unsafe-range`, since it's not really unsafe.

eriksvedang:
@hellerve What's your take on @GrayJack's comment? I kinda agree that unsafe should be reserved for things that can crash... (which `unsafe-range` can't, right?)

Funny how such a innocent looking function can cause so much trouble :)

hellerve:
Yeah, I’d be happy to rename it; I guess `range-or-default` is the appropriate name?

eriksvedang:
OK, lets go with that.

core/ArrayExt.carp
```diff
@@ -0,0 +1,31 @@
...
+               op (if (< start end) <= >=)]
+        (while (op e end)
+          (do
+            (aset! &x i e)
+            (set! i (Int.inc i))
+            (set! e (+ e step))))
+        (Result.Success x))))
+
+  (doc range-or-default "is an unsafe version of [`range](#range). It returns
+an empty array on failure.")
```
GrayJack, line 28: Maybe a better doc string could be "is a versions of [`range`](#range) that defaults to an empty array on failure"?

hellerve: Good point! I updated the docstring :)

PR #816, opened 2020-05-18
Add missing longref implements annotations

hellerve:
This PR adds missing `implements` annotations to `LongRef`.

Cheers


scolsen:
looks good, we should probably add explicit tests for these impl functions too if that makes sense—I mostly relied on getting tests to pass for the original PR, which means none of these leftovers are presently tested.

PR #819, opened 2020-05-19
Interface Implementation improvements

scolsen:
This PR bundles a few small improvements to our handling of interfaces and implementations:

- [x] Updates `primitiveMeta` to handle qualified cases
  
   Before `meta` use to fail to find its argument binding in cases
   like this one:

   ```
   (defmodule Foo (hidden bar) (meta Foo.bar "hidden"))
   ```

   This is because `meta` used to cons path strings to the path even if
   the path was already qualified with a module name as in the example
   above. We now treat these cases separately, and instances like the
   above work.

- [x] Updates `primitiveImplements` to set the "implements" meta to a list of implemented interfaces, instead of a single interface. Duplicates aren't added.
  
   This makes sense since functions could implement multiple generic
   interfaces. An example of the new behavior:

   ```
   (implements inc foo)
   (implements dec foo)
   (meta foo "implements")
   => (dec inc)
   ```
  
- [x] Adds an `implements?` macro that returns a bool indicating whether or not a given binding implements a given interface:

  ```
  鲤 (implements inc foo)
  鲤 (implements dec foo)
  鲤 (eval (implements? 'inc foo))
  => true
  鲤 (eval (implements? 'dec foo))
  => true
  鲤 (eval (implements? '+ foo))
  => false 
  ```

I also fixed an erroneous `implements` declaration for `PairRef.>` that I happened to spot.

### DIFF
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -149,7 +149,7 @@ primitiveImplements xobj ctx [x@(XObj (Sym interface@(SymPath _ _) _) _ _), i@(X
                     in  return (ctx' {contextGlobalEnv = envInsertAt global (getPath defobj) (Binder adjustedMeta defobj)},
                                dynamicNil)
                   _ ->
-                    let impls = XObj (Lst [x]) (Just dummyInfo) (Just UnitTy)
+                    let impls = XObj (Lst [x]) (Just dummyInfo) (Just DynamicTy)
                         adjustedMeta = meta {getMeta = Map.insert "implements" impls (getMeta meta)}
                     in  return (ctx' {contextGlobalEnv = envInsertAt global (getPath defobj) (Binder adjustedMeta defobj)},
                                  dynamicNil)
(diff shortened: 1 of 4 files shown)


src/Primitives.hs
```diff
@@ -632,6 +645,19 @@ primitiveMeta (XObj _ i _) ctx [XObj (Sym path _) _ _, XObj (Str key) _ _] = do
       return (evalError ctx
                         ("`meta` failed, I can’t find `" ++ show path ++ "`")
                         i)
+primitiveMeta (XObj _ i _) ctx [XObj (Sym path _) _ _, XObj (Str key) _ _] = do
```
hellerve, line 648: This looks almost identical to the case above, can we do some code sharing here?

scolsen: Good call! Done!

scolsen:
Looks like the CI checks were cancelled for some reason? Any way we can trigger those? All the tests pass locally.

eriksvedang:
@scolsen Not sure why, I'll trigger them again! This looks ace!

eriksvedang:
Looks good to me @scolsen, thanks a ton! You managed to work around the problems I ran into when trying to solve this myself, nicely done :D

This will fix a whole bunch of other issues too, so that's awesome!

PR #821, opened 2020-05-21
Honor sigs when unifying types; do not constrain let bindings by sigs

### LINKED ISSUE #693: Unification in `sig` doesn’t handle indirections
Thank you for fixing some of the issues in sig unification @eriksvedang! #685 is now fixed, but it seems like indirection still isn’t handled correctly. Consider `Phantom`:

```clojure
deftype (Phantom a) [])

(defmodule Phantom
  (sig ensure (Fn [a (Phantom a)] a))
  (defn ensure [a p] a)
)

(defn m [a] (Phantom.ensure a (the (Phantom Int) (Phantom.init))))

(defn main [] (println* (m 123l)))
```

The `a` used in `ensure` should be the same everywhere, but it seems like it’s typed as `(Fn [a b] a)`.

Cheers

scolsen:
This PR improves the functionality of sigs in two ways:

- [x] Updates `sigFromDefnOrDef` to contextualize it's path lookups--before we were not covering the context-dependent case of an unqualified name within a module (e.g. `(defmodule Foo (sig bar....))`-- this resulted in any such signature being ignored, leading to confusing types. Now such signatures work.
- [x] Updates constraint generation to only apply sig constraints to root xobjs (aka the `def` or `defn` forms they apply to). Before, we applied sig constraints to interior let bindings as well as the top level definitions they were actually meant to apply to. Once sigs within modules were honored, it opened up a host of functions that were prey to this issue. We now check whether or not an xobj is indeed the root xobj (aka the sig constrained form) before generating a sig constraint.

I also fixed a typing issue for `Control.iterate-until` which had a `sig` and definition mismatch (which went under the radar before since the sig wasn't being used to constrain anything due to the first issue).

Fixes #693 (and possibly some others? I feel like a couple of people have mentioned sigs not constraining the types?)


scolsen:
Here's a small illustration of the intended behavior in #693 working:

```
;;phantom.carp
(deftype (Phantom a) [])                                                                                                                                                            
                                                                                                                                                                                    
(defmodule Phantom                                                                                                                                                                  
  (sig ensure (Fn [a (Phantom a)] a))                                                                                                                                               
  (defn ensure [a p] a)                                                                                                                                                             
)                                                                                                                                                                                   
                                                                                                                                                                                    
(defn m [a] (Phantom.ensure a (the (Phantom Int) (Phantom.init))))                                                                                                                  
                                                                                                                                                                                    
(defn main [] (println* (m 123)))  
--------------------------------------------------------
鲤 (load "phantom.carp")
鲤 :i ensure
Phantom.ensure : (Fn [a, (Phantom a)] a)
```

When the source is kept as is in #693 it leads to the expected error since the long value passed to `ensure` in main is not an Int.

scolsen:
hm, so two of the CIs built fine but it looks like the mac one was subject to weird/random cancellation again? https://github.com/carp-lang/Carp/actions/runs/111081152

at any rate all the tests pass on my macos excepting SDL which I don’t have installed and can’t check locally.

eriksvedang:
This should fix the issues you’ve experienced with sigs @timdeve ..?!

eriksvedang:
⭐️⭐️⭐️⭐️⭐️

PR #826, opened 2020-05-22
Add `s-exp` primitive and Introspection module

### LINKED ISSUE #560: Add Reflection Module Proposal
Inspired by @sdilts excellent proposal for resource handlers, I'd like to open a proposal for a `Reflection` module that (hopefully) achieves a similar level of fidelity.

## Why?

The purpose of the `Reflection` module is to give authors macro/dynamic/pre-compile time insight into the *properties of the forms*, in particular, of the forms that define the bindings in the current environment. One can check the arity of a function, whether a given binding is a macro, function, or dynamic function, whether a `def` binds a name to a static value or a form that requires evaluation, whether the name of a function satisfies certain criteria, and even details about a function's body.

This functionality is only useful at the macro level, when one might generate a series of new forms based on the contents and properties of some forms previously defined.

### Example Use Case

Let's say we want to create a generic version of the `curry` function. Our function should take a function, `f`, of any arity, figure out how many argument's it's been provided, and generate a requisite function to handle taking the remaining arguments required by `f`. With the `Reflection` module, we can achieve this using the proposed `arity` function:

(it's about 2am here, so I'm not paying too close attention to correctness, hope this gets the point across!)

```clojure
(defndyanmic gen-arg-names [x acc] 
  (if (> 0 x)
       (gen-arg-names (- x 1) (append acc (gensym x)))
       acc))

;; assume concat has the type: (Fn [(List List)] List)
(defndynamic curry-internal [f args remaining]
  (list 'fn (gen-arg-names remaining (array)) 
    (concat (list f) args (gen-arg-names remaining (list)))))

(defndynamic curry [f :rest args]
  (curry-internal f args (- (arity f) (length args)))

(defn foo [x y z] x)
(curry 'foo "hello")
=> (fn [a1 a2] (foo "hello" a1 a2))
(defn bar [x y z g w] x)
(curry 'bar "hello" "world")
=> (fn [a1 a2 a3] (foo "hello" "world" a1 a2 a3))
```

### Another Example Use Case

I'm working on a library now that attempts to provide a macro-based implementation of "typeclasses" for carp. This approach requires, it turns out, a good amount of introspection. One needs to take a look at various definitions in the environment and check their properties to ensure they meet certain criteria. For example, in my approach, one implements a typeclass using a `deftypeclass` macro. The macro, to really be robust, needs to verify:

- the number of type parameters of a type definition passed to the macro. 
- whether the user passed interfaces, and whether these interfaces either return instances of the typeclass or take it as arguments.

This requires inspecting the contents of definitions passed to the macro by the user. At present, the user needs to pass in the defining forms themselves for the macro to work--however, if the macro could call utility functions that facilitate the inspection of bindings, the user could simply pass bound names instead of definitions. 

This:

```clojure
(deftypeclass (Monad a) (definterface return (Fn [a] (Monad a))))
(Monad (deftype (Maybe a) (Just [a]) (Nothing [])))
```

becomes this:

```clojure
(definterface return (Fn [a] b))
(deftypeclass (Monad a) '(return join)) ;; assuming join is also defined somewhere
(Monad Maybe) ;; like most languages, the type declaration does not have to be specially 
;; passed to the macro, potentially restricting its use, but can instead be used flexibly,
;; with or without the "typeclass"
```

Which makes the macros easier to use and more adaptable, and prevents the interfaces from being "captured" by the macro, retaining maximal generality (in the first example, the interface is defined within the confines of the macro and only usable based on whatever the macro decides to do with the interface definition it's given, whereas, in the second instance, a highly generic interface is both available to whatever modules want to implement it, as well as the macro for defining a typeclass).

## Reflection Module

In order to support the use-case described in the section above, as well as a number of other macrological use cases, I propose adding `Reflection` *submodule* to the existing `Dynamic` module, that exposes the following functions, expressed as pseudo signatures:

```clojure
(defmodule Reflection
  ;; Returns true when the binding resolves to a `defn` form.
  (defn? (Fn [Binding] Bool))
  ;; Returns true when the binding resolves to a `defndynamic form`
  (dynamic? (Fn [Binding] Bool))
  ;; Returns true when the binding resolves to a `def` form
  (def? (Fn [Binding] Bool))
  ;; Returns true when the binding resolves to a `defmacro` form
  (macro? (Fn [Binding] Bool))
  ;; Returns the arity, number of arguments, of function bindings. 
  (arity (Fn [Binding] Int))
  ;; Returns true when the return value of a function or def form is a static value, 
  ;; false when it requires evaluation. 
  (static? (Fn [Binding] Bool))
  ;; Returns true if the function contains a possible recursive call, i.e. the binding name is 
  ;; referred to in the function body.
  (recursive? (Fn [Binding] Bool))
)
```

I'd also like to support the following functions, which at present, afaict, cannot be implemented using the current evaluator:

- `deftype?`: Can't be implemented at present. Unlike functions, types do not resolve to their type definitions. while, `'foo` resolves to, e.g. `(defn foo [x] x)` `'Bar` presently resolves to just `Bar`, even if there's a corresponding `(deftype Bar Nothing)`.
- `arity` on types: Because of the limitation described above (we can't retrieve the definition of a type), `arity` cannot tell us the number of type variables a given type takes, which could be useful information.
- `enum?`: Returns true if a given type is an enumeration, i.e. `(deftype Bar A B C)`Again, because of the same limitation around types, currently impossible to implement in pure carp.

### Special Commands

Other functions that implement similar "reflective" behavior are already provided as special commands, for example the `members` function on structs and sumtypes. This suggests that perhaps some of the functions proposed here, if deemed useful, should similarly be implemented as commands. 

I motion against this simply because Carp already provides us the ability to gather this reflective information in Carp itself, so, rather than bloat the compiler with additional special functionality, we can simply implement the functions as a carp module. To me, this feels cleaner and easier to manage and it might help keep the compiler complexity lower.

### Third Party Library

Perhaps these functions don't belong in core carp at all, but should be implemented as a third party library, maintained separately and adopted only by those who want to use it. This approach  has the benefit of keeping the core language libs small and simple, and there's nothing in particular that recommends against it aside from ubiquity/availability of the utilities.

Thanks for reading! Let me know if you have any thoughts. I'm happy to implement this as a separate library, but figured I'd open a discussion to see if there's any interest for this sort of thing in core.

scolsen:
This PR adds an `s-exp` primitive, which returns the s-expression or form associated with a binding. It's similar to the default behavior of dynamic xobjs, but it does some extra checking on receiving module xobjs to return the original deftype forms whenever the modules are generated from a type.

`s-exp` is the backbone of a new `Introspection` module, which contains predicates about bindings and their associated forms. One can use this module in a dynamic context to determine if a binding is a function, struct, sumtype, interface, it's arity, etc.

Here are a few examples of `s-exp`:

```clojure
鲤 (def foo 3)
鲤 (s-exp foo)
=> (def foo 3)
鲤 (defn bar [x] x)
鲤 (s-exp bar)
=> (defn bar [x] x)
鲤 (deftype Foo [x Int])
鲤 (s-exp Foo)
=> (deftype Foo [x Int])
鲤 (deftype (Bar a) (Baz [a])) 
鲤 (s-exp Bar)
=> (deftype Bar (Baz [a]))
鲤 (definterface carp (Fn [a] a))
鲤 (s-exp carp)
=> (definterface carp)
鲤 
```

and the `Introspect` module:

```clojure
鲤 (Introspect.function? bar)
=> true
鲤 (Introspect.arity bar)
=> 1
鲤 (Introspect.function? foo)
=> false
鲤 (Introspect.variable? foo)
=> true
鲤 (Introspect.type? Foo)
=> true
鲤 (Introspect.struct? Foo)
=> true
鲤 (Introspect.arity Foo)
=> 1
鲤 (Introspect.sumtype? Foo)
=> false
鲤 (Introspect.sumtype? Bar)
=> true
鲤 (Introspect.arity Bar)
=> (1)
鲤 (deftype (Qux a) (Goo [a]) (Loo [a Int]))
鲤 (Introspect.arity Qux)
=> (1 2)
```

Let's go ahead and say this fixes #560?

### DIFF
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -725,7 +725,7 @@ primitiveSexpression (XObj _ i _) ctx [XObj (Sym path _) _ _] =
                return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
            Just (_, Binder _ xobj'') ->
              return (ctx, Right xobj'')
-           Nothing -> return (ctx, dynamicNil)
+           Nothing -> return (ctx, Right (XObj (Lst []) (Just dummyInfo) (Just DynamicTy)))
 primitiveSexpression (XObj _ i _) ctx xobj =
   return $ evalError ctx ("s-exp expected a symbol argument but got: " ++ unwords (map pretty xobj)) i
 
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -698,7 +698,7 @@ primitiveDeftemplate _ ctx [x, _, _, _] =
 
 primitiveSexpression :: Primitive
 primitiveSexpression (XObj _ i _) ctx [XObj (Sym path _) _ _] =
-  let env = contextEnv ctx
+  let env = contextGlobalEnv ctx
       tyEnv = getTypeEnv $ contextTypeEnv ctx
   in case lookupInEnv path env of
        Just (_, binder@(Binder _ xobj)) ->
@@ -710,7 +710,7 @@ primitiveSexpression (XObj _ i _) ctx [XObj (Sym path _) _ _] =
              Just (_, Binder _ (XObj (Lst forms) i t)) ->
                return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
              Just (_, Binder _ xobj') ->
-               return (ctx, Right xobj')
+               trace ("not a list") return (ctx, Right xobj')
              -- Okay, this is just a `defmodule` not a type.
              Nothing ->
                return (ctx, Right mod)
@@ -718,11 +718,16 @@ primitiveSexpression (XObj _ i _) ctx [XObj (Sym path _) _ _] =
            return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
          _ -> return (ctx, Right xobj)
        Nothing ->
-         -- Just to be sure, check the type env--this might be an interface
+         -- Just to be sure, check the type env--this might be an interface or
+         -- type
          case lookupInEnv path tyEnv of
+           Just (_, Binder _ (XObj (Lst forms) i t)) ->
+               return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
            Just (_, Binder _ xobj'') ->
              return (ctx, Right xobj'')
-           Nothing -> return (ctx, dynamicNil)
+           Nothing -> trace ("Nothing found: " ++ show env ++ show path) return (ctx, dynamicNil)
+primitiveSexpression (XObj _ i _) ctx xobj =
+  return $ evalError ctx ("s-exp expected a symbol argument but got: " ++ show xobj) i
 
 toSymbols :: XObj -> XObj
 toSymbols (XObj (Defn _) i t) = (XObj (Sym (SymPath [] "defn") Symbol) i t)
@@ -730,4 +735,5 @@ toSymbols (XObj Def i t) = (XObj (Sym (SymPath [] "def") Symbol) i t)
 toSymbols (XObj (Deftype _) i t) = (XObj (Sym (SymPath [] "deftype") Symbol) i t)
 toSymbols (XObj (DefSumtype _) i t) = (XObj (Sym (SymPath [] "deftype") Symbol) i t)
 toSymbols (XObj (Interface _ _) i t) = (XObj (Sym (SymPath [] "definterface") Symbol) i t)
+toSymbols (XObj Macro i t) = (XObj (Sym (SymPath [] "defmacro") Symbol) i t)
 toSymbols x = x
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -710,7 +710,7 @@ primitiveSexpression (XObj _ i _) ctx [XObj (Sym path _) _ _] =
              Just (_, Binder _ (XObj (Lst forms) i t)) ->
                return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
              Just (_, Binder _ xobj') ->
-               trace ("not a list") return (ctx, Right xobj')
+               return (ctx, Right xobj')
              -- Okay, this is just a `defmodule` not a type.
              Nothing ->
                return (ctx, Right mod)
@@ -725,7 +725,7 @@ primitiveSexpression (XObj _ i _) ctx [XObj (Sym path _) _ _] =
                return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
            Just (_, Binder _ xobj'') ->
              return (ctx, Right xobj'')
-           Nothing -> trace ("Nothing found: " ++ show env ++ show path) return (ctx, dynamicNil)
+           Nothing -> return (ctx, dynamicNil)
 primitiveSexpression (XObj _ i _) ctx xobj =
   return $ evalError ctx ("s-exp expected a symbol argument but got: " ++ show xobj) i
 
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -712,11 +712,20 @@ primitiveSexpression (XObj _ i _) ctx [XObj (Sym path _) _ _] =
              -- Okay, this is just a `defmodule` not a type.
              Nothing ->
                return (ctx, Right mod)
-         _ -> trace (show binder) return (ctx, Right xobj)
+         (XObj (Lst forms) i t) ->
+           return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
+         _ -> return (ctx, Right xobj)
        Nothing ->
-         -- Just to be sure, check the type env--this might be an interface or
+         -- Just to be sure, check the type env--this might be an interface
          case lookupInEnv path tyEnv of
            Just (_, Binder _ xobj'') ->
              return (ctx, Right xobj'')
            Nothing -> return (ctx, dynamicNil)
 
+toSymbols :: XObj -> XObj
+toSymbols (XObj (Defn _) i t) = (XObj (Sym (SymPath [] "defn") Symbol) i t)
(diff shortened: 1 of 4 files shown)


src/Primitives.hs
```diff
@@ -695,3 +695,45 @@ primitiveDeftemplate _ ctx [s@(XObj (Sym (SymPath _ _) _) _ _), _, _, _] = do
...
+         -- Just to be sure, check the type env--this might be an interface or
+         -- type
+         case lookupInEnv path tyEnv of
+           Just (_, Binder _ (XObj (Lst forms) i t)) ->
+               return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
+           Just (_, Binder _ xobj'') ->
+             return (ctx, Right xobj'')
+           Nothing -> return (ctx, dynamicNil)
+primitiveSexpression (XObj _ i _) ctx xobj =
+  return $ evalError ctx ("s-exp expected a symbol argument but got: " ++ show xobj) i
```
hellerve, line 730: Instead of `show` we should probably use `pretty` here to get a nicer representation.

src/Primitives.hs
```diff
@@ -695,3 +695,45 @@ primitiveDeftemplate _ ctx [s@(XObj (Sym (SymPath _ _) _) _ _), _, _, _] = do
...
+         _ -> return (ctx, Right xobj)
+       Nothing ->
+         -- Just to be sure, check the type env--this might be an interface or
+         -- type
+         case lookupInEnv path tyEnv of
+           Just (_, Binder _ (XObj (Lst forms) i t)) ->
+               return (ctx, Right (XObj (Lst (map toSymbols forms)) i t))
+           Just (_, Binder _ xobj'') ->
+             return (ctx, Right xobj'')
+           Nothing -> return (ctx, dynamicNil)
```
hellerve, line 728: Should we return `()` or an error here?

scolsen: I like returning `()` instead of an error since it's quite possible a binding isn't defined and feels like a smoother experience to me to handle that case with an identity value for the purposes of these functions--that way users can decide whether and how to error out when some binding isn't defined.

`dynamicNil` is equivalent to `()` but doesn't actually emit it--how would we get the emission? Do we just return an empty `Lst` xobj here instead?

src/Primitives.hs
scolsen: Ah, good thinking! Thanks!

src/Primitives.hs
scolsen: Yep, `Lst` does it. Thanks!

scolsen:
> This looks really great, amazing work!
> 
> Maybe we should point out somewhere what happens on name conflicts, e.g. when a thing is both a type and a module. Which binding wins? I see below that this is special-cased and we get the type, so we should probably tell people about that.
> 
> And how would we go about getting the module, if we _really_ wanted to? This is not necessarily a breaker for this PR, but IMO there should definitely be a way. Maybe we could separate type and value handling and push that logic into Carp?

Great points as always. I think it'd be really valuable to be able to get either the type or module of a binding. I'm not certain, but that functionality may still need to remain in the compiler since it requires peeking into two different environments. 

If you don't mind, I'll leave this out of this PR for the sake of simplicity, but it's definitely worth pursuing, so I'll add an issue for it!

sdilts:
I like it! Instead of `s-exp` it might be more ergonomic to name it something like `get-sig` or something similar, as it seems like the getter version of `sig`. To me, the name feels kind of like `stoi`, or `itoa` in C: the name makes sense, but it's hard to read or remember if you don't use it all the time.

eriksvedang:
Looks great to me! I kinda like the `s-exp` name but we can discuss it a bit more before merge:ing, perhaps there's something better.

scolsen:
I also like `s-exp`, but I'm open to alternatives! @sdilts I like what `get-sig` suggests, but I think it's too narrow for this case. Can you elaborate your line of thinking around `stoi` and `itoa` a bit more? That also feels somewhat narrow to me, since we're working with forms not strings, but maybe you're thinking of it in a more general sense that I'm missing?

sdilts:
With the naming thing, I don't like abbreviations unless they are used quite often and are pretty common. In most cases, it makes it hard for me to remember the correct names, and I think that `stoi` and `itoa`are particularly egregious examples. The name should at least should use `expr` instead of `exp`, as `exp` is usually an abbreviation for an [exponent function](http://www.cplusplus.com/reference/cmath/exp/). 

Looking over what the function does again, I agree that `get-sig` is too narrow, My bad eyes missed that it returns whole function definitions and not just the signatures. The name makes much more sense with that in mind.

scolsen:
> With the naming thing, I don't like abbreviations unless they are used quite often and are pretty common. In most cases, it makes it hard for me to remember the correct names, and I think that `stoi` and `itoa`are particularly egregious examples. The name should at least should use `expr` instead of `exp`, as `exp` is usually an abbreviation for an [exponent function](http://www.cplusplus.com/reference/cmath/exp/).

Good point! I agree that full names are better, unless anyone objects, I'm fine with calling it `s-expression`; otherwise we can use the clearer abbreviation that you suggested `s-expr`

eriksvedang:
I agree that `expr` is better to avoid mixing it up with `exp`. Maybe `expr` is an OK name actually? It's definitely easier to type...

scolsen:
Sounds good! Made the name change...but now I'm weirdly getting an error in `InitialTypes.hs`

scolsen:
Weirdly the error doesn't happen when the form is named `s-exp` but it happens when it's named `expr`....

scolsen:
Ah, looks like it has to do with the fact that some macro bindings are named `expr` so it doesn't work due to  #659

is `s-expr` ok for now then?

sdilts:
Sounds good to me!

eriksvedang:
Haha oops, that’s weird. Works for me.

PR #827, opened 2020-05-22
Command line switches to evaluate code

jacereda:
This is a more flexible alternative to #803. Implements `--eval-preload` and `--eval-postload` with the intention of changing compiler and flags, but could be useful for other purposes.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -86,7 +86,7 @@ main = do setLocaleEncoding utf8
                                   , projectGenerateOnly = generateOnly
                                   , projectCarpDir = fromMaybe (projectCarpDir p) $ lookup "CARP_DIR" sysEnv
                                   , projectCompiler = fromMaybe (projectCompiler p) compiler
-                                  , projectPrompt = prompt
+                                  , projectPrompt = fromMaybe (projectPrompt p) prompt
                                   }
               project = applySettings defaultProject
               noArray = False
@@ -158,7 +158,7 @@ data OtherOptions = OtherOptions
   , otherLogMemory :: Bool
   , otherOptimize :: Bool
   , otherGenerateOnly :: Bool
-  , otherPrompt :: String
+  , otherPrompt :: Maybe String
   } deriving Show
 
 parseOther :: Parser OtherOptions
@@ -168,7 +168,7 @@ parseOther = OtherOptions
   <*> switch (long "log-memory" <> help "Log memory allocations")
   <*> switch (long "optimize" <> help "Optimized build")
   <*> switch (long "generate-only" <> help "Stop after generating the C code")
-  <*> strOption (long "prompt" <> help "Set REPL prompt")
+  <*> optional (strOption (long "prompt" <> help "Set REPL prompt"))
 
 parseExecMode :: Parser ExecutionMode
 parseExecMode =
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -59,26 +59,30 @@ main = do setLocaleEncoding utf8
               noProfile = NoProfile `elem` otherOptions
               optimize = Optimize `elem` otherOptions
               generateOnly = GenerateOnly `elem` otherOptions
-              projectWithFiles = defaultProject { projectCFlags = ["-D LOG_MEMORY" | logMemory] ++
-                                                                  ["-O3 -D NDEBUG" | optimize] ++
-                                                                  projectCFlags defaultProject,
-                                                  projectCore = not noCore,
-                                                  projectGenerateOnly = generateOnly}
+              compileFast = CompileFast `elem` otherOptions
+              flagsSettings p = p { projectCFlags = ["-D LOG_MEMORY" | logMemory] ++
+                                                    ["-O3 -D NDEBUG" | optimize] ++
+                                                    projectCFlags p
+                                  , projectCore = not noCore
+                                  , projectGenerateOnly = generateOnly
+                                  , projectCarpDir = case lookup "CARP_DIR" sysEnv of
+                                      Just carpDir -> carpDir
+                                      Nothing -> projectCarpDir p
+                                  , projectCompiler = if compileFast then "tcc -lm" else projectCompiler p
+                                  }
+              applySettings = flagsSettings . setCustomPromptFromOptions otherOptions
+              project = applySettings defaultProject
               noArray = False
-              coreModulesToLoad = if noCore then [] else coreModules (projectCarpDir projectWithCarpDir)
-              projectWithCarpDir = case lookup "CARP_DIR" sysEnv of
-                                     Just carpDir -> projectWithFiles { projectCarpDir = carpDir }
-                                     Nothing -> projectWithFiles
-              projectWithCustomPrompt = setCustomPromptFromOptions projectWithCarpDir otherOptions
               startingContext = Context
-                                 (startingGlobalEnv noArray)
-                                 Nothing
-                                 (TypeEnv startingTypeEnv)
-                                 []
-                                 projectWithCustomPrompt
-                                 ""
-                                 execMode
-                                 []
+                                (startingGlobalEnv noArray)
+                                Nothing
+                                (TypeEnv startingTypeEnv)
+                                []
+                                project
+                                ""
+                                execMode
+                                []
+              coreModulesToLoad = if noCore then [] else coreModules (projectCarpDir project)
           context <- loadFilesOnce startingContext coreModulesToLoad
           carpProfile <- configPath "profile.carp"
           hasProfile <- doesFileExist carpProfile
@@ -110,6 +114,7 @@ data OtherOptions = NoCore
                   | LogMemory
                   | Optimize
                   | GenerateOnly
+                  | CompileFast
                   | SetPrompt String
                   deriving (Show, Eq)
 
@@ -130,6 +135,7 @@ parseArgs args = parseArgsInternal [] Repl [] args
             "--log-memory" -> parseArgsInternal filesToLoad execMode (LogMemory : otherOptions) restArgs
             "--optimize" -> parseArgsInternal filesToLoad execMode (Optimize : otherOptions) restArgs
             "--generate-only" -> parseArgsInternal filesToLoad execMode (GenerateOnly : otherOptions) restArgs
+            "--compile-fast" -> parseArgsInternal filesToLoad execMode (CompileFast : otherOptions) restArgs
             "--prompt" -> case restArgs of
                              newPrompt : restRestArgs ->
                                parseArgsInternal filesToLoad execMode (SetPrompt newPrompt : otherOptions) restRestArgs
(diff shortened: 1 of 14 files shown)


jacereda:
Well, it does lots of other things besides those stated in the description, but all of them are related to replacing `clang` as the compiler.

src/Emit.hs
```diff
@@ -701,10 +701,11 @@ memberToDecl indent (memberName, memberType) =
 
 defStructToDeclaration :: Ty -> SymPath -> [XObj] -> String
 defStructToDeclaration structTy@(StructTy typeName typeVariables) path rest =
-  let indent' = indentAmount
+  let indent = indentAmount
 
       typedefCaseToMemberDecl :: XObj -> State EmitterState [()]
-      typedefCaseToMemberDecl (XObj (Arr members) _ _) = mapM (memberToDecl indent') (pairwise members)
+      typedefCaseToMemberDecl (XObj (Arr []) _ _) = sequence $ pure $ appendToSrc (addIndent indent ++ "char __dummy;\n")
```
eriksvedang, line 707: What is this fix about?

src/Eval.hs
```diff
@@ -813,18 +813,21 @@ printC xobj =
...
-            , case ty xobj of
-                Just UnitTy -> xobj
-                Just (RefTy _ _) -> XObj (Lst [XObj (Sym (SymPath [] "println*") Symbol) (Just dummyInfo) Nothing, xobj])
-                                       (Just dummyInfo) (Just UnitTy)
-                Just _ -> XObj (Lst [XObj (Sym (SymPath [] "println*") Symbol) (Just dummyInfo) Nothing,
-                                     XObj (Lst [XObj Ref (Just dummyInfo) Nothing, xobj])
-                                           (Just dummyInfo) (Just UnitTy)])
-                               (Just dummyInfo) (Just UnitTy)
-            ]) (Just dummyInfo) (Just (FuncTy [] UnitTy StaticLifetimeTy))
+  XObj (Lst [ XObj (Defn Nothing) di Nothing
```
eriksvedang, line 816: What is this fix about?

test/init_global.carp
```diff
@@ -6,4 +6,6 @@
 
 ;; The one allocation left after 'carp_init_globals' should be 'g' itself:
 (defn main []
-  (assert (= 1l (Debug.memory-balance))))
+  (do
```
eriksvedang, line 9: Is there a reason you want to explicitly return 0 here?

app/Main.hs
```diff
@@ -38,109 +48,120 @@ defaultProject =
...
 
-setCustomPromptFromOptions :: Project -> [OtherOptions] -> Project
-setCustomPromptFromOptions project (o:os) =
-  case o of
-    SetPrompt newPrompt -> setCustomPromptFromOptions (project { projectPrompt = newPrompt }) os
-    _ -> setCustomPromptFromOptions project os
-setCustomPromptFromOptions project _ =
-  project
+parseFiles :: Parser [FilePath]
+parseFiles = some (argument str (metavar "FILES...")) <|> pure []
```
eriksvedang, line 167: I don't understand this piece of the code, what is the metavar thingy?

examples/basics.carp
```diff
@@ -234,4 +234,5 @@
       (two-lengths-in-same-func)
       (changing-target-of-ref)
       (resolve-correctly)
+      0
```
eriksvedang, line 237: Is there a reason you want to explicitly return 0 here?

stack.yaml
```diff
@@ -15,7 +15,7 @@
 # resolver:
 #  name: custom-snapshot
 #  location: "./custom-snapshot.yaml"
-resolver: lts-14.9
+resolver: lts-15.3
```
eriksvedang, line 18: Is this because of the optparse lib? I get a warning from stack:

```
Stack has not been tested with GHC versions above 8.6, and using 8.8.2, this may fail
```

src/Emit.hs
jacereda: Empty structs are illegal in C. There's at least one, the `(deftype Simple [])`
 in `examples/basics.carp`.

src/Eval.hs
jacereda: If you try at the REPL something like `(IO.println "42")` with `tcc` it will fail. If you don't set explicitly the return value to 0, the return value of that expressions will just be whatever the `rax` register contains (in this case, IIRC it was the length of the printed string). This wraps the expression in `(do xxx 0)`.

test/init_global.carp
jacereda: Same reason, not setting the return value explicitly to 0 will return non-zero from the executable and the test suite will fail.

app/Main.hs
jacereda: The metavar is used in `--help`. If you don't set it, it will say `ARG` instead IIRC.

stack.yaml
jacereda: Same here, we both have an old `stack` executable most probably.

jacereda: I bumped because `optparse` introduced a dependency that failed with old `ansi-terminal` IIRC.

app/Main.hs
jacereda: The `<|> pure []` is to allow an empty argument list. Maybe there's a better way. You could use `optional` instead to return a `Maybe [String]` instead. On a second thought, this should be unified with `optPreload/Postload`, let me change it.

jacereda: `many` was the right answer. I tried it at some point and it didn't work (there was probably some ambiguity or an error somewhere else). It seems to work correctly now.

stack.yaml
eriksvedang: OK, I'll try to upgrade my Stack.

eriksvedang: That worked, so I think we're good.

eriksvedang:
This is nice indeed:

```bash
$ carp --help                                                        
Usage: carp [--check | -b | -x | -i ARG] [--no-core] [--no-profile] 
            [--log-memory] [--optimize] [--generate-only] [--prompt ARG] 
            [--eval-preload CODE] [--eval-postload CODE] [FILES...]

Available options:
  --check                  Check project
  -b                       Build project
  -x                       Build an run project
  -i ARG                   Install built product
  --no-core                Don't load Core.carp
  --no-profile             Don't load profile.carp
  --log-memory             Log memory allocations
  --optimize               Optimized build
  --generate-only          Stop after generating the C code
  --prompt ARG             Set REPL prompt
  --eval-preload CODE      Eval CODE after loading config and before FILES
  --eval-postload CODE     Eval CODE after loading FILES
  -h,--help                Show this help text
```

app/Main.hs
```diff
@@ -38,109 +48,120 @@ defaultProject =
...
+                                Windows -> "clang-cl.exe"
+                                _ ->       "clang"
           , projectCore = True
           , projectEchoCompilationCommand = False
           , projectCanExecute = False
           , projectFilePathPrintLength = FullPath
           , projectGenerateOnly = False
           , projectForceReload = False
           }
+  where win = platform == Windows
```
scolsen, line 60: Looks like this local binding isn't used?

examples/basics.carp
scolsen: Also curious about this

src/Emit.hs
scolsen: lgtm--mind adding a comment that explains this? Could see it being an edge case we forget about ~N years down the line.

src/Eval.hs
scolsen: Even though it shouldn't have any adverse effects for other compilers/systems, I'm a bit wary of baking in a solution for the peculiarities of one particular system right into the compiler. Even though it doesn't have a behavioral cost, it does increase the complexity of the compiler code slightly and since it's such a specific case we're addressing, I could easily envision our knowledge of why this was added being lost in the far future.

Would it be possible to solve this in Carp instead, using, e.g. a library?

I think it's OK to do this if there's really no other solution. I'd add a brief explanatory comment just so we don't forget why this is here and accidentally delete it in some code cleanup in the future.

jacereda:
> I mentioned this in some comments, really the only changes here that I'm reluctant about are the ones specific to `tcc` and returning `0`s.

Do you mean the `#ifdef __TINYC__` bits? Right, it should probably detect some feature macro identifying the `__builtin_x_overflow()` availability, it's just I don't know how to identify it...

Those `0`s aren't tcc-specific, AFAICS any C compiler could end up returning an error code and we shouldn't rely on undefined behaviour.

> If possible, I really think this should be solved as a library instead--I think that's a better option than introducing code into the compiler for a very specific use case.
> 
> For example, for the `println*` case, you could define a `TCC.println*` that does exactly what you do here, wrap the call in a `do` that returns `0`.

But what about, say, `cl`? Or `zig cc`? We should only emit standard C code.

app/Main.hs
jacereda: Oh, right, thanks.

jacereda:
OK, removed the tcc-specific `__builtin_x_overflow()` detection.

scolsen:
> > I mentioned this in some comments, really the only changes here that I'm reluctant about are the ones specific to `tcc` and returning `0`s.
> 
> Do you mean the `#ifdef __TINYC__` bits? Right, it should probably detect some feature macro identifying the `__builtin_x_overflow()` availability, it's just I don't know how to identify it...
> 
> Those `0`s aren't tcc-specific, AFAICS any C compiler could end up returning an error code and we shouldn't rely on undefined behaviour.
> 
> > If possible, I really think this should be solved as a library instead--I think that's a better option than introducing code into the compiler for a very specific use case.
> > For example, for the `println*` case, you could define a `TCC.println*` that does exactly what you do here, wrap the call in a `do` that returns `0`.
> 
> But what about, say, `cl`? Or `zig cc`? We should only emit standard C code.

Gotcha, thanks for the additional explanation! I suppose I'm just curious as to why everything has worked ok for people up to this point, I suppose it's the opposite scenario in which compiler quirks were actually saving us from erroneous code--you're right that we shouldn't produce anything that's invalid. So this sounds good!

I'll take another looksee

jacereda:
In fact I'm wondering if we should have a CI job compiling with `tcc` so we can detect more undefined behaviour.

eriksvedang:
> In fact I'm wondering if we should have a CI job compiling with `tcc` so we can detect more undefined behaviour.

That sounds like a good idea if Github has support for it (I'm very uninformed about CI stuff).

jacereda:
> > In fact I'm wondering if we should have a CI job compiling with `tcc` so we can detect more undefined behaviour.
> 
> That sounds like a good idea if Github has support for it (I'm very uninformed about CI stuff).

I can try to add a job in a separate PR.

hellerve:
Does this mean we should also change the compiler in general not to accept `void main` as a possibility and to always be `int main`?

jacereda:
I'd say that's the way to go.

hellerve:
Oh, I see. We alsways emit `int main`, even if the body returns `void`.  That’s not good.

I also looked at the C standard again, and `void main` is, while used often, not part of the standard, which I didn’t know. We should definitely get rid of that, then. Or, if we want to keep those semantics, we should add an implicit `return 0` to void `main` bodies.

eriksvedang:
It's finally in! Thanks for the hard work @jacereda !

PR #831, opened 2020-05-23
Info Refactor

scolsen:
This PR moves `Info` and related functions into its own module in `Info.hs`.

`Obj.hs` is quite large at this point and contains functions that have a wide variety of responsibilities. I think it's worthwhile to try and separate concerns a bit more and simplify the `Obj.hs` code. Unfortunately, there are mutually recursive dependencies between most of the key types in `Obj` which makes it difficult to move them into isolated modules. `Info` however is less problematic, and it's straightforward to port it into its own module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ dist/
 main.obj
 examples/out
 /stack.yaml.lock
+tags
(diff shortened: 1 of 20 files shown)


eriksvedang:
Really nice!!!

PR #833, opened 2020-05-24
Autogenerate Tuple types and document

### LINKED ISSUE #830: The `Pair` module is undocumented.

hellerve:
This PR fixes #830 by autogenerating tuple modules (`Pair`, `Triple`, and `Quadruple`, for now).

It also adds `Dynamic.flatten`, which flattens a list.

I also added the tuple modules to documentation generation.

Cheers

### DIFF
--- a/core/Tuples.carp
+++ b/core/Tuples.carp
@@ -1,43 +1,95 @@
-(deftype (Pair a b) [a a b b])
+(defmodule Dynamic
+  (private deftuple-type-)
+  (hidden deftuple-type-)
+  (defndynamic deftuple-type- [name props]
+    (list 'deftype (cons name props)
+      (collect-into (flatten (map (fn [x] (list x x)) props)) array)))
 
-(defmodule PairRef
-  (defn = [p1 p2]
-    (and (= (Pair.a p1) (Pair.a p2))
-         (= (Pair.b p1) (Pair.b p2))))
-  (implements = PairRef.=)
+  (private deftuple-lt-)
+  (hidden deftuple-lt-)
+  (defndynamic deftuple-lt- [name props]
+    (if (empty? props)
+      'false
+      (let [fst (Symbol.prefix name (car props))]
+        (if (= (length props) 1)
+          (list '< (list fst 't1) (list fst 't2))
+          (list 'if (list '= (list fst 't1) (list fst 't2))
+            (deftuple-lt- name (cdr props))
+            (list '< (list fst 't1) (list fst 't2)))))))
 
-  (defn < [p1 p2]
-    (if (= (Pair.a p1) (Pair.a p2))
-      (< (Pair.b p1) (Pair.b p2))
-      (< (Pair.a p1) (Pair.a p2))))
-  (implements < PairRef.<)
+  ; this is basically just a giant template
+  (private deftuple-module-)
+  (hidden deftuple-module-)
+  (defndynamic deftuple-module- [name props]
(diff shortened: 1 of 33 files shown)


hellerve:
Because so many docs change (the navigation is changed on every core doc file) the change looks huge. The most important change is in `core/Tuples.carp`, though.

eriksvedang:
Very clever, this looks great!

core/Tuples.carp
```diff
@@ -1,43 +1,95 @@
...
+          (list 'implements '> (Symbol.prefix name '>))
+
+          (list 'doc 'reverse
+            (String.concat ["reverses a `" sname "` by reversing its member positions."]))
+          (list 'defn 'reverse ['t]
+            (cons 'init (map (fn [x] (list 'copy (list x 't))) (reverse props))))
+
+          (list 'defn 'zero [] (cons 'init (map (fn [_] '(zero)) props))))
+          (list 'implements 'zero (Symbol.prefix name 'zero))
+          ;(list 'doc 'zero
```
hellerve, line 70: For some reason, removing the comment here makes the compiler loop. The `doc` commands above don’t do that, so I’m not quite sure what’s happening there. Is it because it’s an `interface`?

scolsen: I suspect it has to do with the logic in `MetaSet`--the lookup is a bit messy atm

```haskell
let env = contextGlobalEnv ctx
      pathStrings = contextPath ctx
      fppl = projectFilePathPrintLength (contextProj ctx)
      fullPath = consPath pathStrings path
  case lookupInEnv fullPath env of
    Just (foundEnv, binder@(Binder _ xobj)) ->
      -- | Set meta on existing binder
      setMetaOn ctx (Just foundEnv) binder
```

Normally, consing the path should make it resolve to the right module-specific impl of `zero`, but there are issues with submodules and contextstack popping whereby for some reason MetaSet only gets the end of the stack. My guess is that the issue here might be similar. Either that or probably the local `setMetaOn` causes an infinite loop via it's different approach to resolving the pathname:

```
 setMetaOn :: Context -> Maybe Env -> Binder -> IO (Context, Either EvalError XObj)
      setMetaOn ctx foundEnv binder@(Binder metaData xobj) =
        do let globalEnv = contextGlobalEnv ctx
               newMetaData = MetaData (Map.insert key value (getMeta metaData))
               xobjPath = getPath xobj
               prefixPath = fromMaybe [] (fmap pathToEnv foundEnv)
               fullPath = case xobjPath of
                            SymPath [] _ -> consPath prefixPath xobjPath
                            SymPath _ _ -> xobjPath
```

Probably we should only resolve the path in one place, and not both the main and local `setMetaOn` bindings.

Just a hunch, not sure if this is really the root cause yet. It's also possible simply looking in the Global Env causes it to find the interface first and there's some problems with setting docs on that (shouldn't happen if we reconstruct the proper fullpath, but again, there are submodule related issues to getting the right fullpath in MetaSet, so I wouldn't be surprised if the same problems sometimes occur in edge cases for non-submodules.)

Presumably, we could resolve this doing `doc module-name.zero` -- but forward declarations don't support qualified names yet.

eriksvedang: Yeah your hunch seems spot on, maybe it's something we can fix by doing the meta data cleanup we've discussed earlier..?

scolsen: Yep, sounds good! I'll hopefully get around to Meta cleanup sometime this week.

PR #837, opened 2020-05-24
update with cond

### LINKED ISSUE #829: The `cond` keyword is undocumented.
There is no documentation on how to use `cond`.

mitchell-frost:
I added the documentation for cond which was not added in the LanguageGuide.md. #
Fixes #829

### DIFF
--- a/docs/LanguageGuide.md
+++ b/docs/LanguageGuide.md
@@ -70,32 +70,11 @@ foo ; symbol
 ```
 
 ### Conditional statements with 'cond'
-```(doc cond "this is the documentation for cond")```
 The ```cond``` statement executes a block of code if a specified condition is true. If the condition is false, another block of code can be executed. 
 
 ```(doc cond "this is the documentation for cond")```
 
-Examples are given below:
-
-```clojure
-(defndynamic cond-internal [xs]
-  (if (= (length xs) 0)
-    (list)
-    (if (= (length xs) 2)
-      (macro-error "cond has even number of branches; add an else branch")
-      (if (= (length xs) 1)
-        (car xs)
-        (list
-         'if
-         (car xs)
-         (cadr xs)
-         (cond-internal (cddr xs)))))))
-
-(defmacro cond [:rest xs]
-  (cond-internal xs))
-
-```
-Sample Example:
+Usage:
 
 ```clojure
 (cond
@@ -104,21 +83,13 @@ Sample Example:
           (<code_3>) ;; code_3 gets executed if condition_1 and condition_2 are false
 ``` 
 
-Example 1. Printing a statement depending on whether it is < or > 10:
+Here's an example about printing a statement depending on whether it is < or > 10:
 
 ```clojure
 (cond
           (< 10 1) (println "Don't print!")
           (> 10 1) (println msg)
           (println "Don't print!"))
-``` 
-
-Example 2. Condition for a guess-the-number game:
-
-```clojure
-(cond (< n answer) (guess-again "low")
-                      (> n answer) (guess-again "high")
-                      (correct!))
 ``` 
          
 ### Special Forms
--- a/docs/LanguageGuide.md
+++ b/docs/LanguageGuide.md
@@ -69,6 +69,16 @@ foo ; symbol
 (defmodule <name> <definition1> <definition2> ...) ;; The main way to organize your program into smaller parts
 ```
 
+### Conditional statements with 'cond'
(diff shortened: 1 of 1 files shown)


hellerve:
This is a great start!

I suggest we also add a docstring directly to `cond`. This would look like this:

```clojure
(doc cond "this is the documentation for cond")
```

For an example, look for instance at the documentation for `collect-into`:

https://github.com/carp-lang/Carp/blob/69ffac40975a24914e518b908c283df2f024b52e/core/Macros.carp#L127-L128

A more deep dive into how to document things in Carp can be found [in my blog post on documentation in Carp](https://blog.veitheller.de/Carp_Patterns_I:_Documentation.html).

I hope that helps!

docs/LanguageGuide.md
```diff
@@ -69,6 +69,36 @@ foo ; symbol
 (defmodule <name> <definition1> <definition2> ...) ;; The main way to organize your program into smaller parts
 ```
 
+### Conditional statements with 'cond'
+The cond statement executes a block of code if a specified condition is true. If the condition is false, another block of code can be executed. Examples are given below:
+
+```clojure
```
hellerve, line 75: I’m not sure we need to give the whole definition of `cond` here. I think more examples might be more useful!

OTOH it might be instructive for people to understand how `cond works, so I’m not sure.

docs/LanguageGuide.md
```diff
@@ -69,6 +69,36 @@ foo ; symbol
 (defmodule <name> <definition1> <definition2> ...) ;; The main way to organize your program into smaller parts
 ```
 
+### Conditional statements with 'cond'
+The cond statement executes a block of code if a specified condition is true. If the condition is false, another block of code can be executed. Examples are given below:
```
hellerve, line 73: I’d put `cond` in backticks here, to make sure it is rendered as a code element. What do you think?

docs/LanguageGuide.md
```diff
@@ -69,6 +69,36 @@ foo ; symbol
...
+         'if
+         (car xs)
+         (cadr xs)
+         (cond-internal (cddr xs)))))))
+
+(defmacro cond [:rest xs]
+  (cond-internal xs))
+
+```
+Here's an example to print a statement depending on whether it is < or > 10
```
hellerve, line 93: Some sort of punctuation at the end of the sentence might be good!

docs/LanguageGuide.md
mitchell-frost: I'll do that and add the working of cond as well.Thank you so much.

docs/LanguageGuide.md
mitchell-frost: I put in a few more examples. Shall be glad to make changes once you check into it.

docs/LanguageGuide.md
mitchell-frost: Done! :)

docs/LanguageGuide.md
eriksvedang: Yeah, the definition of cond-internal should not be here.

docs/LanguageGuide.md
```diff
@@ -69,6 +69,58 @@ foo ; symbol
...
+Example 1. Printing a statement depending on whether it is < or > 10:
+
+```clojure
+(cond
+          (< 10 1) (println "Don't print!")
+          (> 10 1) (println msg)
+          (println "Don't print!"))
+``` 
+
+Example 2. Condition for a guess-the-number game:
```
eriksvedang, line 116: I don't think we need more than one example.

docs/LanguageGuide.md
```diff
@@ -69,6 +69,58 @@ foo ; symbol
 (defmodule <name> <definition1> <definition2> ...) ;; The main way to organize your program into smaller parts
 ```
 
+### Conditional statements with 'cond'
+```(doc cond "this is the documentation for cond")```
```
eriksvedang, line 73: Not sure what purpose this line has....

hellerve: I think this was because of my comment above demonstrating the usage of `doc`. I should’ve been more clear about this: That line goes into the file where `cond` is defined, and the docstring shouldn’t be literally what I wrote above, but an actual documentation for `cond`. Sorry for the confusion, @eriksvedang and @mitchell-frost!

mitchell-frost:
I have made the necessary changes @hellerve and @eriksvedang. Please have a look!

hellerve:
Looks good to me now! Thanks!

eriksvedang:
Cheers!

PR #844, opened 2020-05-25
Improve register-type documentation

sdilts:
Add info about specifying c names in the `register-type` doc string.


hellerve:
Thanks!

PR #845, opened 2020-05-26
Make ref literals work in the repl

hellerve:
This PR makes reference literals work in the REPL, similarly to what @eriksvedang and I did for externals, modules, matches, and so on.

Cheers


eriksvedang:
Does this mean I can evaluate something like `(ref (foo))` in the REPL?

hellerve:
That should work now, yes!

PR #847, opened 2020-05-27
Add base16 and base2 literals

hellerve:
This PR adds base16 and base 2 literals. It only needs a parser change to work, and they work for integers and long types.

The literals:
- `0xabcd` is a literal hexadecimal integer,
- `0b1010` is a literal binary integer.

By suffixing with `l` or we can make them `Long` literals. Theoretically, they could also work for byte and float types without decimals, but currently only binary literals work with them, since the suffixes `b` and `f` are indistinguishable from hexadecimal literals. We could add additional syntax (like opting for prefixes instead of suffixes), but I don’t think that would add much value for the added complexity.

Cheers

### DIFF
--- a/src/Parsing.hs
+++ b/src/Parsing.hs
@@ -2,6 +2,7 @@
 
 module Parsing (parse, validCharacters, balance) where
 
+import Data.Bits (shift)
 import Numeric (readHex)
 import Text.Parsec ((<|>))
 import qualified Text.Parsec as Parsec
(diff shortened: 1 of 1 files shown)


jacereda:
I always liked the approach taken by some Forths that allows specifying arbitrary bases, something like:

```
2'1001
16'deadbeef
8'0777
```

hellerve:
I thought about that, and I think it wouldn’t be too hard to implement; I didn’t go down that route for now, but I think we can think about changing to that approach in the future!

trentgill:
Wow this is amazing! Feels a very natural fit as the most common reason to use hex & binary is for interfacing directly with memory addresses & most documents use this same syntax.

:D:D:D

PR #848, opened 2020-05-27
Introspection improvements

### LINKED ISSUE #839: Improve module introspection
Currently, `Introspect` has minimal support for introspecting modules. Inspired by @hellerve's recommendations, I think we can make a few improvements:

- [ ] Introspecting a module should return the complete contents of the module body
       In order to accomplish this we'd have to:
       - Get the module and its `env`
       - for all bindings in the module `env` get their s-expressions
       - join the results into a single form `(defmodule Mod ...binding s-exprs)`
- [ ] Given a type, give users the ability to get either the type declaration (currently supported) *or* the generated equivalent of the module declaration (see previous bullet).

scolsen:
There are a couple of changes in this PR, all related to introspection:

1. Change `s-expr` from a primitive to a command--this form is easier to use when it evaluates its arguments, when it doesn't eval its argument we have to perform an extra lookup and it requires a lot of finagling for use in macros/dynamic code, the first commit message has more detail.
2. Support introspecting `defmodules`. Introspecting a defmodule now returns the entire contents of the module:

```clojure
(defmodule Goo (defn bar [x] x) (defn qux [y] y) (def x 4) (defmodule T (defn g [z] z)))
(Dynamic.s-expr Goo)
=> (defmodule Goo (defmodule T (defn Goo.T.g [z] z)) (defn Goo.bar [x] x) (defn Goo.qux [y] y) (def Goo.x 4))
```

3. Allow users to get the module for a type, instead of just the type definition. If desired, users can now pass a second bool argument to `s-expr` to get a type's module declaration instead of its type definition:

```clojure
(deftype Bar [x Int])
(Dynamic.s-expr Bar)
=> (deftype Bar [x Int])
(Dynamic.s-expr Bar true)
=> (defmodule Bar (instantiate Bar.copy) (instantiate Bar.delete) (instantiate Bar.init) (instantiate Bar.prn) (instantiate Bar.set-x) (instantiate Bar.set-x!) (instantiate Bar.str) (instantiate Bar.update-x) (instantiate Bar.x))
```

4. Updated `Introspect` to account for `s-expr` changes and added basic tests for the module.

Many thanks to @hellerve for the idea and the advice!

Fixes #839


core/Introspect.carp
```diff
@@ -4,75 +4,94 @@
...
-  (defmacro sumtype? [binding]
-    (if (empty? (eval (list 's-expr binding)))
-        false
-        (list? (caddr (eval (list 's-expr binding))))))
-  (doc arity 
+  (defndynamic sumtype? [binding]
+    (let [s (s-expr binding)]
+      (if (empty? s)
+          false
+          (list? (caddr s)))))
```
hellerve, line 51: Should we add a length check here (and above in struct)?

scolsen: Will do! Might be nice to have something generic for this since it happens quite often in my experience. I did implement this elsewhere which more or less does the trick:

```
(defndynamic deconstruct [n else f]                                                                                                                                                  
    (fn [xs]                                                                                                                                                                           
      (if (< (length xs) n)                                                                                                                                                            
          else                                                                                                                                                                         
          (f xs))))                                                                                                                                                                    
                                                                                                                                                                                       
  ;; "safe" versions of cdr, cadr, etc.                                                                                                                                                
  (hidden cdr?)                                                                                                                                                                        
  (private cdr?)                                                                                                                                                                       
  (defndynamic cdr? [xs]                                                                                                                                                               
    (let [f (Folders.Consumers.deconstruct 1 '() cdr)]                                                                                                                                 
      (f xs)))                                                                                                                                                                         
  (hidden cddr?)                                                                                                                                                                       
  (private cddr?)                                                                                                                                                                      
  (defndynamic cddr? [xs]                                                                                                                                                              
    (let [f (Folders.Consumers.deconstruct 2 '() cddr)]                                                                                                                                
      (f xs)))  
```

hellerve: Sounds good, but here I think `(or (empty? s) (< (length s) 3))` should be enough, right?

Issue #805, opened 2020-05-15
Deeply nested lambdas can't refer to identifiers from their parent environments


scolsen:
Lambdas that are nested deeper than 1 level cannot refer to identifiers from their ancestors:

```
鲤 (defn my-curry [f] (fn [x] (fn [y] (f x y))))
鲤 (my-curry Int.+)
out/main.c:9149:18: error: use of undeclared identifier 'f'
    _14_env->f = f;
                 ^
1 error generated.
carp: callCommand: clang -fPIC -lm -g out/main.c -o "out/Untitled"  -I/Users/scottolsen/dev/Carp//core/   (exit 1): failed
[i] Carp ∎ 
```

Vertmo:
From reading the generated code, I think I can guess the problem. I have to look a bit more into the compiler source, but I think I should be able to fix it :)

scolsen:
That would be amazing!

scolsen:
Thanks, @Vertmo!

PR #850, opened 2020-05-28
Fixed the issue with nested capture and nested lambdas

Vertmo:
This should be fixing #805 
The solution I went for is to modify the `CaptureMode` datatype. The `Capture` constructor now takes an int which indicates the "Capture Distance" that is the number of lambdas between the binding of the variable and its usage (this is easy to calculate during Qualification).
This allows to differentiate between a variable bound by the last encountered lambda, and one which is present in its environment, and can be used to fill the next environment.

### DIFF
--- a/examples/nested_lambdas.carp
+++ b/examples/nested_lambdas.carp
@@ -1,0 +1,2 @@
+(defn my-curry [f] (fn [x] (fn [y] (f x y))))
+(defn double-curry [f] (fn [x] (fn [y] (fn [z] (f x y z)))))
--- a/examples/nested_lambdas.carp
+++ b/examples/nested_lambdas.carp
@@ -1,2 +1,5 @@
 (defn my-curry [f] (fn [x] (fn [y] (f x y))))
 (defn double-curry [f] (fn [x] (fn [y] (fn [z] (f x y z)))))
+
+(defn main []
+  (do (((my-curry (fn [x y] (Int.+ x y))) 1) 2)))
--- a/run_carp_tests.sh
+++ b/run_carp_tests.sh
@@ -43,6 +43,7 @@ done
 ./carp.sh ./examples/guessing.carp -b
 ./carp.sh ./examples/no_core.carp --no-core --no-profile -b
 ./carp.sh ./examples/check_malloc.carp -b
+./carp.sh ./examples/nested_lambdas.carp -b
 
 # Run tests which rely on SDL unless the `--no_sdl` argument was passed in
 if [[ ${NO_SDL} -eq 0 ]]; then
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -110,8 +110,7 @@ concretizeXObj allowAmbiguityRoot typeEnv rootEnv visitedDefinitions root =
              let -- Analyse the body of the lambda to find what variables it captures
                  capturedVarsRaw = collectCapturedVars okBody
                  -- and then remove the captures that are actually our arguments
-                 capturedVars = filter (\xobj -> obj xobj `notElem` argObjs)
-                                capturedVarsRaw
+                 capturedVars = filter (\xobj -> obj (toGeneralSymbol xobj) `notElem` argObjs) capturedVarsRaw
 
                  -- Create a new (top-level) function that will be used when the lambda is called.
                  -- Its name will contain the name of the (normal, non-lambda) function it's contained within,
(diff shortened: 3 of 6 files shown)


scolsen:
Awesome! Overall looks pretty good to me, but I'm not super familiar with this code so I requested @eriksvedang's review too.

Looks like we may need to update the evaluator too-- currently doing:

```
鲤 (defn my-curry [f] (fn [x] (fn [y] (f x y))))
鲤 (my-curry Int.+)
λ
```

Works as expected but, applying the lambdas:

```
鲤 (defn my-curry [f] (fn [x] (fn [y] (f x y))))
鲤 (((my-curry Int.+) 1) 2)
;; hangs
```

In the repl causes it to hang, but it works fine if it's in a source file that one builds/executes.

eriksvedang:
Very cool! Looks great to me at a first glance, but I'll wait with merge:ing until I've had some time to play around with it.

Vertmo:
> Awesome! Overall looks pretty good to me, but I'm not super familiar with this code so I requested @eriksvedang's review too.
> 
> Looks like we may need to update the evaluator too-- currently doing:
> 
> ```
> 鲤 (defn my-curry [f] (fn [x] (fn [y] (f x y))))
> 鲤 (my-curry Int.+)
> Compiled to 'out/Untitled' (executable)
> λ
> ```
> 
> Works as expected but, applying the lambdas:
> 
> ```
> 鲤 (defn my-curry [f] (fn [x] (fn [y] (f x y))))
> 鲤 (((my-curry Int.+) 1) 2)
> ;; hangs
> ```
> 
> In the repl causes it to hang, but it works fine if it's in a source file that one builds/executes.

With respect to this issue, the case of `Defn seems to be treated weirdly in the interpreter. Indeed, it seems like a `Defn` XObj is put in the environment "as-is", which seems fine in itself, however, when retrieved from the environment, the `Defn` is never treated, it simply returns the same XObj (see line 218 of **Eval.hs**). 
My opinion is that a `Defn` should either be evaluated into a Closure (with an empty environment) at this point, or stored as if it was a `Def (Fn [])` in the environment (although the second one may cause complications elsewhere). What do you think ?

hellerve:
Just as a heads-up: changing how `defn` is treated will have repercussions on the whole compiler, most prominently the emitter. It will probably lead to a few problems down the road. I think the emitted C code might be problematic, but I’m not sure and that shouldn’t prevent you from trying to rework it!

Vertmo:
> Just as a heads-up: changing how `defn` is treated will have repercussions on the whole compiler, most prominently the emitter. It will probably lead to a few problems down the road. I think the emitted C code might be problematic, but I’m not sure and that shouldn’t prevent you from trying to rework it!

Oh I was only planning to make changes inside the evaluator (aka the `Eval.hs` file). I don't think that would impact the compilation ?

hellerve:
The way `defn` is stored gets passed through all of the stages to the backend, because `defn` is how functions are defined that then need to be emitted. That’s why they are passed through the evaluator unchanged: because it’s actually the emitter’s job to do things with it.

I hope that makes sense. If not, I’d be happy to hop on a quick call and talk about the design of the compiler at some point today.

Vertmo:
> The way `defn` is stored gets passed through all of the stages to the backend, because `defn` is how functions are defined that then need to be emitted. That’s why they are passed through the evaluator unchanged: because it’s actually the emitter’s job to do things with it.
> 
> I hope that makes sense. If not, I’d be happy to hop on a quick call and talk about the design of the compiler at some point today.

That would be useful I think, because I'm not sure I understand how the compiler and the evaluator interact together. I will message you on Gitter :)

eriksvedang:
Just to give some more context for the current design: `defn` is it's own thing and not the more general `def` with a lambda because I wanted easy to read, normal C style, function definitions to be emitted,  not something with lambdas getting assigned to global variables (this can be important for optimizations on the C side too). If we can use the more general solution and still emit such code, I'm all for changing it. It might be hard though since `def`:s are not immutable (maybe they should be, or we could use meta data for it as in @scolsen:s recent proposal).

hellerve:
Actually, it seems like this is an issue from before, and I think this doesn’t affect the mergeability of this PR (since it still fixes the problem).

The problem exhibited by the example by @scolsen is related to the way the REPL special-cases static expressions, anmely in `executeCommand`. Here, we try to match what needs to be evaluated statically and wrap it in a `main` function if needed. The problem is that we don’t match lists in the calling position here yet, and I’m not quite sure how to get that to work easily, since it might be arbitrarily deep—we might need to implement a fix-point-like search there.

Anyway, this means that this PR is actually correct and can be merged, and that we should open another issue.

scolsen:
> Actually, it seems like this is an issue from before, and I think this doesn’t affect the mergeability of this PR (since it still fixes the problem).
> 
> The problem exhibited by the example by @scolsen is related to the way the REPL special-cases static expressions, namely in `executeCommand`. Here, we try to match what needs to be evaluated statically and wrap it in a `main` function if needed. The problem is that we don’t match lists in the calling position here yet, and I’m not quite sure how to get that to work easily, since it might be arbitrarily deep—we might need to implement a fix-point-like search there.
> 
> Anyway, this means that this PR is actually correct and can be merged, and that we should open another issue.

Great insight! Maybe we can express it with a `fold(M)` (gotta `foldM` all)?

hellerve:
We might! Go ham with those morphisms if you want to!

eriksvedang:
OK, thanks a lot for clearing that up @hellerve !

Vertmo:
> Actually, it seems like this is an issue from before, and I think this doesn’t affect the mergeability of this PR (since it still fixes the problem).
> 
> The problem exhibited by the example by @scolsen is related to the way the REPL special-cases static expressions, namely in `executeCommand`. Here, we try to match what needs to be evaluated statically and wrap it in a `main` function if needed. The problem is that we don’t match lists in the calling position here yet, and I’m not quite sure how to get that to work easily, since it might be arbitrarily deep—we might need to implement a fix-point-like search there.
> 
> Anyway, this means that this PR is actually correct and can be merged, and that we should open another issue.

I'm not completely sure the problem actually comes from `executeCommand`, I still have a feeling it's coming from `eval` (the case at line 218 to be exact). I'm experimenting to see if I can fix it, I've made a bit of progress but I still have some issues. I'll update you when I find a good solution :)

hellerve:
@Vertmo that sounds great, please do push that branch!

eriksvedang:
Thanks so much for this @Vertmo !!!
Really nice solution, I'm impressed.
Feels promising that you're investigating some other bugs :)

PR #854, opened 2020-05-29
Use unlines instead of joinWith.

jacereda:
Just a cosmetic change.


hellerve:
The failing tests look like there is an extra empty line inserted somewhere by the compiler now; is that possible?

jacereda:
Yes, I probably messed it in some place, let me check.

scolsen:
Ahh yeah, `unlines` appends an `\n` after the final item in the list, while `joinWith` uses `intercalate` which doesn't.

jacereda:
Oh, right, too many hours in front of a computer today :)

scolsen:
> Oh, right, too many hours in front of a computer today :)

I like the general idea though--maybe we can replace `joinWith "\n"` by a new function partially applied to `\n`:

```
joinLines :: [[a]] -> [a]
joinLines = joinWith "\n"
```

Issue #855, opened 2020-05-29
Decide target standard


jacereda:
Which standard should the emitted code adhere to?

- C89 is supported everywhere, but lacks some recent niceties (`stdint.h` for one).

- C99 is supported by all current compilers, even in the embedded world you'd have trouble finding a non-compliant compiler. 

So I guess the deciding factor is whether to support legacy platforms like, say, Amiga/Atari.

scolsen:
I don't know too much about the details of each standard, so forgive me if this is a silly question:

Would it be feasible to support C99 by default and tuck C89 support away under a compiler option?

It wouldn't be insignificant work, but seems like it'd be possible by using different templates and emitting the appropriate variant based on the requested standard.

If we have to choose just one, choosing C99 seems right to me, but on the other hand it would be cool to support use cases like Amiga/Atari.

jacereda:
There's a table at https://en.wikipedia.org/wiki/C99#Implementations

I would read Partial/Mostly as valid for our purposes, since our codegen won't use the obscure features not present in those.

jacereda:
Partial list of things we would need to get rid of:

- `// comments`

- Variable declarations in `for` loops, i.e. `for (int i = 0...`.

- `inline` keyword.

jacereda:
Even legacy platforms will probably have an old GCC (2.95.X was mostly the compiler of choice in the 68k platforms) that supports all of those when compiling with extensions enabled IIRC. (Maybe `inline` would need to be a `define`d as `__inline__` or something like that).

GrayJack:
I agree that C99 is a better standard to be used, language-wise. I think supporting those super legacy systems is not worth it compared with with the niceties.

I also think that if anyone wants very-very much write Carp for those legacy systems, they can always do their version of the corelib ( they pretty much would just need to rewrite the `deftemplate` ones, right?)

sdilts:
It might be worth the upgrade just for `stdbool.h`
There are a few things we probably shouldn't include in Carp such as variable-length arrays though.

rkaippully:
Would it make sense to have a "freestanding" target that does not depend on any standard library? This would help in writing very low level code such as operating system, bootloaders etc in Carp. AFAIK, Rust and Nim support this.

hellerve:
@rkaippully the way we currently support this is by doing `--no-core`, which will not load the standard library. You can then load the parts of it that you need/can load yourself.

This has been enough in the firmware/embedded experiments that are going on until now.

[...]

PR #856, opened 2020-05-29
C99 fixes

jacereda:
This PR adds support for compiling with `-std=c99`.

### DIFF
--- a/src/Sumtypes.hs
+++ b/src/Sumtypes.hs
@@ -105,19 +105,18 @@ tokensForCaseInit allocationMode sumTy@(StructTy (ConcreteNameTy typeName) typeV
                        , case allocationMode of
                            StackAlloc -> "    $p instance;"
                            HeapAlloc ->  "    $p instance = CARP_MALLOC(sizeof(" ++ typeName ++ "));"
-                       , joinLines (map (caseMemberAssignment allocationMode correctedName)
-                                        (zip anonMemberNames (caseTys sumtypeCase)))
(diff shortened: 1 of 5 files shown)


jacereda:
If we decide in #855 that c89 is the way to go the fixes are still valid.

src/Sumtypes.hs
```diff
@@ -345,6 +339,15 @@ copyCase typeEnv env concreteStructTy@(StructTy _ typeVariables) (theCase, isFir
   let (name, tys, correctedTagName) = namesFromCase theCase concreteStructTy
   in unlines
      [ "    " ++ (if isFirstCase then "" else "else ") ++ "if(pRef->_tag == " ++ correctedTagName ++ ") {"
-     , joinLines (map (memberCopy typeEnv env) (zip (map (\anon -> name ++ "." ++ anon) anonMemberNames) tys))
+     , joinLines $ memberCopy typeEnv env <$> unionMembers name tys
      , "    }"
      ]
+
+anonMemberName :: String -> String -> String
```
eriksvedang, line 346: What issue does this resolve? (mainly curious)

jacereda: Anonymous unions aren't allowed in C99. Sumtype cases are now accessed as `.u.memberX`.

eriksvedang: OK, I see. Thanks!

eriksvedang:
Thanks!

PR #857, opened 2020-05-30
Add joinLines.

jacereda:


core/carp_pattern.h
```diff
@@ -87,9 +87,7 @@ String Pattern_internal_classend(PatternMatchState *ms, String p) {
             } while (*p != ']');
             return p + 1;
         }
-        default: {
-            return p;
-        }
+        default: { return p; }
```
jacereda, line 90: I always have to discard this diff before committing. My `clang-format` version must be different from the one you're using. I'd suggest getting rid of the braces, maybe both versions will behave the same in that case.

jacereda: My version is 7.1.0 BTW.

jacereda: Just pushed it without the braces, can someone confirm older versions format it the same?

hellerve: Actually, my version (10.0.0) does not reformat that line.

jacereda: Oh, right, I'll update my version then.

PR #858, opened 2020-05-30
Evaluation of nested defn

Vertmo:
As was discussed in #850, evaluation (in the REPL) of code such as:
```
(defn my-curry [f] (fn [x] (fn [y] (f x y))))
(((my-curry +) 1) 2)
```
hangs the evaluator.

I believe I found a solution to this problem, by changing one of the case of the evaluator. This examples (and other tries) do work.
However, using `Int.+` instead of `+` causes the error:
```
I couldn’t find the symbol 'x' at line 1, column 39 in 'REPL'.
```
`x` referring to the variable bound by the first lambda-abstraction. This may be an unrelated bug ? I'm not sure I understand the codebase well enough to say.
Also a weird side effect is that the code above actually compiles in the current state, despite #805.
That's because the application is fully evaluated to `3` during compilation. I don't know if this partial evaluation is a desired outcome ? It of course doesn't happen when `my-curry` is called under `(defn main [] ...)` for instance.

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -186,6 +186,17 @@ eval ctx xobj@(XObj o i t) =
                     return (newCtx, res)
                   Left err -> return (newCtx, Left err)
 
+       XObj (Lst [XObj (Defn _) _ _, name, (XObj (Arr params) _ _), body]) _ _:args ->
+         case checkArity params args of
+           Left err -> return (evalError ctx err (info xobj))
+           Right () ->
+             do (newCtx, evaledArgs) <- foldlM successiveEval (ctx, Right []) args
+                case evaledArgs of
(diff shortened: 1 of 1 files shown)


hellerve:
That does in fact not look like what we want, because it will evaluate `defn`s dynamically. We want them to stay in place and not be evaluated.

This change as-is changes the semantics of the entire dynamic evaluator, because it moves evaluation of static code into the dynamic evaluator.

Vertmo:
Oh right that makes sense, I was not in the correct frame of mind regarding the evaluator. I will rethink my approach and try to come up with something more appropriate :)

hellerve:
Don’t worry, it’s pretty easy to get mixed up in there, especially in the evaluator!

PR #862, opened 2020-06-01
Generalization of the evaluation of static defn

Vertmo:
Once again trying to fix the evaluation issue outlined in #850.
After the mistake made in #858, I tried to generalize the current "static evaluation", rather than change it completely. The results fixes the example shown in #850, but there is still a few issues that I would like to discuss before implementing a definitive solutions.
The current push code does evaluate correctly the following example:
```
(defn ident [x] x)
((ident (fn [x] x)) 1)
```
which was not the case before this PR. However I have found 2 categories of bug this does not fix:

* consider `(+ (ident 2) (ident 2)` which should evaluate to `4`. With the current code, the evaluator returns an error, as it tries to apply the primitive `+` to `((defn ident [...]) 2)`. 

    * I think this could be solved by checking `isStaticCall` on the arguments of a call (primitive or otherwise before applying it, but I worry this would make the evaluator code too complex. 
    * Another, maybe more robust solution would be to replace the return type of `eval` with a three way sum type : one representing an error, another a fully evaluated value, and the last a value waiting for a static evaluation; this would make treatement more explicit and hopefully less prone to forget "special cases", but would also necessitate a hefty refactor.

* Second problem : consider either `(let [x 1] (ident x)` or ((fn [x] (ident x) 1). Both these cases exhibit a similar bug : after dynamic evaluation, we're left with the expression `(ident x)`. The issue is that, as we go back up from recursive call in the `eval` function, the environment linking `x` to it's value `1` is lost. 
   * the first solution I thought of was to `eval` the arguments of a static call during the dynamic evaluation. This unfortunately caused other problems (for instance, with `((ident (fn [x] x)) 1)` the `fn` became a closure which was refused by typing when using `callFromRepl`; in the case of `((ident ident) 1)`, there was an issue with the borrow checker.
   * another, maybe safer solution would be to "save" the environment around the static call, in order to pass it when using `callFromRepl`. I'm not 100% sure how that would work however, and maybe that would need to introduce some new weird syntactic constructs, which would pollute the AST.

Here, I'd very much like to hear about you on the subject of these issues (and of course of the code I've already written). Thanks in advance !

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -215,13 +215,13 @@ eval ctx xobj@(XObj o i t) =
 
        x@(XObj (Lst [XObj (Primitive prim) _ _, _]) _ _):args -> (getPrimitive prim) x ctx args
 
-       XObj (Lst (XObj (Defn _) _ _:_)) _ _:_ -> return (ctx, Left HasStaticCall)
-       XObj (Lst (XObj (Interface _ _) _ _:_)) _ _:_ -> return (ctx, Left HasStaticCall)
-       XObj (Lst (XObj (Instantiate _) _ _:_)) _ _:_ -> return (ctx, Left HasStaticCall)
-       XObj (Lst (XObj (Deftemplate _) _ _:_)) _ _:_ -> return (ctx, Left HasStaticCall)
-       XObj (Lst (XObj (External _) _ _:_)) _ _:_ -> return (ctx, Left HasStaticCall)
-       XObj (Match _) _ _:_ -> return (ctx, Left HasStaticCall)
-       [XObj Ref _ _, _] -> return (ctx, Left HasStaticCall)
+       XObj (Lst (XObj (Defn _) _ _:_)) _ _:_ -> return (ctx, Left (HasStaticCall xobj i))
+       XObj (Lst (XObj (Interface _ _) _ _:_)) _ _:_ -> return (ctx, Left (HasStaticCall xobj i))
+       XObj (Lst (XObj (Instantiate _) _ _:_)) _ _:_ -> return (ctx, Left (HasStaticCall xobj i))
+       XObj (Lst (XObj (Deftemplate _) _ _:_)) _ _:_ -> return (ctx, Left (HasStaticCall xobj i))
+       XObj (Lst (XObj (External _) _ _:_)) _ _:_ -> return (ctx, Left (HasStaticCall xobj i))
+       XObj (Match _) _ _:_ -> return (ctx, Left (HasStaticCall xobj i))
+       [XObj Ref _ _, _] -> return (ctx, Left (HasStaticCall xobj i))
 
        l@(XObj (Lst _) i t):args -> do
          (newCtx, f) <- eval ctx l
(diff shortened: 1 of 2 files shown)


eriksvedang:
@Vertmo Sorry for being slow to respond. This is a very interesting and tricky problem indeed...

How about checking (after macro expansion) if there are *any* calls to something static in the form and if so, treat the whole thing as static code? (Perhaps this is what your first suggestion for a fix means, excuse me if I'm just reiterating your solution in my own words)

To be more clear with what I mean, for example this:

```clojure
[(IO.println "hello")]
```

currently returns a dynamic array but should in that case become a static one, since `println` is a static function. 

We don't really have a way of returning values from static calls, since they just print the result to stdout. So any variant that relies on something like that is probably very tough to implement.

You've got more hands on experience with the problem though, so perhaps I'm missing something?

Vertmo:
@eriksvedang I feel like that would work yes, and probably simplify the whole thing, but it would also mean more code would get compiled and "statically executed" rather than being evaluated. Maybe that's not too much of a loss though. I will tinker with it

Vertmo:
Ok, new approach, this time following @eriksvedang recommandation simply compiling any code that contains a static call.
This is implemented by adding a new "exception" (see `EvalError` and its added constructor `HasStaticCall`) which triggers the compilation of the code.
There is still as I expected a problem when using macros (because of the interleaving of `eval` and `expand`) which I will try to fix.

eriksvedang:
Oh, very clever approach! I'll give it a spin...

eriksvedang:
Works well for me (as long as I don't use macros, as you said). Very cool, I like the simplicity of this solution! Do you have any ideas for how to fix macros? Does the "exception" need to be threaded back to the original form or something like that..?

Vertmo:
Well, I'm kind of stumped on macros to be honest. The issue is that macros do need to be expanded before code generation. I've tried an approach consisting in "catching the exception" thrown by `eval` when expanding, but the issue is that this doesn't allow us to get the expanded form back. See for instance the following program:
```
(defn ident [x] x)
(defmacro macident [x] (ident x))
```
With my current tinkering, I can run `(macident 42)` but not `(macident (macident 42))` because the latter would need to `eval` and `expand` a parameter to the static function `ident`. Actually I can't even compile a program where `(macident (macident 42))` would be wrapped under a main, but it seems that it was already the case before my changes.

So all in all... I'm not really sure of what would be a good solution, without changing the evaluation / expansion more (as in making them independant from one another, which would also involve some code copy, with the expansion being a "weaker" form of evaluation. I'm not sure how efficient that would be though.

hellerve:
As promised in the chat, let me outline a few idea and opinions around this PR:

Firtstly the examples provided by @Vertmo are partially a little flawed, which is what I think led you down a somewhat misguided path. Let me explain what I mean. Your last example was:

```clojure
(defn ident [x] x)
(defmacro macident [x] (ident x))
```

This example shouldn’t work as is, because it tells the evaluator to dynamically evaluate a static form. There are two ways in which something like this could work:

```clojure
(defn ident [x] x)
(defmacro macident [x] (list 'ident x)) ; builds a static call to ident

; vs

(defndynamic ident [x] x)
(defmacro macident [x] (ident x)) ; does everything dynamically
```

The example your provided should throw an error, because it tries to evaluate a static form dynamically. This is somewhat intricate to catch, but with our new evaluator, I suppose any form inside `macroExpand` that returns `HasStaticCall` should be considered an error. With this, I think we are good here; although I’m not positive this approach in general will cover all cases—but I think it’s valuable to try out!

Vertmo:
@hellerve Right, that all makes sense ^^ With that in mind I will do a few more test, and try to improve my error message as well.

Vertmo:
Ok I'm fairly convinced now ^^ Thanks for your feedback

eriksvedang:
@hellerve Are you happy with this now? :)

hellerve:
@eriksvedang I haven’t had time to look at it or test it yet, but if the latest changes solved the initial problems I think we’re good!

eriksvedang:
I don't think it actually do find the error, but I merged it anyways since I think it's miles better than what we had before. Please open an issue for the bug (if it actually do persist) and we can work from there.

PR #863, opened 2020-06-02
Allow "Instantiate" functions to implement interfaces

scolsen:
This change makes it possible to declare auto-derived getter/setter
functions on types as implementations of interfaces, e.g. Pair.a.

The same checks for normal functions apply to the instantiations. For
example, given an interface:

```
(definterface left (Fn [(Ref (f a b))] (Ref a)))
```

`Pair.a` can implement this interface, but `Maybe.Just` cannot--since it
results in a kind mismatch.

I also realized that interfaces are currently too flexible/permissive.
For example, given an interface:

```
(definterface id (Fn [a] a))
```

Most, if not all, functions can implement this interface, even when the
implementation clearly doesn't conform. For example, `Pair.a` or even a
function with the type `([String] Int)` are both valid implementations
of this interface. This is because we currently only check if interfaces
and implementation signatures are unifiable. Since we consider any type
unifiable with a variable, any implementation passes, even when the
argument type and return type are clearly not the same. We'll need to
constraint solve on the interface and implementation signatures to fix
this--since that's a more significant change, it's omitted from this PR.
For now a new TODO will have to suffice.

### DIFF
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -179,6 +179,11 @@ registerInInterfaceIfNeeded ctx path@(SymPath _ _) interface@(SymPath [] name) d
   in case lookupInEnv interface typeEnv of
        Just (_, Binder _ (XObj (Lst [XObj (Interface interfaceSignature paths) ii it, isym]) i t)) ->
          if checkKinds interfaceSignature definitionSignature
+           -- TODO: Constrain interfaces and implementations.
+           -- The areUnifiable check alone is not sufficient. It will allow almost any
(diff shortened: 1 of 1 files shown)


eriksvedang:
Oh, that’s a pretty serious error you’ve found there! Hopefully it’s pretty easy to fix though...

scolsen:
> Oh, that’s a pretty serious error you’ve found there! Hopefully it’s pretty easy to fix though...

About to open a PR :)

Issue #521, opened 2019-07-07
UnificationFailure


jacereda:
I'm getting the following error:

```
carp: UnificationFailure {unificationFailure = {a == (HitRecord a) (ord OrdMultiSym)} Nothing, Nothing in XObj {obj = Sym a Symbol, info = Nothing, ty = Nothing}, unificationMappings = fromList []}
CallStack (from HasCallStack):
  error, called at src/Concretize.hs:419:19 in CarpHask-0.3.0.0-inplace:Concretize
```

Looks like it's triggered by `checkForConflict` in `Constraints.hs`:

```
    if doesTypeContainTyVarWithName name otherTy
    then Left (UnificationFailure constraint mappings)
```

Is that supposed to be be raised by a problem in my code or unimplemented stuff?

eriksvedang:
Not sure..! Will investigate asap.

chrisosaurus:
@jacekschae what are you doing when you see this error? do you have reproduction steps?

scolsen:
It's been a while, but I ran into a couple of these whilst playing with recursive type definitions.

Is your type in this case recursive? I can't quite tell from the error, but `a == (Hitrecord a)` makes it seem like it might be? ...or at least as though there's some other sort of use of type vars that Carp is not expecting. I had been investigating implementing support for expressing recursive types before I went traveling (afaik, Carp currently doesn't handle recursive type definitions such as `(deftype (My-list a) (Nil []) (List [a (My-list a)])` but there are ways around this).

jacereda:
> what are you doing when you see this error? do you have reproduction steps?

It happens on a branch where I've changed the `Vector` functions to be generic, I'll try to isolate the problem and post a reproduction.

jacereda:
> Is your type in this case recursive?

No, it was something like:

```
(deftype (HitRecord a)
    [
     t a
     p (Vector3 a)
     normal (Vector3 a)
     ])
```

jacereda:
I simplified it and it's happening in master:

```clojure
(deftype (HitRecord a) [ t a ])
(deftype (CurrentHit a) [ hr (Maybe (HitRecord a)) ])
(defn hit [elts]
  (let [
        starting (CurrentHit.init (Maybe.Just (HitRecord.init 0.0)))
        closest (fn [cur h] cur)
        red (Array.reduce &closest starting elts)
        ]
    @(CurrentHit.hr &red)))
```

jacereda:
Ok, simpler:

[...]

PR #864, opened 2020-06-02
Constrain Interfaces

scolsen:
This PR predominantly accounts for the issue noted in #863, namely, interfaces were too permissive and actually allowed ill-typed implementations.

## The problem

Previously, when unifying an interface and implementation, we only checked whether their types are unifiable. This alone is not enough to ensure the integrity of our typings, as the types `(a -> a)` and `(Int -> Double)` are considered unifiable but clearly the second type is not a correct interpretation of the former type. If you declare such an interface `a -> a` and a concrete function `Int -> Double` in carp today, and declare that your concrete function implements the interface it will work, even though it's not typed correctly and doesn't honor the interface's typing.

## The solution

After this PR, we match interfaces to implementation by solving a constraint, which fixes this problem and makes the type system more rigorous. Doing this also required solving #521 for this case as, while this issue is annoying for custom structs, it pretty much makes constrained interfaces unusable. While that bug existed, given an interface `a -> a` an implementation of `(Ref a) -> (Ref a)` which is clearly correct, wouldn't unify simply because it happened to use `a` as a type variable. I've added a special case that bypasses the check for **interface implementation solving only**, since this is a safer change than outright removing the check.

## Caveats

This change does break *one* thing. The `functor` example, had to be retyped in order to be correct now that interfaces are constrained. It's previous type `(definterface fmap (λ [(Ref (λ [a] a)) b] b))
` changed to the more accurate typing `(definterface fmap (λ [(Ref (λ [a] b)) (f a)] (f b)))`. Which works fine for direct uses of `fmap`, but it does break the one higher-order use of fmap in the example: `(defn higherOrder [x] (fmap &Int.inc x))`. Use of this function, for whatever reason, assigns it the return type of its first usage; so, in the example file it gets assigned the type `(Box Int) -> (Box Int)`. The second call on an Array produces erroneous code. I'm not sure why the function get's assigned the type of first use -- direct applications of fmap work fine and as expected.

Some further investigation into `higherOrder` reveals that it's precisely a function that. relied on the old problematic over-permissive nature of interface/implementation unification. If one expects its type in carp prior to this PR it is typed `a -> a` which, while accurate for this particular function, was again, a bad typing that was actually serving as the universal type, and would admit anything. So my opinion is that this particular function reveals some other potential issue in the type system and not a reason to deter this PR.

To help illustrate, here are some things that passed the type checker prior to this PR and how they work after this PR:
**Before this PR**:
```clojure
鲤 (definterface id (Fn [a] a))
鲤 (sig foo (Fn [Double] Int))
鲤 (defn foo [x] 2)
鲤 (implements id foo)
鲤 :i id
id : (Fn [a] a) = {
    foo
}
Defined at line 1, column 15 in 'REPL'
id : (Fn [a] a)
Defined at line 76, column 1 in '/Users/scottolsen/dev/Carp/core/Generics.carp'
鲤 :i foo
foo : (Fn [Double] Int)
;; clearly, the above situation is incorrect, and yet foo is registered as an implementation of id
```

AFTER ff0596e no longer solved.
~~related to #521, loading this file used to result in a Unification Failure (though defining the same in the repl worked fine)~~

```
(deftype (HitRecord a) [ t a ])
(deftype (CurrentHit a) [ hr (Maybe (HitRecord a)) ])
(defn hit [elts]
  (let [
        starting (CurrentHit.init (Maybe.Just (HitRecord.init 0.0)))
        ]
    @(CurrentHit.hr &starting)))
鲤 (load "hit.carp")
carp: UnificationFailure {unificationFailure = {a == (HitRecord a) (ord OrdMultiSym)} , unificationMappings = fromList []}
  error, called at src/Concretize.hs:438:19 in CarpHask-0.3.0.0-JeTRwcTheISKNh9w0jc5KF:Concretize
```
**After this PR**:

The ill-typed interface implementation is no longer permitted:

```clojure
鲤 (definterface id (Fn [a] a))
鲤 (sig foo (Fn [Double] Int))
鲤 (defn foo [x] 2)
鲤 (implements id foo)
[INTERFACE ERROR] foo : (Fn [Double] Int) doesn't match the interface signature (Fn [a] a)
[INTERFACE ERROR] foo : (Fn [Double] Int) doesn't match the interface signature (Fn [a] a) at REPL:1:13.

Traceback:
(implements id foo) at REPL:1:1.

鲤 
;; correct implementations pass
鲤 (defn bar [x] x)
鲤 (implements id bar)
鲤 :i id
id : (Fn [a] a) = {
    bar
}
Defined at line 1, column 15 in 'REPL'
id : (Fn [a] a)
Defined at line 76, column 1 in '/Users/scottolsen/dev/Carp/core/Generics.carp'
```

AFTER ff0596e -- no longer true
~~The old variable name sharing problem is also gone:~~

```
鲤 (load "hit.carp")
鲤 :i CurrentHit
CurrentHit : Type
Defined at line 2, column 11 in '/Users/scottolsen/dev/Carp/hit.carp'
CurrentHit : Module = {
    copy : (Fn [(Ref (CurrentHit a) q)] (CurrentHit a))
    delete : (Fn [(CurrentHit a)] ())
    hr : (Fn [(Ref (CurrentHit a) q)] (Ref (Maybe (HitRecord a)) q))
    init : (Fn [(Maybe (HitRecord a))] (CurrentHit a))
    prn : (Fn [(Ref (CurrentHit a) q)] String)
    set-hr : (Fn [(CurrentHit a), (Maybe (HitRecord a))] (CurrentHit a))
    set-hr! : (Fn [(Ref (CurrentHit a) q), (Maybe (HitRecord a))] ())
    str : (Fn [(Ref (CurrentHit a) q)] String)
    update-hr : (Fn [(CurrentHit a), (Ref (Fn [(Maybe (HitRecord a))] (Maybe (HitRecord a)) fq) q)] (CurrentHit a))
}
Defined at line 2, column 11 in '/Users/scottolsen/dev/Carp/hit.carp'
```


scolsen:
Sorry for the long post--I'm trying to document/capture as much as I can so that we don't lose state!

hellerve:
Thanks for this detailed descriütion, that’s extremely helpful!

src/Constraints.hs
```diff
@@ -161,26 +161,22 @@ checkForConflict :: TypeMappings -> Constraint -> String -> Ty -> Either Unifica
 checkForConflict mappings constraint name otherTy =
   let (Constraint _ _ xobj1 xobj2 ctx  _) = constraint
       found = recursiveLookup mappings name
-  in
-    if doesTypeContainTyVarWithName name otherTy
-    then Left (UnificationFailure constraint mappings)
-    else
-      case found of --trace ("CHECK CONFLICT " ++ show constraint ++ " with name " ++ name ++ ", otherTy: " ++ show otherTy ++ ", found: " ++ show found) found of
+  in case found of --trace ("CHECK CONFLICT " ++ show constraint ++ " with name " ++ name ++ ", replaced: " ++ show replaced ++ ", found: " ++ show found) found of
```
eriksvedang, line 164: Correct me if I'm wrong, but I feel like the removed error check here actually did catch an error, namely the unification of something like `(T a) <=> a`. I actually think the problem is that `concretizeType` in Concretize.hs, in the Sumtype case might be the true culprit. That's where it sends both the looked up types to unification (via `instantiateGenericSumtype`) without doing anything to the type variables. They must be renamed using `renameVarTys` or similarly, I believe. What do you think?

scolsen: Thanks for calling this out! I did some investigating, and the check was introduced in commit b1248ac37 back in 2018. Looks like it was introduced to solve an issue with local variable names:

```diff
-*** Defining a local variable messes up unqualified lookup of function with the same name, e.g. (let [words (words &s)] ...). Infinite loop now!
+*** Defining a local variable messes up unqualified lookup of function with the same name, e.g. (let [words (words &s)] ...)
```

After this PR the behavior is the same as after b1248ac37, namely the above still doesn't work, but it won't trigger an infinite unification loop. 

We still perform the check, the difference is we just use it to rename variables with conflicting names instead of returning a unification error, which still prevents the infinite loop (see line 214):

```haskell
replaced = if doesTypeContainTyVarWithName name otherTy
                   then replaceConflicted name otherTy
                   else otherTy
```

I'm fairly sure this ensures behavioral parity, but I could be wrong about that--do you have an example case that might break?

I do think that if there's a better way to solve it elsewhere we should take that route--it'd involve more than sumtypes though, as this particular issue also arises in the case of interfaces. When we return a unification error for that check we can't unify anything that coincidentally shares a variable name, making it impossible to unify something like `interface a -> a` with `implementation (Ref a) -> (Ref a)` which should be unifiable.

scolsen: All that said, it's possible the better approach is to call some function like `renameTyVars` in `Primitives` where we constrain the interface and implementation rather than replacing them here.

Actually, looking into it, it looks like the correct approach here is what you're suggesting, Erik. 

To solve the interface unification problem, we should do something like the following in the interfaces code instead:

```
unifiedInterface = replaceTyVars interface . (unifySignatures interface implementation)
solve interface unifiedInterface
```

Then, as you noted Erik, the original check should stand, and the real issue lies somewhere in sumtypes.

scolsen: I'm going to give the above a try and unsolve #521 which is likely separate issue.

scolsen: > I'm going to give the above a try and unsolve #521 which is likely separate issue.

Ah nope--that still doesn't work since the variables may still be the same. We'd have to work around the check regardless...

scolsen: @eriksvedang I can't find any cases that removing the Type var check effects, but I've added ff0596e which takes a more conservative approach--instead of removing that check, we just bypass it for the special cases of interface registration. This should ensure we don't break anything by accident.

This makes the change more focused, and we'll have to solve #521 elsewhere.

eriksvedang:
@scolsen Thanks, this looks good now. Do you think we will be able to remove the special casing once #521 is solved? (my hunch is that it should be possible...) If so, maybe we should add a coment there about that so we don't forget. I'll comment on ideas for #521 in the thread, separately.

And finally, thanks for finding and fixing this serious problem!

scolsen:
> @scolsen Thanks, this looks good now. Do you think we will be able to remove the special casing once #521 is solved? (my hunch is that it should be possible...) If so, maybe we should add a coment there about that so we don't forget. I'll comment on ideas for #521 in the thread, separately.
> 
> And finally, thanks for finding and fixing this serious problem!

Done! Added a note in f3aa543. We probably can remove the special case once #521 is solved but it will depend on how generic the solution is--at the moment, the issue actually crops up for interfaces when trying to do match something like `a -> a` with `Ref a -> Ref a` since the Ref form contains an `a`--so for this particular case it's not limited to `deftypes`--but the issue might be the same

PR #866, opened 2020-06-09
Meta Refactor 

scolsen:
This change introduces a `Meta` module to abstract away the underlying `Data.Map` representation. Where we previously modified `MetaData` using `Map`, we can now use `Meta` module functions, to ensure caller's manipulation of metadata isn't dependent on the underlying representation.

This should also make it possible to swap out `Data.Map` for some other representation easily (should we ever have to do that, likely not).

### DIFF
--- a/CarpHask.cabal
+++ b/CarpHask.cabal
@@ -17,6 +17,7 @@ library
   hs-source-dirs:      src
   exposed-modules:     Info,
                        Obj,
+                       Meta,
                        Project,
                        Parsing,
                        Infer,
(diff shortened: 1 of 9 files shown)


eriksvedang:
Hooray!
Will do a proper code review later but initial reaction is ⭐️⭐️⭐️⭐️⭐️

jacereda:
Looks good. I wonder if it's worth having something like:

```
data Slot = Doc | Annotations | Sig | Implements | Hidden
```

But I'd leave that for another PR.

PR #868, opened 2020-06-10
Move interfaces code into its own module

scolsen:
**N.B. this PR includes changes from #866 and #867 ** 

This PR moves interface handling code out of Primitives.hs and into its
own module. The actual primitives user's rely on to define and implement
interfaces still live in Primitives.hs, but now all the
registration/internal context mgmt code related to interfaces lives in
its own module.

Hopefully this will make it easier to modify and extend!


eriksvedang:
Very nice cleanups!

PR #870, opened 2020-06-12
Only allow Int as return type for main

jacereda:
With this change the test suite passes using `tcc` as the compiler.

Sits on top of #869.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 if [ -z "$CARP" ]
 then
     if [ -z "$NIX_CC" ]
@@ -7,7 +8,4 @@ then
 	cabal build
     fi
 fi
-if which clang-format > /dev/null
-then
-  clang-format -i core/*.h
-fi
+command -v clang-format >/dev/null && clang-format -i core/*.h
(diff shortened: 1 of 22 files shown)


hellerve:
I’ll have to change a bunch of my libraries, but I think this change is worth it if it means standard-compliance.

bjorn3:
Maybe make Carp add an implicit `return 0;` instead?

jacereda:
I prefer explicit behaviour, and sticking to standard C conventions, but I'm not opposed to having `main` return unit.

Making `main` return 0 implicitly would imply requiring an `exit` call to return an error from an executable. Some people have aversion to having several exit points in a program.

hellerve:
I don’t think implicit behavior is usually a good thing. If we find that it’s too much, we can also add a macro that just appends `0` to the body.

eriksvedang:
I don't think returning Unit really counts as "implicit", it seems like a pretty good indication that you're not interested in reading the return value from the executable after it has run, no?

Why not make main return an int when its type is `(Fn [] Int)` and return 0 when its type is `(Fn [] ())`? This seems to strike a balance between standard compliance and practicality.

eriksvedang:
We can close this now that #881 is merged, right?

PR #874, opened 2020-06-13
RFC: introduce code climate tooling

hellerve:
This PR introduces [ormolu](https://github.com/tweag/ormolu), a Haskell code formatter, to make our style more consistent. While I don’t think it formats using my favorite style, it doesn’t matter to me: what I want is consistency. Having one definitive style guide, and enforcing it in CI, is one step towards a more uniform code base.

To this end, this PR introduces:
- a reformatted code base (I expect that merging this will break any and all pending PRs, so we should proceed with caution and rebase often)
- a step in the Linux CI that enforces that the linter doesn’t complain.

I understand that code formatting and reading is a contentious topic, and as such I submit this as a RFC for us to think and talk about. This is not so much meant as a finished product as a haven for ideas.

What do y’all think?

Cheers

**EDIT**:

After thinking about this some more, and talking about it with other people, I think this PR should introduce more than just a formatter. Ideally, this would be the checklist:

- [ ] Settle on a code formatter, configure it, and add it to CI
- [ ] Settle on a linter (hlint?), configure it, and add it to CI
- [ ] Add a style guide (can be embryonic)


jacereda:
I'm all in in this. I've been using https://hackage.haskell.org/package/brittany for a while, but I really don't care which tool is used as long as I don't have to think again about formatting.

scolsen:
Woo!

hellerve:
Currently the linting step fails with exit code 100, without any diagnostics whatsoever. I’m not quite sure what’s happening: it could be a stack issue, but without any output it’s a little hard to find out what the culprit is.

jacereda:
Could the reason be that the windows workflow can't find the ormolu executable?
Maybe if you add it in the .cabal file as a dependency for, say, the test target it will be present?

hellerve:
Hm, the Windows workflow is not the problem, it’s Linux, actually. But I’ll tinker with the cabal file and try to fix it.

jacereda:
Oh, right, I misread.

Maybe switch it to brittany to see if it's a problem with ormolu itself or the workflow?

hellerve:
I’m playing around with brittany now, and it seems okay, but it might be worth noting that it’s _very_ slow. Reformatting the entire codebase in `src/` takes about 15 seconds (compared to 2 seconds for ormolu).

hellerve:
I found the issue for brittany, it’s lspitzer/brittany#297. I’d like to hold off from merging this until that is fixed, or revert back to ormolu if we can figure that out.

eriksvedang:
Turns out I have quite a few thoughts on this :)

Thanks for getting the ball rolling on this, I agree that the current code base is inconsistent and that fixing it is worthwhile!

I do not share the sentiment that a uniform style (even a "bad" one) trumps anything. Bug fixes, great additions to the compiler, etc. are much more important from my perspective. I believe there's a significant "effort cost" in having everyone follow a certain style (especially if it requires extra tooling) and I'd hate to miss out on contributions just because we have made things unappealing or overly annoying. This does not mean we shouldn't have a code formatter, just that I'd like everyone to keep such considerations in mind.

To me, the biggest consistency concern in the code base right now is not in regards to whitespace (which seems like the only thing these tools address..?) I think a bigger point is the more stylistic choices regarding things like point free style, when to use parenthesis and `$`, when to use `fmap` vs `<$>`, etc. I've been using a fairly "lispy" style when writing the Haskell for this project, primarily to make it easier to jump between the two languages (Haskell and Carp), and because I really like the structural editing made possible when relying more on parenthesis than on `$`. I have let a lot of these differences slide when reviewing contributions because I think it's much more important to have some great code in the project (and to let the other person go work on something else) rather than having them redo a bunch of code – or to start arguing about the merits of either style.

I'm not sure if such matters can (or should) be resolved in a code formatting tool – but perhaps having a style document and trying to adhere to it would be a good idea?

When it comes to whitespace I think that using something like what's introduced in the PR is great. It should be configured to be a lot closer to the style that we've used so far in the project though. That must surely be possible? I actually think our current "whitespace style" is fairly consistent so there shouldn't be such a big change as the diff suggest. The proposed style I see there looks quite obnoxious and hard to write – I know it would really take a toll on me to work with it.

Finally, what would the workflow be for committing / mergeing once this PR is in? Will it reject patches that are not correctly formatted? And how do you install the formatter?

Cheers,
Erik

jacereda:
> Finally, what would the workflow be for committing / mergeing once this PR is in? Will it reject patches that are not correctly formatted? And how do you install the formatter?

I would run it right after `clang-format` in `build.sh`.

jacereda:
> I'm not sure if such matters can (or should) be resolved in a code formatting tool – but perhaps having a style document and trying to adhere to it would be a good idea?

Sure.


> When it comes to whitespace I think that using something like what's introduced in the PR is great. It should be configured to be a lot closer to the style that we've used so far in the project though. That must surely be possible?

I think just changing the columns limit would yield a smaller diff.

eriksvedang:
The three point check list looks great! Not sure if all of it has to be solved by this PR? Perhaps we should start a separate issue for tracking the overall progress?

Running `hlint` as part of the CI sounds good, I think it usually does an amazing job.

Hopefully with some adjustments the code formatter could create a smaller diff.
Which one of the formatters is most likely to suit our needs? I'd like to install it locally.

I think we should try to piggyback the style guide on top of something that a lot of people like and use, probably with some extra additions of our own.

eriksvedang:
I started a separate issue #885 for discussing the style guide since it might be annoying to have to roll that into this PR.

hellerve:
Sounds good! I’d even argue that it’d make sense to get a style guide first, since it will affect how the linter/formatter are to be configured!

eriksvedang:
I've played around a bit with Brittany and Ormolu – I think Ormolu is a lot better in some tricky situations. Does anyone know of any other similar tool to try? (I know of hindent but it doesn't do enough, it seems...)

eriksvedang:
Just found this https://github.com/ennocramer/floskell, might be worth looking into too.

eriksvedang:
I'll close this for now, still interested in figuring it out.

Issue #843, opened 2020-05-25
match-ref tries to delete reference passed to it in catch all branch


TimDeve:
Got an C compiler error a few times while using match-ref,:
```sh
out/main.c:9231:28: error: passing 'Example *' to parameter of incompatible type 'Example'; dereference with *
            Example_delete(wildcard_30);
                           ^~~~~~~~~~~
                           *
out/main.c:6337:29: note: passing argument to parameter 'p' here
void Example_delete(Example p) {
                            ^
1 error generated.
```

```clojure
(deftype Example
 One
 (Two [String]))

(defn main []
  (let-do [ex [(Example.Two @"OKOK")]]
    (match-ref (Array.unsafe-nth &ex 0)
     (Example.Two s) (println* s)
     _ ())))
```
Looking at the C code it seems as if it's trying to delete what's passed in the `match-ref`. Maybe a remnant of `match`.

```C
9206 int main(int argc, char** argv) {
9207     carp_init_globals(argc, argv);                                                                                                                                                                                                                                         9208     /* let */ {
9209         Array _10 = { .len = 1, .capacity = 1, .data = CARP_MALLOC(sizeof(Example) * 1) };
9210         static String _7 = "OKOK";
9211         String *_7_ref = &_7;
9212         String _8 = String_copy(_7_ref);
9213         Example _9 = Example_Two(_8);
9214         ((Example*)_10.data)[0] = _9;
9215         Array__Example ex = _10;
9216         Array__Example* _17 = &ex; // ref
9217         Example* _19 = Array_unsafe_MINUS_nth__Example(_17, 0);
9218         if(_19->_tag == Example_Two_tag) {
9219             Example* _19_temp = _19;
9220             String* s = &_19_temp->Two.member0;
9221             // Case expr:
9222             String _27 = String_str(s);
9223             String* _28 = &_27; // ref
9224             IO_println(_28);
9225             String_delete(_27);
9226         }
9227         else if(true) {
9228             Example* _19_temp = _19;
9229             Example* wildcard_30 = _19_temp;
9230             /* () */
9231             Example_delete(wildcard_30);
9232         }
9233         else {
9234           // This will not be needed with static exhaustiveness checking in 'match' expressions:
9235           fprintf(stderr, "Unhandled case in 'match' expression at line 7, column 5 in '/mnt/c/Users/tym0/dev/sandbox/match-ref/main.carp'\n");
9236           exit(1);
9237         }
9238         Array_delete__Example(ex);
[...]

PR #876, opened 2020-06-13
Attempt at fixing #843

leblowl:
I am new to Haskell. Just met @hellerve at ZuriHac today and they pointed me to this issue: https://github.com/carp-lang/Carp/issues/843. My solution appears to fix 843 from the perspective that the example code provided in the issue now runs without exception (the generated C code no longer contains a call to `Example_delete` in the catch all block of code generated for `Match`), however I don't really understand why a deletion was taking place to begin with. I am trying to understand why a deletion may have made sense at this point... not sure if my fix breaks another test case. Thanks!

Related commit (which added the delete call): https://github.com/carp-lang/Carp/commit/c80d6429202f8437445afded85b4fafa0997b461

### DIFF
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -371,7 +371,6 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
                        when isNotVoid $
                          appendToSrc (addIndent indent' ++ retVar ++ " = " ++ caseExprRetVal ++ ";\n")
                        let Just caseLhsInfo' = caseLhsInfo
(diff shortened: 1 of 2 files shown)


hellerve:
I assume that the call to delete is there because for a regular `match`, this is exactly correct: it consumes the value. When we introduced `match-ref` later, we just forgot to add an extra check, and since this piece of code treats both, we need to watch out.

Thank you for fixing this!

eriksvedang:
Awesome, @leblowl !

PR #878, opened 2020-06-14
Add a doc capturing common patterns

scolsen:
The first pattern is one I've found handy for returning some data
structure that contains the results of running two different
ownership-taking functions on some argument.

### DIFF
--- a/docs/patterns.md
+++ b/docs/patterns.md
@@ -1,0 +1,42 @@
+# Carp Patterns
+
+This doc captures some common programming patterns in Carp.
+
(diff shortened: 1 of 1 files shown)


scolsen:
Ah, definitely! Sorry—wrote this late last night and totally didn’t think of the context and voicing issues—I’ll remove the `I`s!

I’m also open to hosting this elsewhere if we’d like to just keep the main repo clean and document patterns elsewhere.

eriksvedang:
Neat idea!

Issue #807, opened 2020-05-15
info should display the implemented interface


hellerve:
Now that interfaces can be implemented by functions that have different names, it would be great if we could show the interface(s) a function implements when you type `info`.

Cheers

PS: a function cannot implement multiple interfaces, right?

scolsen:
Re: the P.S., they actually can implement more than one if the interface signatures are equivalent, but not if they're not equivalent.

```
鲤 (defn foo [x y] y)
鲤 (implements + foo)
鲤 :i +
+ : (Fn [a, a] a) = {
   ... bunch of impls
    foo
}
鲤 (implements - foo)
鲤 :i -
- : (Fn [a, a] a) = {
    ... bunch of impls
    foo
}
鲤 (implements zero foo)
[INTERFACE ERROR] foo : (Fn [a, b] b) doesn't match the interface signature (Fn [] a)
[INTERFACE ERROR] foo : (Fn [a, b] b) doesn't match the interface signature (Fn [] a) at REPL:1:13.

Traceback:
(implements zero foo) at REPL:1:1
```

which seems reasonable to me--it sort of reflects the idea of shared properties. e.g. any monoidal operation can be used as `+` or `*`

hellerve:
Yeah I think this makes sense! I just wasn’t sure whether `implements` override the meta information. :)

scolsen:
> Yeah I think this makes sense! I just wasn’t sure whether `implements` override the meta information. :)

Great point, actually, it does! The additional implements aren't captured in the meta--it'll just be set to the last interface `implements` was called with--but the registrations still happen so that's why the calls above still report the function as an implementation.

Probably the `implements` meta should take a list and accumulate the interface args, that way we can just use the meta to print the implemented interfaces rather than performing lookups.

hellerve:
That sounds reasonable to me, and there is actually a precedent for this: `annotate` works that way!

https://github.com/carp-lang/Carp/blob/ed4c4431dcc39bcfc56bb8303ddf940ecd7f1536/core/Macros.carp#L48-L50

PR #880, opened 2020-06-14
include implemented functions in info

TheAceShinigami:


hellerve:
This PR fixes #807!

src/Primitives.hs
```diff
@@ -286,6 +286,12 @@ primitiveInfo _ ctx [target@(XObj (Sym path@(SymPath _ name) _) _ _)] = do
           case Meta.get "doc" metaData of
             Just (XObj (Str val) _ _) -> liftIO $ putStrLn ("Documentation: " ++ val)
             Nothing -> return ()
+          case Meta.get "implements" metaData of
+            Just xobj@(XObj object info _) -> do
+              case info of
+                Just info' -> putStrLn $ "Implementing: " ++ getName xobj
```
hellerve, line 292: This currently prints a list, i.e. `:i Int.+` will print `Implements: (+)`, and if there are multiple it would be formatted as `(interface-a interface-b ...)`. To me, this is acceptable, but if we wanted we could make this a comma-separated list?

Issue #879, opened 2020-06-14
Lambdas need to capture from implicit environments


hellerve:
After the work done by @Vertmo in #850 we can finally refer to variables inside nexted contexts in `fn`. Sadly, we forgot to add support for implicit environments, i.e. `let`s. This will lead to weird bugs.

Consider this (very silly) example:

```clojure
(defn make-cb []
  ((fn []
    (let [x "hi"]
      (fn [] (IO.println x))))))

(defn main [] ((make-cb)))
```

Here, we have two nested `fn`s, the inner of which depends on `x`, introduced in a let in the outer `fn`. This should be fine, except that `x` propagates into the lookup list of the outer `fn`, because implicit environments aren’t checked by the current code. This will lead to invalid C being emitted.

What needs to happen is that a check for implicit environments needs to be inserted into the lookup routine, so that we can short-circuit before bubbling those lookups up.

Cheers

Vertmo:
Oh yeah, that's an issue for sure. Since I was the last one to touch this part of the code, I will be looking into it, if that's ok with you ^^

hellerve:
That’s perfect, thank you! I looked very briefly yesterday, but since it wasn’t a quick fix I decided to defer until I find the time, but I think you’re even better suited than I am to work on this!

Vertmo:
This issue was resolve by #883 and should probably be closed :)

PR #883, opened 2020-06-15
Fixed lambda capture in presence of let bindings

Vertmo:
This should be fixing #879 
As described, my changes in #850 didn't support implict environments, causing error in the calculation of the set of captured variables for a lambda.
To fix this issue, I have added a special case in `collectCapturedVars`; when encountering a `Let` construct, we filter the identifiers defined in the bindings out of the variables captured by the body, as well as the following bindings (where they could be used).

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -326,20 +326,18 @@ collectCapturedVars root = removeDuplicates (map decreaseCaptureLevel (visit roo
                                                     else LookupLocal (Capture (n-1))))
       (Just dummyInfo) ty
 
-    pairBindings :: [XObj] -> [(XObj, XObj)]
-    pairBindings [] = []
-    pairBindings (k:v:tl) = (k, v):pairBindings tl
-
     visit xobj =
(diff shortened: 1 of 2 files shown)


src/Concretize.hs
```diff
@@ -326,14 +326,26 @@ collectCapturedVars root = removeDuplicates (map decreaseCaptureLevel (visit roo
                                                     else LookupLocal (Capture (n-1))))
       (Just dummyInfo) ty
 
+    pairBindings :: [XObj] -> [(XObj, XObj)]
+    pairBindings [] = []
+    pairBindings (k:v:tl) = (k, v):pairBindings tl
```
hellerve, line 331: There is a function in `Util` called `pairwise` that does this.

Vertmo:
@hellerve done ^^

Vertmo:
Just a small rebase I forgot to do earlier to make sure the future merge wont break anything :)

eriksvedang:
Epic!

PR #887, opened 2020-06-17
Allow "phantom constructors"

### LINKED ISSUE #886: Type inhabitants of polymorphic constructors can't be used as type inhabitants
Currently, telling carp:

```
(deftype (Index (f a) b) [value a at b])
```

Will produce the error:

```
Invalid type definition for 'Index':

I can’t use the type `a` as a member type at line 4, column 38 in '/Users/scottolsen/dev/dimensions/src/naperian.carp'.

Is it defined and captured in the head of the type definition?
```

Using the full `(f a)` works fine, as expected, but we should also be able to refer to `a` in this case.

scolsen:
Prior to this commit, a type such as: `(deftype (Foo (f a) [bar a]))`
was rejected by the compiler. However, such a type is sensible--Foo
ranges over some container type `f`, and it is essentially Foo's job to
provide a view into `f`, granting access to `a` and whisking `f` away
via the phantom constructor.

After this change, this type definition is now valid.

Note that one often needs to wrap phantom constructor cases in `the`
unless the type can be determined by the overarching context:

```
(deftype (Foo (f a) [bar a]))
(the (Foo (Maybe Int)) (Foo.init 1))
;; => (Foo 1)
```

fixes #886

### DIFF
--- a/src/Validate.hs
+++ b/src/Validate.hs
@@ -80,7 +80,12 @@ canBeUsedAsMemberType typeEnv typeVariables t xobj =
                            else Left (UninhabitedConstructor ty xobj (length tyvar) (length vars))
                          _ -> Left (InvalidMemberType ty xobj)
              _ -> Left (InvalidMemberType t xobj)
-    VarTy _ -> if t `elem` typeVariables
+    VarTy _ -> if foldr (||) False (map (isCaptured t) typeVariables)
                then return ()
                else Left (InvalidMemberType t xobj)
+               where
+                 -- If a variable `a` appears in a higher-order polymorphic form, such as `(f a)`
+                 -- `a` may be used as a member, sans `f`.
+                 isCaptured t v@(VarTy _) = t == v
+                 isCaptured t (StructTy (VarTy v) vars) = any (== t) vars
     _ -> Left (InvalidMemberType t xobj)


scolsen:
An example that's more practical and illustrates the context helping the compiler out:

```
(sig from-maybe (Fn [(Maybe a)] (Foo (Maybe a)))
(defn from-maybe [m] (match m (Maybe.Nothing) (Foo.init (zero)) (Maybe.Just x) (Foo.init x)))
(from-maybe (Maybe.Just 1))
(Foo 1)
```

There's no need for `the` here but do note that it does require a `sig` to work. Generally speaking, the compiler will keep `(f a)` generic more often than not, so these types over "phantom constructors" usually need some help from explicit type annotations and aren't completely inferable.

scolsen:
Also, for the curious, the case that motivated such a signature for me was a type level abstraction of indicies:

```
(deftype (Index (f a) b) [value a at b])
```

The polymorphic constructor, `f`, is important since this representation of indicies must range over container types only and those containers should support `fmap` and be functors (thus the type must not be of base kind). However, all we care is that the type it ranges over is constructible--we don't actually care about the constructor itself, so there's no need for `Index` to carry along the constructor information beyond the type level.

scolsen:
It's also fun to note that if we have two functors that implement `fmap`, `G` and `F`, types like `(deftype (G (f a)) (G a)))` are a nice and compact way of hinting at a natural transformation between two functors

hellerve:
I learned so many fun facts in this thread, thank you!

eriksvedang:
Great addition, thanks!

PR #888, opened 2020-06-17
core: add Phantom

hellerve:
This PR adds `Phantom` to core, which is a type that has no inhabitants but a type variable. They enable us to express type level constraints in code. Consider the example I added to the tests.

```clojure
(defn m [a] (Phantom.ensure a (the (Phantom Int) (Phantom.init))))

(println* (m 123l))
```

This will not unify, since `m` is passed a `Long`, not an `Int`, as required by the phantom. Wait, but don’t we already have `sig` and `the` to express such constraints? Yes, but Phantom types can be passed around like data, making those `ensure`s more expressive.

They also exist [in Rust](https://doc.rust-lang.org/std/marker/struct.PhantomData.html), for borrow-checking purposes, and we could add that sort of functionality if we ever add lifetime annotations, but for now they are useful as is.

Cheers

### DIFF
--- a/core/Phantom.carp
+++ b/core/Phantom.carp
@@ -1,0 +1,6 @@
+(deftype (Phantom a) [])
+
+(defmodule Phantom
+  (sig ensure (Fn [a (Phantom a)] a))
+  (defn ensure [a p] a)
+)
--- a/core/Phantom.carp
+++ b/core/Phantom.carp
@@ -1,6 +1,22 @@
+(doc Phantom "is a type without inhabitants and a free type variable, which is
+a fancy way of saying that it has no runtime representation.
+
+When trying to mark types programmatically using something like `the` but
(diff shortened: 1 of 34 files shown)


scolsen:
Woo, awesome! My only comment is that we should add a `doc` for `ensure` and probably for `Phantom` too, since beginners may not be familiar with this type-level super power. I think we can do that in a follow-up PR though.

hellerve:
I added docs for `Phantom`, I hope they make sense. I have trouble communicating why this is interesting in a clear, beginner-friendly manner.

scolsen:
> I added docs for `Phantom`, I hope they make sense. I have trouble communicating why this is interesting in a clear, beginner-friendly manner.

I think it's actually very clear!

eriksvedang:
Very cool!

PR #890, opened 2020-06-17
Add the Opaque type

scolsen:
The opaque type is an uninhabited type with no constructors.  Opaque can
be used to force some abstract type to range over a type constructor
without concerning oneself with the inhabitant of the constructor--in
other words, it may be used to enable a type to work for all inhabitants
and can express relationships between types. It can facillitate generic
programming.

Consider an example:

```
;; The type of indicies over containers of a single type argument
(deftype (Index (f Opaque) b) [at b])

(definterface tabulate (Fn [(Ref (Fn [(Index (f Opaque) b)] c))] (f c)))
(definterface positions (f (Index (f Opaque) b)))

(implements tabulate tabulate)
(defn tabulate [h]
  (fmap h @&positions))

(deftype (Tuple a) [x a y a])

(defmodule Tuple
  (sig positions (Tuple (Index (Tuple Opaque) Bool)))
  (def positions (Tuple.init (Index.init true) (Index.init false)))
)
```

In the above example, the Opaque type allows us to define tabulate
generically defined over Tuples without having to ensure their
inhabitants match, allowing us to fully determine the resulting values
type via tabulate's function argument. Without Opaque, Index would
contain a generic type which would be unreseolved upon the call to
`tabulate`. It allows us to ensure the `positions` we call are the
positions of the correct constructor type `f` wihtout worrying having to
restrict ourselves to only calling `Indexes` over an `f` of a specific
type  (e.g. `(f a)`)--in other words, it allows us to constrain
functions by constructor types only.

Thanks to Opaque, tabulate can generate an `(Array Int)`, `(Array
Bool)`, `(Array String)` all solely dependent on the return type of `h`.

### DIFF
--- a/core/Opaque.carp
+++ b/core/Opaque.carp
@@ -1,8 +1,8 @@
 (doc Opaque "The opaque type is an uninhabited type with no constructors.
 Opaque can be used to force some abstract type to range over a type constructor
-without concerning oneself with the inhabitant of the constructor--in other
-words, it may be used to enable a type to work for all inhabitants and can
-express relationships between types. It can facillitate generic programming.
+without concerning oneself with the inhabitant of the constructor. It may be
+used to enable a type to work for all inhabitants and can facillitate generic
+programming.
 
 Consider an example:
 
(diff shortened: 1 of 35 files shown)


scolsen:
I'm not married to the name `Opaque` for this type, so if you have any other ideas, lmk! We could also use `Ur` `Proxy` `Surrogate` etc.

scolsen:
The example might be a little odd looking--the basic idea is that we can define index types for some container, fill a container with its indexes, then replace all indexes with values at that index. In the example here we use true and false since `Tuple` can only ever have two positions and `true` and `false` are conveniently manipulated with ifs/

An function that we could pass to tabulate might be something like:

```
鲤 (defn fill-up-to [i] (if (= true @(Index.at &i)) 1 2))
(tabulate &fill-up-to)
(Tuple 1 2)
```

Which will produce a `Tuple Int` without the need for any `the` declarations or unresolved generic types since the type of `positions` is well-defined over the `Opaque` type and the inhabitant type is strictly determined by the return type of the function.

core/Opaque.carp
```diff
@@ -0,0 +1,41 @@
+(doc Opaque "The opaque type is an uninhabited type with no constructors.
```
hellerve, line 1: Could we load this file in `docs/core/generate_core_docs.carp` and add it to the documentation generation?

scolsen: Done, thanks! And I also re-generated the docs. I always forget to do that -.-

eriksvedang:
Needs a rebase @scolsen !

As for the name, I think it's good.

scolsen:
rebased!

PR #892, opened 2020-06-18
Added some of GLFW constants

Vertmo:
I started playing around with OpenGL. As I'm a beginner, my main goal is to implement [this tutorial](http://www.opengl-tutorial.org/) in Carp.
As this tutorial is using GFLW, I have quickly realised that none of the constants defined in *gflw3.h* have bindings in the GFLW module.
I have started adding some of these missing constants (mainly the one I needed to be honest). Would you like me to add the rest ?

### DIFF
--- a/core/GLFW.carp
+++ b/core/GLFW.carp
@@ -2,6 +2,7 @@
 (add-pkg "glfw3")
 
 (defmodule GLFW
+  ;; Methods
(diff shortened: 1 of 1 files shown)


eriksvedang:
Sounds like a really fun project @Vertmo, adding the rest would be great!

PR #893, opened 2020-06-19
Implement Privacy Checking 

scolsen:
This implements private by adding a simple check to the evaluator and expander. If a
binding is found in the global context, we check if it's marked as
private. If so, we inform the user that it can't be called from outside
of its module.

Note that we don't perform this check if the binding is found in the
internal env, since that means it's a function called within the same
module and thus is ok (even if marked as private).

After this change, something like the following works, granting us
proper encapsulation:

```
;; File Foo.carp
(deftype Foo [bar Int])

(private Foo.bar)

(defmodule Foo
  (defn get [foo]
    (Foo.bar foo))
)

;; Enter REPL
(load "Foo.carp")
(Foo.bar &(Foo.init 1))
The binding: Foo.bar is private; it may only be used within the module
that defines it. at REPL:1:2.
@(Foo.get &(Foo.init 1))
1
=> 0
```

N.B. I also had to remove a private declaration from fmt-internal--this
declaration didn't really make much sense anyway, as fmt-internal is a
global function, so module-based privacy is not enforceable.

### DIFF
--- a/core/Format.carp
+++ b/core/Format.carp
@@ -1,4 +1,3 @@
-(private fmt-internal)
 (hidden fmt-internal)
 (defndynamic fmt-internal [s args]
   (let [idx (String.index-of s \%)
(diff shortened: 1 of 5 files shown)


eriksvedang:
AWESOME!

PR #894, opened 2020-06-21
Added add-c to include additional compilation units in the compiler i…

jacereda:
…nvocation.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -71,6 +71,7 @@ defaultProject =
           , projectForceReload = False
           , projectPkgConfigFlags = []
           , projectCModules = []
+          , projectLoadStack = []
           }
 
 -- | Starting point of the application.
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -70,6 +70,7 @@ defaultProject =
           , projectGenerateOnly = False
           , projectForceReload = False
           , projectPkgConfigFlags = []
+          , projectCModules = []
           }
 
 -- | Starting point of the application.
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -707,7 +707,9 @@ loadInternal ctx xobj path i reloadMode = do
         _ <- liftIO $ setCurrentDirectory cur
         doGitLoad path fpath
       else do
-        (x0, _, stderr0) <- liftIO $ readProcessWithExitCode "git" ["clone", "--recurse-submodules", path, "."] ""
+        _ <- liftIO $ readProcessWithExitCode "git" ["init"] ""
+        _ <- liftIO $ readProcessWithExitCode "git" ["remote", "add", "origin", path] ""
+        (x0, _, stderr0) <- liftIO $ readProcessWithExitCode "git" ["fetch", "--all", "--tags"] ""
         case x0 of
           ExitFailure _ -> do
             _ <- liftIO $ setCurrentDirectory cur
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -638,9 +638,14 @@ loadInternal ctx xobj path i reloadMode = do
                           files' = if canonicalPath `elem` (map fst files)
                                    then files
                                    else files ++ [(canonicalPath, reloadMode)]
-                          proj' = proj { projectFiles = files', projectAlreadyLoaded = canonicalPath : alreadyLoaded }
+                          prevStack = projectLoadStack proj
+                          proj' = proj { projectFiles = files'
+                                       , projectAlreadyLoaded = canonicalPath : alreadyLoaded
+                                       , projectLoadStack = canonicalPath : prevStack
+                                       }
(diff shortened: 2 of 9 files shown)


jacereda:
The name is somewhat misleading, since you can use it to include C++ or ObjC modules.

jacereda:
Maybe the additional compilation units should just be a `Project` field?

eriksvedang:
> Maybe the additional compilation units should just be a Project field?

Yeah I think that makes sense!

TimDeve:
Oh, I've been wanting this for a while, your way makes a lot more sense than what I had in mind ^^

hellerve:
> How about `add-native-source` for a name?

I feel like `native` is a bit of an overloaded word, and keeping `c` somewhere in the name might make sense?

TimDeve:
I gave this a try and it seems to be outputting path relative to the source file, would there be a way for this to output absolute path instead so it can be used in libraries?

jacereda:
@TimDeve can you check if this works better for you?

TimDeve:
Thanks for taking the time to update it. The latest change are closer but still wouldn't work for libraries. Here's a quick program I made to demonstrate:
```clojure
(defndynamic test-add-c [relpath]
  (eval (Dynamic.String.concat [(Dynamic.String.absolute (Dynamic.String.directory (file)))
                                "/"
                                relpath])))

(defn main [] (println* (test-add-c "something.c")))
```

It works if where you're invoking the compiler is where the file is:
```sh
~/d/s/file ❯❯❯ ls
main.carp   out         something.c
~/d/s/file ❯❯❯ carp -x main.carp
/Users/tim/dev/sandbox/file/something.c
```

But not if the source file is in a different folder than where the compiler is invoked (in the case of library for example):
```sh
~/d/s/file ❯❯❯ cd ..
~/d/sandbox ❯❯❯ ls file
main.carp   out         something.c
~/d/sandbox ❯❯❯ carp -x file/main.carp
/Users/tim/dev/sandbox/something.c
```

Ideally it would work like `relative-include` which creates an absolute path from the path of the carp source file.

It would be neat to have a `Dynamic.Path.current-file` utility, as it could be reused for tons of different things.

jacereda:
@TimDeve I guess you're right, I thought `file` was `current-file`.

As for `Dynamic.Path` I guess we need that. In particular, the `absolute` in this PR probably belongs there.

jacereda:
@TimDeve yet another attempt. This time I implemented a `load-stack` command because I didn't find a better way. It also adds the `Dynamic.Path` module.

jacereda:
And looks like I messed it with another change somehow. 

I had a stash for cloning modules with submodules. Do you want me to submit that as another PR?

TimDeve:
Thanks for taking the time to tweak this. It's awesome, it'll allow me to put stuff in libraries when I couldn't before.

src/Eval.hs
```diff
@@ -702,9 +707,7 @@ loadInternal ctx xobj path i reloadMode = do
         _ <- liftIO $ setCurrentDirectory cur
         doGitLoad path fpath
       else do
-        _ <- liftIO $ readProcessWithExitCode "git" ["init"] ""
-        _ <- liftIO $ readProcessWithExitCode "git" ["remote", "add", "origin", path] ""
-        (x0, _, stderr0) <- liftIO $ readProcessWithExitCode "git" ["fetch", "--all", "--tags"] ""
+        (x0, _, stderr0) <- liftIO $ readProcessWithExitCode "git" ["clone", "--recurse-submodules", path, "."] ""
```
eriksvedang, line 710: Is this change equivalent?

app/Main.hs
```diff
@@ -70,6 +70,8 @@ defaultProject =
           , projectGenerateOnly = False
           , projectForceReload = False
           , projectPkgConfigFlags = []
+          , projectCModules = []
+          , projectLoadStack = []
```
eriksvedang, line 74: What's the purpose of this?

src/Eval.hs
hellerve: This actually assumes that this is cloned for the first time. It shouldn’t be equivalent?

eriksvedang: That's what I thought too...

jacereda: That was accidentally included in this PR, I'll send a separate PR to checkout submodules.

app/Main.hs
jacereda: `projectCModules` is the list of modules to be included in the compilation command.
`projectLoadStack` is the list of `.carp` files being loaded so far. It's used in `Dynamic.carp:current-file` to get the file currently being loaded.

jacereda:
Any idea regarding the MacOS build failure?

hellerve:
This has been happening since (at least #903). Since we didn‘t change anything in the CI pipeline I assume this is related to Github CI and Haskell actions somehow, but I haven‘t had time to dig into it yet.

PR #895, opened 2020-06-22
FIX #811: Make path relative to the invoked file in the read-file macros

### LINKED ISSUE #811: Behavior of `read-file` is inconsistent with `load`
`read-file` behavior is inconsistent with `load`'s as the path is relative to where the compiler is ran rather than relative to which file the function is invoked in.

VPashkov:


src/Commands.hs
```diff
@@ -808,8 +808,14 @@ commandNot ctx [x] =
 commandReadFile :: CommandCallback
 commandReadFile ctx [filename] =
   case filename of
-    XObj (Str fname) _ _ -> do
-         exceptional <- liftIO ((try $ slurp fname) :: (IO (Either IOException String)))
+    XObj (Str fname) i _ -> do
+         let relativeTo = case i of
```
VPashkov, line 812: This is a copypaste from the [load](https://github.com/VPashkov/Carp/blob/7e440cae8ce1a40322fdd0e7db20b08c299ef429/src/Eval.hs#L612) command.
I'm not sure where to put it to keep it DRY. Or maybe it's OK to leave it here?

hellerve: You could move this code out into a helper function, for instance:

```haskell
getRelativeOf i =
  case i of
    Just ii ->
      case infoFile ii of
        "REPL" -> "."
        file -> takeDirectory file
    Nothing -> "."
```

And then call that there. I don’t mind the duplication too much, though!

eriksvedang:
Thanks for the PR, looks good! 

I have a slight worry there is a way to break this, so I need to try that before merge:ing.

eriksvedang:
Will take a look at it in the weekend, sorry for being slow.

eriksvedang:
Thanks again for the PR @VPashkov!

I finally got around to assert my suspisions and unfortunately they were right. This code takes the path of the file where the *argument* to `load-file` is defined, not the location of the call to `load-file`. This can lead to very hard to find bugs I'm afraid. Unfortunately there's currently no way to get hold of the `Info` object for the call since it's not part of the `CommandCallback` type, which is all that the commands get access to.

To solve this, we could add an additional info to the `CommandCallback` type (but it would of course go unused in most cases). Or maybe there's another macro-oriented way to write it? Any ideas are most welcome!

eriksvedang:
Pruning PR:s and since this one has a bug I'll close it for now, please re-open if you can find a way around the problem I described above!

PR #896, opened 2020-06-23
Add unreachable

hellerve:
This PR adds the macro `unreachable`, which asserts that a branch cannot be reached, e.g. in a match.

As this is currently a macro, it does not help type checking. I’d assume that it’d be nice if it were a feature of the language and its return type would match all types, allowing us to ignore a branch properly. To simulate this, the current version appends a faux `zero`, which will never be called; this means that `unreachable` can be used for any type that implements `zero`. Ideally it could stand in for any type, though (but that’s not possible from within the language).

Cheers


hellerve:
To simplify the tests, I also added `Bool.zero`, which returns `false`. This is consistent with e.g. [Rust](https://doc.rust-lang.org/std/default/trait.Default.html#tymethod.default).

eriksvedang:
How about defining `bottom` to use instead of `zero`, along these lines:

```clojure
(deftemplate bottom (Fn [] a) "$a $NAME()" "$DECL { return ($a)NULL; }")
```

hellerve:
@eriksvedang Sadly, this might lead to the emission of invalid C, since you can’t cast `NULL` to e.g. a `Maybe`.

eriksvedang:
Right.... But maybe we can do some other trick, like calling abort in the function..?

hellerve:
I just made that change; it seems to work, but I’m very unsure how standards-compliant this is—I assume it isn’t at all. Maybe @jacereda has some insight into this?

jacereda:
I'm not sure, but I tried the PR with `zig cc` and `tcc` and both seem happy with it. The reason might be that `abort` is usually annotated with a `noreturn` attribute, but I might be wrong.

hellerve:
Hm, maybe we could simplify further by just adding `noreturn` to an empty function then.

jacereda:
I wouldn't use `noreturn`, how to do that depends on the compiler (attributes in gcc and friends, __declspec__ in cl...). Just let the headers do their job. I would merge as is.

PR #900, opened 2020-06-30
Add support for cross-compilation.

jacereda:


jacereda:
Rebased on top of current master.

src/StartingEnv.hs
```diff
@@ -220,6 +220,7 @@ dynamicModule = Env { envBindings = bindings
                     , addCommand (SymPath path "load") 1 commandLoad "loads a file into the current environment." "(load \"myfile.carp\")"
                     , addCommand (SymPath path "load-once") 1 commandLoadOnce "loads a file and prevents it from being reloaded (see `reload`)." "(load-once \"myfile.carp\")"
                     , addCommand (SymPath path "expand") 1 commandExpand "expands a macro and prints the result." "(expand '(when true 1)) ; => (if true 1 ())"
+                    , addCommand (SymPath path "arch") 0 commandArch "prints the native architecture (as returned by the Haskell function `System.Info.arch`)." "(arch)"
```
hellerve, line 223: I think `target` is missing here?

src/StartingEnv.hs
```diff
@@ -270,6 +271,7 @@ dynamicStringModule = Env { envBindings = bindings
                                 , addCommand (SymPath path "slice") 3 commandSubstring "creates a substring from a beginning index to an end index." "(String.slice \"hello\" 1 3) ; => \"ell\""
                                 , addCommand (SymPath path "length") 1 commandStringLength "gets the length of a string." "(String.length \"hi\") ; => 2"
                                 , addCommand (SymPath path "concat") 1 commandStringConcat "concatenates a list of strings together." "(String.concat [\"hi \" \"there\"]) ; => \"hi there\""
+                                , addCommand (SymPath path "split-on") 2 commandStringSplitOn "split a string at separator." "(String.split-on \"-\" \"hi-there\") => [\"hi \" \"there\"]"
```
hellerve, line 274: There should probably be a `;` before the `=>` in the code example.

src/Commands.hs
```diff
@@ -456,6 +460,16 @@ commandOS :: CommandCallback
 commandOS ctx _ =
   return (ctx, (Right (XObj (Str os) (Just dummyInfo) (Just StringTy))))
 
+-- | Command for getting the native architecture.
+commandArch :: CommandCallback
+commandArch ctx _ =
+  return (ctx, (Right (XObj (Str arch) (Just dummyInfo) (Just StringTy))))
+
+-- | Command for getting the name of the operating system you're on.
```
hellerve, line 468: Isn‘t it for the operating system youre compiling for rather than the one youre on?

eriksvedang:
Thanks for the review @hellerve!

src/StartingEnv.hs
eriksvedang: Missing from where?

hellerve: It‘s defined as a primitive, but not wired up in the registration here, unless I‘m missing something?

src/Commands.hs
jacereda: `target-triple` and its accessors refer to the target. I didn't want to mess with `os` and added `arch` with that name, but they should probably be renamed to `host-arch` and `host-os`. I can do that in this PR if you want, didn't do so because `os` might be used in a bunch of places.

src/StartingEnv.hs
jacereda: Oh, right. Fixed.

src/Commands.hs
hellerve: Ah, right. I think that‘s fine?

jacereda:
Would it make sense to have host/target stuff in modules? Something like`Target.os` and `Host.os`?

jacereda:
I've squashed the commits and rebased on top of current master.

eriksvedang:
I'm sorry I dropped the ball on this one. Would be happy to merge once rebased (again).

jacereda:
Rebased.

eriksvedang:
@jacereda Thanks a lot! Unfortunately this fails on Windows, do you know why?

jacereda:
No idea. Looks like it failed prior to building Carp itself.
It compiles fine on Windows for me.

eriksvedang:
It worked great on my Windows machine, so I think we're good!

PR #901, opened 2020-07-01
Honor project logo and url during module doc gen

scolsen:
Previously, some of the documentation generations specific to Carp's
core library doc generation were hardcoded in renderdocs, thus, even
though users could set their own logo and url, these weren't honored
during the generation of module documentation, resulting in a bunch of
docs pointing to logos and urls that weren't intended.

This change updates the rendering code to honor the url and logo set in
the user's Project config.


eriksvedang:
Thanks, great fix!

Issue #898, opened 2020-06-29
Project.config


Vertmo:
While tinkering with the compiler (related to #889) I've run into a problem : when giving a title to the core library project via
`(Project.config "title" "core")`, the config is propagated to any code loading it (that is any code compiled without the *--no-core" option. It means that any project that does not set it's own project title would end up named "core" instead of "Untitled". In particular, this is a problem for the compiler test suite, which fails because the executable that are ran are not the one expected.

Now, this last issue is not that big, as the fix is simply to not title "core", as was done previously. However this behavior also means that any titled library that is loaded by a user after calling `Project.title` would override the name of the user's project, which may be counter-intuitive.

What I propose is that some or all of a project's config should only be changeable in the project's "main" file (that is the one on which carp is called first when compiling in batch mode, or in the REPL when using it. What do you think ?

jacereda:
I don't know if it matters, but you can also set it via `--eval-preload '(Project.config "title" "xxx")'` and via `profile.carp`.

scolsen:
I've definitely run into this before but I'm not sure about the best way to fix it yet.

Issue #889, opened 2020-06-17
Separate Compilation


Vertmo:
##### Disclaimer : maybe I've missed something and this is already in the work, but I haven't been able to find anything in the issues / PRs. Sorry if this issue is stupid/redundant

As I explore and write more with the language, one drawback of the current compiler is apparent : it does not support separate compilation and linking, which mainly means that any used library are compiled from source at the same time as the program. This makes the compilation of any program slower (both for the *carp* compiler and the *c* compiler which has to handle a big C file), and I imagine it will get slower as the size of the standard library increases.

A solution to this problem is to allow for separate compilation : that is allowing to compile a unit of code (let's say a file) to an intermediary representation that can be linked to the complete program. This will allow to compile the core library (or any other) once, and being able to link it when using it into other programs.

Here is my precise proposal for carp: when compiling a project, add an option `--as-lib` to compile as a linkable library. The compiler would then generate three artifacts:
* a *.o* file containing the implementation of the library (compiled as normal, the only difference beeing asking the c compiler to not link it)
* a *.h* interface file that will be used by the C compiler when treating a program using the library
* a *.carpi* (?) interface file that will be used by the Carp compiler when treating a program using the library. This last one will mainly contain the types of the values defined in the source file. I'm not yet sure what format this one would have (text or binary). The second one would probably be more efficient

These artifacts could then be used when loading a compiled library, with a directive such as `(static-load "lib.carpi")` which would tell the compiler to:
* use the *lib.carpi* interface when type/borrow-checking the current program
*  `#include "lib.h"` in the generated c file
* ask the C compiler to link *lib.o* when linking the final executable

Now there may be something I'm missing (in particular regarding the interactions between the repl and a statically compiled library), but this seems doable (although it would need a bit of work). Please tell me what you think ^^

hellerve:
I think this is a great idea, and something that makes a lot of sense. I would definitely want to keep completely static compilation as an option, simply becausde you don’t always have `.o` files etc. lying around, especially for cross-compilation, and having a single artifact is a little more convenient then.

But I see no problem making the two co-exist—it’s just a bunch of work to add that, and I think that’s why we stayed away from it until now. The current approach, while suboptimal for compile times, “works”.

I’d love to investigate all kinds of avenues to make the compiler faster, though, and separate compilation seems like a good way to get faster compile times, easily.

jacereda:
[...]

PR #903, opened 2020-07-05
Separate compilation

Vertmo:
I started working on separate compilation, as described in #889 
My current schemes allow for compiling (non-polymorphic) functions, constants and types present in a library. This generates three files:
* a *.o* file containing the c-compiled object code for the definitions
* a *.h* file containing the c-level definitions to be used by the C compiler
* a *.carpi* file containing an interface used by the compiler when loading the library. All the objects are loaded as `external` or `external-types`, which makes the most sense given that, from the point of view of the user, they are indeed externals defined by the *.o* and *.h* file

This also changes the way `Project.title` can be set, as described in #898 , as I felt that the title should only be set in the main file of a project (in the case of this PR, I set a title for the core library, since I need titles to handle some things during linking).

What am I missing here ? Mostly polymorphic functions and macros, which cant be compiled to C-code. What I plan to do is simply leave them "as-is" in the .carpi file, to be compiled later by the user of the library. This kind of "weaken" the purpose of separate compilation, but I don't see any other way of doing it.


Vertmo:
Oh yeah I would also require help to find out how to make `stack build` also trigger the compilation of the core library (that is trigger *core/Makefile*). The way I did it previously seemed to trip MacOS CI...

jacereda:
My https://github.com/carp-lang/Carp/pull/900 also fails on the MacOS CI.

eriksvedang:
@Vertmo I can try it out locally on MacOS, should I just run the example or what's a good test?

Vertmo:
Well, according to the CI its building Carp itself that fails. Does it work on your machine ?

eriksvedang:
Right now I'm getting a bunch of merge conflicts when trying it out (might be because I just merged some other PR). I'll be happy to try to compile it after fixing that!

eriksvedang:
Will close to keep things tidy, feel free to open again in the future!

PR #904, opened 2020-07-08
core: add String.from-bytes

hellerve:
This PR adds `String.from-bytes`, which is the inverse of `String.to-bytes`.

Test cases are included.

Cheers


hellerve:
Similarly to what happened in #903, we have a weird build failure on macOS.

hellerve:
FWIW, on macOS locally it builds fine.

hellerve:
I screwed up the history. Let’s try again.

PR #905, opened 2020-07-08
make bit-* interfaces

hellerve:
This PR makes bitwise operations interfaces.

Cheers


eriksvedang:
Great! I think you have to use Fn instead of the lambda literals for it to work on Windows..? (On my phone so I didn’t investigate this)

hellerve:
Since the other interfaces also use lambda and the Windows tests pass I‘d argue we‘re good?

eriksvedang:
@hellerve OK, yeah...

hellerve:
I think the problem is the rendering, not the source files? I can’t quite remember the issue, TBH.

PR #909, opened 2020-08-07
Add a `with-copy` macro

scolsen:
It's a fairly common pattern in Carp to call a higher-order function on
some structure of values, such as an Array. However, these structures,
and their members, all have lifetimes under Carp's memory management
model, which means they expect functions that are mapped over them to
take *a reference to a value* rather than a pure value. Array.reduce is
one example of such a "referential" higher-order, the type of its
function argument is:

```
(Fn [a, (Ref b c)] a)
```

That is, this function takes some pure initial value, then expects to be
called against the members of an array, which are *references* to the
values that are alive throughout the Array's lifetime.

However, one often wants to use a function that operates on pure values
in such contexts, such as +, which forces the programmer to write
anonymous functions that handle copying referenced values to pass them
to the underlying "pure" function:

```
(Array.reduce &(fn [x y] (+ x @y)) 0 &[1 2 3])
```

So, in using some high-order function over some structure in Carp one
usually has to do two things:

1. Wrap the function in a ref
2. Handle copying references into values in order to pass them into some
   simpler function that can also be used outside of memory-bound
   contexts.

The `proxy` macro captures this pattern. It wraps a given function in a
referenced anonymous function and copies an argument of that function at
a designated position before calling the underlying function. For
example, with `proxy`, the above example becomes:

```
(Array.reduce (with-copy + 2) 0 &[1 2 3])
```

The macro effectively gives a name to a common pattern--typically it
will only save the programmer a few characters, but it perhaps makes the
act of using a "function that doesn't care about references" in a
reference dominant context more apparent.

One can also use the macro to develop more specialized macros for
certain higher-orders, since these usually dictate where copying must be
performed. For instance, the `Array.reduce` function argument always
expects the referenced value to occur in the second position, thus one
could write:

```
(defmacro reducer [function] (eval (list with-copy function 2)))
```

Then the above code becomes even simpler:

```
(Array.reduce (reducer +) 0 &[1 2 3])
```

Which roughly means, "use the + function (which has no concept of
references) in this reference dependent context".

N.B. The examples using `+` won't work as of now due to current bugs
related to calling `arity` directly on an interface--but a synonym for
plus `add` defined as an explicit function will make all the above work
as expected.

### DIFF
--- a/core/Introspect.carp
+++ b/core/Introspect.carp
@@ -26,6 +26,14 @@
           false
           (Dynamic.= (Symbol.from "command") (car s)))))
 
+  (doc primitive?
+    "Is this binding a primitive?")
+  (defndynamic primitive? [binding]
+    (let [s (s-expr binding)]
+      (if (empty? s)
+          false
+          (Dynamic.= (Symbol.from "primitive") (car s)))))
+
   (doc variable?
     "Is this binding a variable?")
   (defndynamic variable? [binding]
@@ -62,6 +70,8 @@
     "What's the arity of this binding?
 
     - When `binding` is a function, returns the number of arguments.
+    - When `binding` is a command, returns the number of arguments.
+    - When `binding` is a primitive, returns the number of arguments.
     - When `binding` is an interface, returns the number of arguments.
     - When `binding` is a struct, returns the number of fields.
     - When `binding` is a sumtype, returns a list of the number of type
@@ -72,6 +82,7 @@
         0
     (cond
         (Introspect.command? binding) (length (caddr s))
+        (Introspect.primitive? binding) (length (caddr s))
         (Introspect.interface? binding) (length (car (cdaddr s)))
         (Introspect.function? binding) (length (caddr s))
         (Introspect.struct? binding) (/ (length (caddr s)) 2)
--- a/core/Introspect.carp
+++ b/core/Introspect.carp
@@ -127,42 +127,38 @@
           false
           (Dynamic.= (Symbol.from "definterface") (car s)))))
 
-  (doc proxy
-    "Returns an reference to an anonymous 'proxied' variant of `function` in
+  (doc with-copy
+    "Returns a reference to an anonymous 'proxied' variant of `function` in
   which the argument in position `arg` is *copied* from a reference before being
-  passed to `function`.
-
-  This is most useful when using higher orders that operate over structures that
-  return references to their inhabitants, such as arrays or structs. It allows
-  you to use a function over values without writing a custom anonymous function
-  to handle copying, for example:
+  passed to `function`:
 
   ```
-  (defn add [x y] (+ x y))
   ;; Array.reduce expects a function that takes a *ref* in its second argument:
   ;;   (Fn [a (Ref b c)] ...)
-  ;; so we can't use a function like `add` directly.
-  (reduce (proxy add-one 2) 0 &[1 2 3])
+  ;; so we can't use a function like `+` directly; enter proxy
+  (reduce (with-copy + 2) 0 &[1 2 3])
   => 6
   ;; compare this with an inline anonymous function that achieves the same thing:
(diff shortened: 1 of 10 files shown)


scolsen:
I tired of battling with recursive types for a bit, so I took some time to implement this since it's something I run into often.

BTW, happy to choose another name for this macro. I chose `proxy` since it effectively "proxies some value by copying it" before passing it to a function, and also "proxies" the function itself, by wrapping it in a lambda+ref. 

But if anyone has a better name it'd be welcome, as `proxy` has other, stronger connotations in computing.

scolsen:
Not sure why the MacOS CI failed--it succeeds locally.

core/Introspect.carp
```diff
@@ -94,4 +94,50 @@
...
+  return references to their inhabitants, such as arrays or structs. It allows
+  you to use a function over values without writing a custom anonymous function
+  to handle copying, for example:
+
+  ```
+  (defn add [x y] (+ x y))
+  ;; Array.reduce expects a function that takes a *ref* in its second argument:
+  ;;   (Fn [a (Ref b c)] ...)
+  ;; so we can't use a function like `add` directly.
+  (reduce (proxy add-one 2) 0 &[1 2 3])
```
hellerve, line 113: Did you mean to use `add` instead of `add-one` here?

scolsen:
Thanks @hellerve! I've added some changes to better support introspection so that things like `+` work with this.

I've also renamed the macro to `with-copy`.

core/Macros.carp
```diff
@@ -724,6 +724,9 @@ The expression must be evaluable at compile time.")
 (doc *gensym-counter* "is a helper counter for `gensym`.")
 (defdynamic *gensym-counter* 1000)
 
+(defndynamic gensym-local [x]
```
hellerve, line 727: This does not prevent accidental shadowing, correct?

eriksvedang:
This looks great! I'll do a proper review asap.

core/Macros.carp
scolsen: That's correct, no shadowing prevention here. If you wind up using this in a context in which some `gensym-generatedX` already exists (where `x` is whatever you pass as an argument) the resulting generated symbol will shadow the extant one--I guess we could get around this by adding another symbol to the end--something like `gensym-generated-localX`?

jacereda:
> Not sure why the MacOS CI failed--it succeeds locally.

Something unrelated most probably, both https://github.com/carp-lang/Carp/pull/900 and https://github.com/carp-lang/Carp/pull/903 fail.

eriksvedang:
Very cool macro! Seems really useful.

Does the "indexing" of the arg to copy start at 1, or am I misunderstanding?

scolsen:
> Very cool macro! Seems really useful.
> 
> Does the "indexing" of the arg to copy start at 1, or am I misunderstanding?

Yep, that's correct.

eriksvedang:
> > Very cool macro! Seems really useful.
> > Does the "indexing" of the arg to copy start at 1, or am I misunderstanding?
> 
> Yep, that's correct.

That seems a bit unintutive given that everything else is 0 indexed, no? (or am I missing something?) Apart from that I'm all for merge:ing this, I think it's a great addition!

scolsen:
> > > Very cool macro! Seems really useful.
> > > Does the "indexing" of the arg to copy start at 1, or am I misunderstanding?
> > 
> > 
> > Yep, that's correct.
> 
> That seems a bit unintutive given that everything else is 0 indexed, no? (or am I missing something?) Apart from that I'm all for merge:ing this, I think it's a great addition!

Good point! I see it both ways, on the one hand, mechanical consistency suggests we index from `0`, while on the other hand, people typically speak of "the first argument" of a function (rather than the "zeroth" argument).

Let's index from `0` like you're suggesting, Erik, we can always adjust if it proves confusing--I'll make the change later today!

scolsen:
@eriksvedang it should index from 0 after the latest commit 👍

PR #910, opened 2020-08-08
Allow Refs in polymorphic constructor positions

scolsen:
This change makes refs "constructible" in the perspective of the type
system by allowing them to stand for polymorphic constructors such as
`(f a b)`. After this commit one can take an interface such as:

```
(definterface out (Fn [(f a b)] a))
```

and provide an implementation over references:

```
(defn copied [x] @x)
(implements out copied)
```

`copied` is a valid implementation of `out`, since it simply copies the
value of a reference, dropping its lifetime.

Essentially, this makes the type system slightly more flexible, allowing
one to view references as type constructors in their own right.

## Motivation

For a little bit of background--this enables handling copying more flexibly in certain contexts. Some functions rightly need a reference and some rightly need a value, but in certain scenarios a given function should be able to handle both, regardless of the memorial context--in this case, we introduce surrogate reference types (which have no meaning other than "lives forever")

Here was my impetus for this:

```
(deftype (Temporal a b) (Eternal [a b]))
(definterface duplicate (Fn [(f a b)] a))
(defn dup-ref [x] @x) ;; copy can't implement the interface directly
(defn dup-eternal [x] (match x (Temporal.Eternal y _) y)
(implements duplicate dup-ref)
(implements duplicate dup-eternal)

(defn eternalize [x] (Temporal.Eternal x Opaque))

(duplicate &2)
=> 2
(duplicate (eternalize 2))
=> 2
```

Any function that calls into `duplicate` and adds the appropriate `eternalize` call may comfortably handle both references and pure values. Effectively, this defines an API that views *all values* as having lifetimes--which for pure values is an eternal lifetime.

I'll probably write a tiny library around this soon.

### DIFF
--- a/src/Constraints.hs
+++ b/src/Constraints.hs
@@ -149,6 +149,24 @@ solveOneInternal mappings constraint =
             Left err -> Left err
             Right ok -> solveOneInternal ok (Constraint ltA ltB i1 i2 ctx ord)
 
+    -- As a special case, allow Refs to stand for higher-order polymorphic
+    -- structs (f a b) ~ (Ref a b)
+    Constraint (StructTy v@(VarTy _) args) (RefTy b ltB) _ _ _ _ ->
+      let (Constraint _ _ i1 i2 ctx ord) = constraint
+      in  case solveOneInternal mappings (Constraint v (RefTy b ltB) i1 i2 ctx ord) of
+            Left err -> Left err
+            Right ok -> foldM (\m (aa, bb) -> solveOneInternal m (Constraint aa bb i1 i2 ctx ord)) ok (zip args [b, ltB])
+
+    -- TODO: The reverse argument order is necessary here since interface code
+    -- uses the opposite order of most other solving code (abstract, concrete
+    -- vs. concrete, abstract)--we should bring the interface code into
+    -- compliance with this to obviate this stanza
+    Constraint (RefTy b ltB) (StructTy v@(VarTy _) args) _ _ _ _ ->
+      let (Constraint _ _ i1 i2 ctx ord) = constraint
+      in  case solveOneInternal mappings (Constraint v (RefTy b ltB) i1 i2 ctx ord) of
+            Left err -> Left err
+            Right ok -> foldM (\m (aa, bb) -> solveOneInternal m (Constraint aa bb i1 i2 ctx ord)) ok (zip args [b, ltB])
+
     -- Else
     Constraint aTy bTy _ _ _ _ ->
       if aTy == bTy
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -204,6 +204,9 @@ areUnifiable (StructTy a aArgs) (StructTy b bArgs)
 areUnifiable (StructTy (VarTy _) aArgs) (FuncTy bArgs _ _)
   | length aArgs /= length bArgs = False
   | otherwise = all (== True) (zipWith areUnifiable aArgs bArgs)
+areUnifiable (StructTy (VarTy _) args) (RefTy _ _)
+  | length args == 2 = True
+  | otherwise = False
 areUnifiable (StructTy _ _) _ = False
 areUnifiable (PointerTy a) (PointerTy b) = areUnifiable a b
 areUnifiable (PointerTy _) _ = False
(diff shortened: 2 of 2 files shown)


hellerve:
This is extremely interesting! I’m a little concerned about `b` in `f a b`, though, since for a ref it will be a lifetime variable, and for everything else it will be a type variable, essentially “shapeshifting” depending on the definition. I’m not sure that this leads to actual problems, but it seems really weird to me!

scolsen:
That is definitely a weird aspect to this! At first I was hopeful that we'd get "lifetime functions" for free, e.g. something like: `(definterface both-alive (Fn [(f a b) (f c b)] Bool))` could be trivially defined as `true` since the type checker would reject any refs passed to this function that had different lifetime variables--enforcing that certain references have equivalent lifetimes...but this doesn't work--afict, this function will just work for every and any pair of refs even if one uses `the` to explicitly type them as having distinct lifetimes.

Probably the right thing to do, and what makes more sense is to view `(Ref x lt)` as a single argument type constructor `(f a)`, treating lifetimes as something separate from the type system (even though currently they're implemented as part of the type system). This should be possible, so I'll make the change!

eriksvedang:
My gut feeling is that this is somewhat unsound, though I'm not sure I can prove it right now.

The idea is super cool though, and seems like it would be really useful..! (if it actually is sound, which it might be)

I need to think about it a bit.

eriksvedang:
I haven't been able to figure out what to make of this PR still, and if it even makes sense... :P 

I think we should merge it though – any weirdness that comes out of it will be interesting, for sure.

Do you still feel it's a good idea @scolsen ? Did you want to make any changes to it (like you alluded to) or should I just go ahead and push the button?

scolsen:
> I haven't been able to figure out what to make of this PR still, and if it even makes sense... :P
> 
> I think we should merge it though – any weirdness that comes out of it will be interesting, for sure.
> 
> Do you still feel it's a good idea @scolsen ? Did you want to make any changes to it (like you alluded to) or should I just go ahead and push the button?

Well, I ultimately do think it would be a cool feature to have--but I think before we do it, we need to determine our philosophy on lifetimes and their relationship to the type system. 

If we just treat lifetimes as types themselves basically just type vars, just like any other type, I think the current implementation would be fine.

If we think lifetimes aren't really types, we should make changes to the type system that makes it clearer when a type is a lifetime and not a type, in which case we can still support this, but from the type system signature perspective lifetimes would be essentially ignored and in this case we could still support this pr but `Ref`'s signature would be `(Ref a)` aka `(f a)`

eriksvedang:
Right, the lifetimes rely on the type unification algorith but they always stay separated from the other types through dicipline/construction (although in theory you could have any type as a lifetime). It is strange, I'm not sure I like it. But the unification part is really important for it to work, and I couldn't come up with anything better. I guess we could have a separate lifetime datatype in the positions where lifetimes are needed, but then we need a mini-unifier for those that will have to run inside the normal type unifier. Arguable it would be a little bit more clean, but I'm not sure if it changes any semantics.

I think that coming to a conclusion of how to represent lifetimes in the compiler would be great, but perhaps we can just merge this and change it later as these things evolve...

scolsen:
> Right, the lifetimes rely on the type unification algorith but they always stay separated from the other types through dicipline/construction (although in theory you could have any type as a lifetime). It is strange, I'm not sure I like it. But the unification part is really important for it to work, and I couldn't come up with anything better. I guess we could have a separate lifetime datatype in the positions where lifetimes are needed, but then we need a mini-unifier for those that will have to run inside the normal type unifier. Arguable it would be a little bit more clean, but I'm not sure if it changes any semantics.
> 
> I think that coming to a conclusion of how to represent lifetimes in the compiler would be great, but perhaps we can just merge this and change it later as these things evolve...

ah yeah, this is tricky. in order to distinguish between lifetimes and types but still use the only unification we'd need something close dependent types. we could do something like the following:

```haskell
class Unifiable a where
  asTy : a -> Ty

data Lifetime = Static | (Var String)

Unifiable Lifetime where
  asTy Static = StaticLifetimeTy
  asTy (Var v) = VarTy v

Unifiable Ty where
  asTy = id 

unifySignatures :: Unifiable a => a -> a -> TypeMappings 
```

Then we'd call `asTy` to perform the cast for lifetimes in the unification algorithm. In the frontend and other places we can distinguish between when `Ty` is expected and when `Lifetime` is expected then in unification we can `asTy` to treat them the same. Not exactly an elegant solution, the more correct way to do it probably is to use some kind of special mini-unifier, but this would (I think) also work.

eriksvedang:
Sounds good to me!

PR #912, opened 2020-08-13
Fix Path docs.

jacereda:
I messed the docs when moving `directory` and `absolute` to the `Path` module.


eriksvedang:
When did that mess-up happen..? I haven't noticed anything...

jacereda:
Here: https://github.com/carp-lang/Carp/commit/134d9a5b02ce91c53ad61262d3758d85443239a4

eriksvedang:
OK, thanks!

PR #914, opened 2020-08-21
Ensure command last calls handle empty lists

### LINKED ISSUE #897: Prelude.last: empty list
When evaluating invalid macros, carp sometimes throws a Haskelly error:

`carp: Prelude.last: empty list`

I encountered this when evaluating the following macro:

```
(hidden gen-vec)
  (defndynamic gen-vec [vs acc]
    (if (= 0 (length vs))
        acc
        (Vector.gen-vec (cdr vs) (cons-last (list 'Vector.push (car vs)) (cons-last '(ref) acc)))))

  (doc of "
    Creates a new Vector from the contents of an Array literal.")
  (defmacro of [values]
        (Vector.gen-vec values (list '==> '(Vector.empty))))
```

Against an empty array (`[]`) argument. `(of [])`. This causes an error since the `==>` clause winds up receiving an empty list as its final argument.  A length check on `values` is sufficient for a fix; but we should likely throw an actual error--at the moment it looks like we don't account for this case and call `last` when we shouldn't.

scolsen:
The `last` command relies on Haskell's `Prelude.last`--which will crash
on an empty list input. We report an error for incorrect input to last,
but we didn't handle the empty list case. This small change to our
pattern match ensures errors in carp code produce Carp's last command
error rather than causing `Prelude.last` to throw an error on empty list
input.

fixes #897

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -554,8 +554,8 @@ commandCdr ctx [x] =
 commandLast :: CommandCallback
 commandLast ctx [x] =
   case x of
-    XObj (Lst lst) _ _ -> return (ctx, Right (last lst))
-    XObj (Arr arr) _ _ -> return (ctx, Right (last arr))
+    XObj (Lst lst@(x:xs)) _ _ -> return (ctx, Right (last lst))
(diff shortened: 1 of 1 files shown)


eriksvedang:
Excellent, thanks!

PR #921, opened 2020-08-26
Add doto and doto-ref

hellerve:
This PR adds `doto` and `doto-ref`, which are imperative versions of `->`.

Example:

```clojure
(doto-ref @\"hi\"
  (string-set! 0 \o)
  (string-set! 1 \y)) ; => "oy"
```

Test cases are included.

Cheers

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -774,3 +774,46 @@ the program will be aborted with an error message.")
           msg)))
     (list 'System.abort)
     (list 'bottom)))
+
+(doc doto "evaluates `thing`, then calls all of the functions on `x` and
+returns it. Useful for chaining mutating, imperative functions, and thus
+similar to `->`. If you need `thing` to be passed as a `ref` into `expressions`
+functions, use [`doto-ref`](#doto-ref) instead.
+
+Example:
+
+```
+(let [x @\"hi\"]
+  @(doto &x
+    (string-set! 0 \o)
+    (string-set! 1 \y)))
+```
+")
+(defmacro doto [thing :rest expressions]
+  (let [s (gensym)]
+    (list 'let [s thing]
+      (cons-last
+        s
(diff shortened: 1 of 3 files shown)


core/Macros.carp
```diff
@@ -774,3 +774,46 @@ the program will be aborted with an error message.")
           msg)))
     (list 'System.abort)
     (list 'bottom)))
+
+(doc doto "evaluates `thing`, then calls all of the functions on `x` and
```
scolsen, line 778: I think the `x` in this line should read `thing`?

core/Macros.carp
```diff
@@ -774,3 +774,46 @@ the program will be aborted with an error message.")
...
+```
+")
+(defmacro doto [thing :rest expressions]
+  (let [s (gensym)]
+    (list 'let [s thing]
+      (cons-last
+        s
+        (cons 'do (map (fn [expr] (cons (car expr) (cons s (cdr expr)))) expressions))))))
+
+(doc doto-ref "evaluates `thing`, then calls all of the functions on `x` and
```
scolsen, line 799: I think `x` here should be `thing`?

hellerve: Yup, sorry!

hellerve:
I’m not currently aware of any functions that are mutating and put the variable in the last position. I could add it anyway, but I’m not sure how useful it would be?

core/Macros.carp
hellerve: Fixed!

scolsen:
> I’m not currently aware of any functions that are mutating and put the variable in the last position. I could add it anyway, but I’m not sure how useful it would be?

Ahh good point, I guess I was thinking about completeness. Even though we don't have functions that take refs in the last position, library designers might...then again, it's probably not too difficult to get around it or implement the --> variant independently.

PR #926, opened 2020-09-21
Add docs on interfaces and `implements`

### LINKED ISSUE #925: "implements" missing from LanguageGuide.md
It looks like the LanguageGuide.md is missing `implements`

It includes `definterface`, but doesn't state how to implement said interface. I believe `implements` is the way. Example from `Double.carp`

```
  (implements + Double.+)
  (implements - Double.-)
  (implements * Double.*)
  (implements / Double./)
  (implements < Double.<)
  (implements > Double.>)
  (implements = Double.=)
  (implements copy Double.copy)
  (implements neg Double.neg)
  (implements to-int Double.to-int)
  (implements from-int Double.from-int)
```

scolsen:
Interfaces are now explicitly implemented using the `implements` key
word (in the past, implementation of an interface was implicit based on
matching names and function signatures).

This commit updates the language guide to account for this change and
includes an illustrative example to provide guidance on using
interfaces.

Fixes #925

### DIFF
--- a/docs/LanguageGuide.md
+++ b/docs/LanguageGuide.md
@@ -69,6 +69,77 @@ foo ; symbol
 (defmodule <name> <definition1> <definition2> ...) ;; The main way to organize your program into smaller parts
 ```
 
+#### Interfaces
+
+Interfaces specify a generic function signature that multiple concrete
+functions may implement. You can define an interface using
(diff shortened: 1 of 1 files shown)


eriksvedang:
Awesome, thanks a lot for this @scolsen !

PR #927, opened 2020-09-21
Use pre-included image stack in macOS CI

scolsen:
I'm not certain that this will fix the recent build issues with the MacOS CI, but it follows the set up of the Linux CI, which has been more consistent.

I can't completely decipher the logs, but it seems like there may be some issue stemming from the fact that github's macos images [include stack](https://github.com/actions/virtual-environments/blob/macos-10.15/20200916.1/images/macos/macos-10.15-Readme.md#tools) and that we were downloading another version of stack using [setup-haskell](https://github.com/actions/setup-haskell). 

My theory is that somehow this is leading to conflicts.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -15,9 +15,6 @@ jobs:
         uses: actions/checkout@v2
 
       - uses: actions/setup-haskell@v1.1
-        with:
-          stack-version: 'latest'
-          enable-stack: true
 
       - run:   echo ::set-env name=GITHUB_SHA::$GITHUB_SHA
         shell: bash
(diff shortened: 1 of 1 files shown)


eriksvedang:
Thanks! I’ll try this

scolsen:
Still seems to be failing, but because of an HTTP timeout now...not sure what's going on really.

scolsen:
So the connection is still timing out, but the last commit got us further than before--it actually started building and looks like it timed out on fetching a dependency.

scolsen:
@eriksvedang looks like this might work! For some reason setting an explicit stack and GHC version seems to have done the trick...though it's still susceptible to occasional HTTP timeouts I guess.

jacereda:
Nice! Time to merge, I guess, since it's for sure in better shape than the current job.

scolsen:
> Nice! Time to merge, I guess, since it's for sure in better shape than the current job.

Yep, I'll close this PR and start a separate one to both confirm the change and give us a cleaner history.

PR #928, opened 2020-09-22
Use explicit stack and GHC versions in macOS CI

scolsen:
After some investigation, I've found that setting these versions explicitly seems to make the CI work again. I'm unfortunately not certain as to why--I think it has something to do with the cached dependencies and compatibility w/ stack and ghc versions.

At any rate, this *should* enable the CI to build successfully again.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -16,7 +16,8 @@ jobs:
 
       - uses: actions/setup-haskell@v1.1
         with:
-          stack-version: 'latest'
+          stack-version: '2.3.1'
+          ghc-version: '8.8.2'
           enable-stack: true
 
       - run:   echo ::set-env name=GITHUB_SHA::$GITHUB_SHA


scolsen:
So, everything failed this run 😓 , but it was due to an http timeout--in each case, not sure if there's anyway we can combat those...

scolsen:
Closed and reopened to trigger CI again.

eriksvedang:
OK, so it seems random? I retriggered it again to see how it goes.

eriksvedang:
Got this one again...
```
HttpExceptionRequest Request {
  host                 = "casa.fpcomplete.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
  path                 = "/v1/pull"
  queryString          = ""
  method               = "POST"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
```

scolsen:
> Got this one again...
> 
> ```
> HttpExceptionRequest Request {
>   host                 = "casa.fpcomplete.com"
>   port                 = 443
>   secure               = True
>   requestHeaders       = []
>   path                 = "/v1/pull"
>   queryString          = ""
>   method               = "POST"
>   proxy                = Nothing
>   rawBody              = False
>   redirectCount        = 10
>   responseTimeout      = ResponseTimeoutDefault
>   requestVersion       = HTTP/1.1
> }
> ```

Yep, the HTTP errors are in fact random flakiness I believe. The good news is the macOS CI, which is the one this PR updates and that we've had problems with actually passed last run! Looks like the Linux one fell to the HTTP timeout error.

I've poked around a bit but haven't really found too much info on why our CI could be timing out so frequently...

scolsen:
Reopening again to see if we can get at least one run in which all builds pass 🤞

scolsen:
Yikes, looks like the old error returned 😓

eriksvedang:
😭

PR #929, opened 2020-09-24
Test Mac Github actions build

MuhammedZakir:
All 3 runs of mine passed successfully. Hopefully, this passes too!!!

See https://github.com/MuhammedZakir/Carp/actions?query=workflow%3A%22MacOS+CI%22.

P.S. It is showing only 2 runs because I did "re-run jobs" for the last one.

P.S.2. Don't ask me why this is working. I don't know! 😄 Caching seems to work a bit differently for this. Maybe because of that?

Edit: I created a new PR instead of a new commit in #928, is to test whether it works here.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -13,6 +13,13 @@ jobs:
     steps:
       - name: Check out
         uses: actions/checkout@v2
+      
+      - uses: actions/cache@v1
+        name: Cache ~/.stack
+        with:
+          path: ~/.stack
+          key: ${{ runner.os }}-stack-${{ env.GITHUB_SHA }}
+          restore-keys: ${{ runner.os }}-stack-
 
       - uses: actions/setup-haskell@v1.1
         with:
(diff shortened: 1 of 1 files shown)


scolsen (approved):

scolsen:
Awesome, thanks @MuhammedZakir! It worked!

MuhammedZakir:
Why is there such a huge difference in Linux and Mac for Carp tests?

* Ubuntu - ~3.5+ mins
* Windows - ~5 mins
* Mac - ~6.2+ mins

scolsen:
Do you want to reopen this? It looks like its good! @eriksvedang can merge if it seems ok

eriksvedang:
I'm a bit confused, why was it closed? :)
I'm happy to merge if it works!

scolsen:
> > Do you want to reopen this? It looks like its good! @eriksvedang can merge if it seems ok
> 
> I am waiting for a reply to this
> 
> > I think we should move cache before Haskell setup. What do you think?
> > Edit: I have added a new commit [1]. Shall I also add that and enable Mac release workflow?
> > [1] [MuhammedZakir@228aa7c](https://github.com/MuhammedZakir/Carp/commit/228aa7cd7d634cc0a22ccb2817d0aa3b9db3b107)
> 
> If these changes are not needed and the changes in this PR is enough, then I will reopen it. Else, I will first add those changes to this PR, and then reopen it. Well, what should I do?

Ah I see, thanks! I'd say let's just commit this PR as is--if we find out that other commit is important too, we can merge it in a separate PR later.

scolsen:
Looks good to me, @eriksvedang what do you think?

PR #931, opened 2020-09-25
Add Introspect.execution-mode to expose the compiler's execution mode to users

scolsen:
This commit adds two variables, `Introspect.execution-mode`, and `Introspect.execution-mode-dynamic` each of which represent the compilers current execution mode in a Carp program. `execution-mode` is bound to a value of `ExecutionMode`: `Repl`, `Build`, `Run`, `Check`, or `Install`. The `execution-mode-dynamic` variable represents the same modes as symbols: `'Run` `'Build` `'Repl` etc.

Here's an example of this behavior:

```
;; Start Carp in REPL mode.
carp
(if (= Introspect.execution-mode-dynamic 'Repl) 1 2)
=> 1
(match Introspect.execution-mode (ExecutionMode.Repl) @"yay" _ "boo")
=> "yay"
```

```
;; source for foo.carp
(defn main []
  (match Introspect.execution-mode (ExecutionMode.Run) 1 (ExecutionMode.Build) 2 _ 3))
;; Run foo
carp -x foo.carp
=> exited with return value 1
;; Build foo
;; Run foo
carp -b foo.carp
=> exited with return value 2
```


scolsen:
I had a fun idea for a library last night that would rely on this info. Not sure how useful it'll be to most users, but I couldn't think of a way to implement this in Carp itself, so I had to add it to the starting env. If there are alternatives implementations that don't touch the compiler though, I'd be happy to use them!

jacereda:
Would exposing `contextExecMode` be more useful?

scolsen:
> Would exposing `contextExecMode` be more useful?

Interesting, it'd certainly be more general. That sounds like a fine idea to me. 

Do you think it'd make more sense to expose it as a string or to limit access to the dynamic env and expose it as a symbol?

jacereda:
I don't know, I guess this is rare enough that a not-so-perfect implementation could be enough until the need for something better arises. Maybe at some point we'll need more stuff from `Context` and it will be time to implement a `dynamicContextModule`.

hellerve:
It might also make sense to think about making this a dynamic symbol instead, since macros might want to depend on this as well.

scolsen:
> It might also make sense to think about making this a dynamic symbol instead, since macros might want to depend on this as well.

Totally! I think? the dynamic mode should be sufficient by itself....but then I couldn't get something to compile using it even though it worked fine in the REPL.

So, for now, I've just exposed two vars, one at the static and one at the dynamic level.

The static var is bound to an enum type that represents the execution modes.

scolsen:
Updated the description

scolsen:
🤔 not sure what's causing the test failure just yet...

hellerve:
Funny, I found the same thing when working on my article about macros. I think dynamic symbols do not get evaluated in static contexts anymore, which is a little annoying.

hellerve:
To be clear: this is totally on me, and relates to the new evaluator. I think I’m missing a case somewhere (sigh).

scolsen:
> Funny, I found the same thing when working on my article about macros. I think dynamic symbols do not get evaluated in static contexts anymore, which is a little annoying.

Ahhh I see, it's great that you found the issue!

scolsen:
> 🤔 not sure what's causing the test failure just yet...

The problem stemmed from `--no_core` builds. Since I was adding a new type (relies on String for `str` and `prn`) and calling `doc` core has to be enabled...so, now we only add the execution mode stuff when core is enabled, which seems reasonable to me.

app/Main.hs
```diff
@@ -116,9 +116,34 @@ main = do setLocaleEncoding utf8
               preloads = optPreload fullOpts
               postloads = optPostload fullOpts
               load = flip loadFiles
+              executionModeDoc = "(defmodule Introspect (doc execution-mode \"Returns the mode the Carp compiler is currently executing in.\"))"
```
eriksvedang, line 119: I'm not sure I understand why this is defined so early (and in such an unorthodox way...)

Please enlighten me :)

scolsen: I can make it inline in the call to `execStr` -- admittedly I had originally combined the doc declarations and var declarations, so I bound it to a var to help readability since the string was quite long!

scolsen: I inlined it now e0f34f9 hopefully that makes more sense!

app/Main.hs
```diff
@@ -142,8 +140,8 @@ main = do setLocaleEncoding utf8
             >>= whenCore (\ctx -> (execStr "Internal" "(deftype ExecutionMode Repl Build Install Run Check)") ctx
                          >>= (execStr "" "(doc ExecutionMode \"Represents the mode the Carp compiler is executing in.\")")
                          >>= setExecutionMode execMode
-                         >>= (execStr "" executionModeDoc)
-                         >>= (execStr "" executionModeDynamicDoc))
```
hellerve: Can’t the dynamic part be offloaded into `StartingEnv.hs`?

eriksvedang:
I re-ran the action and now it failed on macOS, so it seems like there's no difference with these changes unfortunately.

app/Main.hs
scolsen: I think theoretically it could, but I'm a bit reluctant since then we'd have to add logic to pull out the execmode from the starting env's context, and I kinda feel the current impl is slightly simpler

scolsen:
@eriksvedang the CIs pass now if you wanna take another look!

app/Main.hs
```diff
@@ -116,9 +116,33 @@ main = do setLocaleEncoding utf8
               preloads = optPreload fullOpts
               postloads = optPostload fullOpts
               load = flip loadFiles
+              setExecutionMode Repl ctx =
+                execStr "Internal" ("(defmodule Introspect (defdynamic execution-mode-dynamic 'Repl))") ctx
```
eriksvedang, line 120: I'm afraid I still don't get why these are defined by evaluating strings here, rather than in a file or as built in commands?

scolsen: Oh, I see what you mean now--sorry, I misunderstood your original comment!

We can define these as commands, I think it's more work to plumb the execution mode into the command code but it's certainly possible. There's no good reason for executing these as strings other than it was simpler than plumbing the Execution mode through to a command--but I think its carried through the context? So that should work.

@hellerve suggested this as well, I'll take some more time on this and impl these as commands.

eriksvedang:
This PR still confuses me greatly :)
Do we need it, or has its purpose been superseeded?

scolsen:
I don't think we actually need it anymore--my thinking behind it was to power/execute some kind of tutorial only when the user is in REPL mode--but @hellerve's tutorial lib has obviated this.

I'll close this one!

PR #935, opened 2020-10-01
Make `doc` take a rest parameter of strings

scolsen:
Carp preserves tabulation and other whitespace in strings, as it should.
This sometimes results in awkward code indentation when it comes to long
doc strings that exceed 80 characters. Often, one has to continue the
string on a newline, but needs to avoid adding tabs to prevent Carp from
rendering them in the output.

This change alters the behavior of doc to take a series of strings as a
rest parameter instead, allowing for neater organization in the source,
for example, after this change the following long doc string for
compose:

~~~
    "Returns the composition of two functions `f` and `g` for functions of any
arity; concretely, returns a function accepting the correct number of
arguments for `g`, applies `g` to those arguments, then applies `f` to the
result.

If you only need to compose functions that take a single argument (unary arity)
see `comp`. Comp also generates the form that corresponds to the composition,
compose contrarily evaluates 'eagerly' and returns a computed symbol.

For exmaple:

```
;; a silly composition
((compose empty take) 3 [1 2 3 4 5])
;; => []

(String.join (collect-into ((compose reverse map) Symbol.str '(p r a c)) array))
;; => 'carp'

;; comp for comparison
((comp (curry + 1) (curry + 2)) 4)
;; => (+ 1 (+ 2 4))
```"
~~~

becomes:

~~~
       "Returns the composition of two functions `f` and `g` for functions of any"
       "arity; concretely, returns a function accepting the correct number of"
       "arguments for `g`, applies `g` to those arguments, then applies `f` to the"
       "result."
       ""
       "If you only need to compose functions that take a single argument (unary arity)"
       "see `comp`. Comp also generates the form that corresponds to the composition,"
       "compose contrarily evaluates 'eagerly' and returns a computed symbol."
       "```"
       ";; a silly composition"
       "((compose empty take) 3 [1 2 3 4 5])"
       ";; => []"
       ""
       "(String.join (collect-into ((compose reverse map) Symbol.str '(p r a c)) array))"
       ";; => 'carp'"
       ""
       ";; comp for comparison"
       "((comp (curry + 1) (curry + 2)) 4)"
       ";; => (+ 1 (+ 2 4))"
       "```")
~~~

And the output remains the same; this just enables better alignment in
the source file.

The strings passed to doc are newline separated by default, but one can
circumvent this behavior by passing a bool along with the string as
follows:

~~~
(doc foo
     ("My amazing doc " false)
     "continues on one line."
     ""
     "And then another.")
~~~

The above doc string will result in the following output:

~~~
My amazing doc continues on one line.

And then another.
~~~

Of course, the original behavior of doc also remains valid, so if one
prefers to use the old indentation-mixed single string format, one still
can!

This change also reformats the doc strings in macro to utilize the new
rest parameter and make the source a bit neater.

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -6,9 +6,29 @@
 (defmacro or [x y]
   (list 'if x true y))
 
+;; Defined early so that `doc` can accept a rest arg
+(defndynamic map-internal [f xs acc]
+    (if (= 0 (length xs))
+        acc
+        (map-internal f (cdr xs) (cons-last (f (car xs)) acc))))
+
+(defndynamic list-to-array-internal [xs acc]
+    (if (= 0 (length xs))
+        acc
+        (list-to-array-internal (cdr xs) (append acc (array (car xs))))))
+
 (meta-set! doc "doc" "Set documentation for a binding.")
-(defmacro doc [name string]
-  (eval (list 'meta-set! name "doc" string)))
+(defmacro doc [name :rest strings]
+  (let [newline "
+" ;; Looks a bit odd but the newline literal is important here! (str \newline) currently results in unwanted escapes
+        separated (map-internal (fn [x] (if (list? x)
+                                            (if (cadr x)
+                                                (Dynamic.String.concat [(car x) newline])
+                                                (car x))
+                                            (Dynamic.String.concat [x newline])))
+                                strings
+                                ())]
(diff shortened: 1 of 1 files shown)


hellerve (approved):
This is a great idea, and I especially like that it manages to be backwards compatible and that it’s customizable! Great work!

eriksvedang:
Very nice!

PR #937, opened 2020-10-01
Add DefDynamic to getPath

### LINKED ISSUE #936: `doc` currently doesn't work on `defdynamic` bindings
To reproduce:

```
(defmodule Foo (doc x "a number") (defdynamic x 2))
(print-doc Foo.x)
=> ()
```

scolsen:
getPath returns the path of a form. The DefDynamic case was missing from
the function's definition, resulting in failures when calling functions
such as the meta set primitive on dynamic defs. This change fixes such
issues.

Fixes #936


scolsen:
After this is in, we can also remove the commented out defdynamic doc added in PR #931

PR #938, opened 2020-10-02
Expand comparables in dynamic =

hellerve:
This PR adds a few comparable classes to dynamic `=`: arrays, non-empty lists, and all AST types without arguments (because they’re trivially comparable).

Cheers

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -620,30 +620,49 @@ commandMacroLog ctx msgs = do
 
 commandEq :: CommandCallback
 commandEq ctx [a, b] =
-  return $ case (a, b) of
-    (XObj (Num IntTy aNum) _ _, XObj (Num IntTy bNum) _ _) ->
-      if (round aNum :: Int) == (round bNum :: Int)
-      then (ctx, Right trueXObj) else (ctx, Right falseXObj)
-    (XObj (Num LongTy aNum) _ _, XObj (Num LongTy bNum) _ _) ->
-      if (round aNum :: Int) == (round bNum :: Int)
-      then (ctx, Right trueXObj) else (ctx, Right falseXObj)
-    (XObj (Num FloatTy aNum) _ _, XObj (Num floatTy bNum) _ _) ->
-      if aNum == bNum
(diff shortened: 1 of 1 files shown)


src/Commands.hs
```diff
@@ -620,30 +620,49 @@ commandMacroLog ctx msgs = do
...
+    cmp (XObj SetBang _ _, XObj SetBang _ _) = Right $ True
+    cmp (XObj Macro _ _, XObj Macro _ _) = Right $ True
+    cmp (XObj Dynamic _ _, XObj Dynamic _ _) = Right $ True
+    cmp (XObj DefDynamic _ _, XObj DefDynamic _ _) = Right $ True
+    cmp (XObj The _ _, XObj The _ _) = Right $ True
+    cmp (XObj Ref _ _, XObj Ref _ _) = Right $ True
+    cmp (XObj Deref _ _, XObj Deref _ _) = Right $ True
+    cmp (XObj (Lst []) _ _, XObj (Lst []) _ _) = Right True
+    cmp (XObj (Lst elemsA) _ _, XObj (Lst elemsB) _ _) =
+      foldr cmp' (Right True) (zip elemsA elemsB)
```
jacereda, line 658: Wouldn't this implementation yield True for two different length lists with the same prefix?

src/Commands.hs
```diff
@@ -620,30 +620,49 @@ commandMacroLog ctx msgs = do
...
+    cmp (XObj (Arr []) _ _, XObj (Arr []) _ _) = Right True
+    cmp (XObj (Arr elemsA) _ _, XObj (Arr elemsB) _ _) =
+      foldr cmp' (Right True) (zip elemsA elemsB)
```
jacereda, line 661: Same for arrays?

src/Commands.hs
hellerve: Whoops, good catch! I added a length check.

scolsen:
Comparing ASTs seems like it could be super useful!

eriksvedang:
Neato!

PR #939, opened 2020-10-02
Enhance type reflection

scolsen:
Extends Carp's support for type reflection by returning types for
values as well as bindings.

`type` now also returns a valid Carp expression/s-expression and so its
output can be used as input to dynamic functions and macros (prior to
this commit, `type` printed the type to the REPL but did not return a
meaningful expression in Carp).

Here are a few illustrations of the behavior:

```clojure
(def x 1)
;; type now returns an s-expression/symbol
(type x)
=> Int
;; It also works on values
(type 1)
=> Int
(type 2b)
=> Byte
(type "foo")
=> (Ref String <StaticLifetime>)
;; It works on more complex values as well
(type Maybe)
=> Module
(type Maybe.Just)
(Fn [a] (Maybe a) <StaticLifetime>)
;; reports honestly about polymorphism
(type (Maybe.Nothing))
=> (Maybe a)
(type (Pair.init 1 2))
=> (Pair Int Int)
;; What about the type of types?
(type (type 2))
=> Type
;; Or the type of types of types?
(type (type (type 2)))
=> ()
;; One more time!
(type (type (type (type 2))))
=> ()
;; so, () is the fixpoint of type, and is reached after two applications
(type zero)
;; the type of an interface is all of its implementations
=> (((Fn [] (Array a) <StaticLifetime>) (Fn [] Bool <StaticLifetime>) (Fn
[] Byte <StaticLifetime>) (Fn [] Char <StaticLifetime>) (Fn [] Double
<StaticLifetime>) (Fn [] Float <StaticLifetime>) (Fn [] Int
<StaticLifetime>) (Fn [] Int16 <StaticLifetime>) (Fn [] Int32
<StaticLifetime>) (Fn [] Int64 <StaticLifetime>) (Fn [] Int8
<StaticLifetime>) (Fn [] Long <StaticLifetime>) (Fn [] (Maybe a)
<StaticLifetime>) (Fn [] (Pair a b) <StaticLifetime>) (Fn [] (Quadruple
a b c d) <StaticLifetime>) (Fn [] String <StaticLifetime>) (Fn []
(Triple a b c) <StaticLifetime>) (Fn [] Uint16 <StaticLifetime>) (Fn []
Uint32 <StaticLifetime>) (Fn [] Uint64 <StaticLifetime>) (Fn [] Uint8
<StaticLifetime>)))
```

As shown in the example above, this change also includes a cosmetic
update to the representation of lifetime variables, which are surrounded
in <> to distinguish them from type variables.

This commit also adds a new `kind` primitive that reports on the kind of
a binding or value:

```clojure
(def x 3)
(kind x)
=> Base
(kind 2)
=> Base
(kind Maybe.Just)
=> Higher
(kind (Maybe.Just 2))
=> Higher
```

`kind` and `type` both support interactive development in the repl, for
example, a user can rely on `kind` to check the kind of a type they plan
on using in an interface that demands a higher-kinded argument.

Likewise, they both also support developing macros based on type
information.

### DIFF
--- a/CarpHask.cabal
+++ b/CarpHask.cabal
@@ -54,7 +54,8 @@ library
                        Path,
                        Interfaces,
                        Primitives,
-                       Validate
+                       Validate,
+                       Reify
 
   build-depends:       base >= 4.7 && < 5
                      , parsec == 3.1.*
--- a/docs/core/Dynamic.html
+++ b/docs/core/Dynamic.html
@@ -458,7 +458,14 @@
                     (all? f xs)
                 </pre>
                 <p class="doc">
-                    
+                    <p>Checks whether all of the elements in <code>xs</code> conform to the predicate
+function <code>f</code>.</p>
+<pre><code>(all? (fn [x] (&lt; 1 x)) '(2 3 4))
+=&gt; true
+(all? (fn [x] (&lt; 1 x)) '(-1 0 1))
+=&gt; false
+</code></pre>
+
                 </p>
             </div>
             <div class="binder">
@@ -477,11 +484,12 @@
                     (any? f xs)
                 </pre>
                 <p class="doc">
-                    <p>checks whether all of the elements in <code>xs</code> conform to the predicate
+                    <p>Checks whether any of the elements in <code>xs</code> conforms to the predicate
 function <code>f</code>.</p>
-<p>Example:</p>
-<pre><code>(all? (fn [x] (&lt; 1 x)) '(2 3 4)) ; =&gt; true
-(all? (fn [x] (&lt; 1 x)) '(-1 0 1)) ; =&gt; false
+<pre><code>(any? (fn [x] (= 'a x)) '(a b c))
+=&gt; true
+(any? (fn [x] (= 'a x)) '(e f g))
+=&gt; false
 </code></pre>
 
                 </p>
@@ -1331,7 +1339,6 @@ result.</p>
 <p>If you only need to compose functions that take a single argument (unary arity)
 see <code>comp</code>. Comp also generates the form that corresponds to the composition,
 compose contrarily evaluates 'eagerly' and returns a computed symbol.</p>
-<p>For exmaple:</p>
 <pre><code>;; a silly composition
 ((compose empty take) 3 [1 2 3 4 5])
 ;; =&gt; []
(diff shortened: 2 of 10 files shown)


scolsen:
I should mention that this change deprecates `tyToXObj` in favor of a generic typeclass `Reifiable` that we can use to produce an XObj for any datatype internal to the compiler (e.g. I will use this. for execution modes later)

hellerve (approved):
This is pretty awesome!

I suppose we don’t need to add clauses for `Universe` anywhere else, since it isn’t expected to go through the compiler except in the newly added functions?

scolsen:
> This is pretty awesome!
> 
> I suppose we don’t need to add clauses for `Universe` anywhere else, since it isn’t expected to go through the compiler except in the newly added functions?
 
That's correct! It's basically just a trick to distinguish between the symbol `Type` returned by `(type (type x))` from a potentially user defined type called `Type`. Discriminating on the symbol itself would make `(type Type)` (where type is defined by the user) return the wrong thing, so we tag the `Type` returned by `(type (type))` w/ Universe to discriminate between them, probably easier to illustrate as follows:

```
(deftype Type)
(type Type)
=> Module
(type (type 1))
=> Type
(type (type (type 1)))
=> ()
```

If we didn't have `Universe` the `Type` returned by `(type (type 1))` would conflict with the deftype on line 1--discriminating using Universe enables us to allow users to still use `Type` (e.g. someone might do this if they're defining an embedded language in carp) as a module/type name without returning the wrong result on chained calls to `type`.

And actually, I'm glad you brought this up as the pattern order was slightly incorrect to support ths! (since lookups occurred before type based discriminations) but its fixed now!

scolsen:
Of course, we could also disallow `Type` as a module name if it's confusing, or use a more special name for the type of a type `(type (type 1))` like `CarpType` or something. Though I find the current impl pretty nifty, I could see it confusing people in the rare situation that they actually do something like `(deftype Type)` but this would be pretty rare, and presumably calling such a deftype makes the distinction clear.

scolsen:
Unfortunately, it looks like the macos CI is flaky again :/

scolsen:
Also, the current implementation requires a little bit of indirection for complex uses, for instance, to use this w/ `Dynamic.map`:

```clojure
(Dynamic.map (fn [x] (eval (list type x))) '(Int.+ Double.+))
=> ((Fn [Int Int] Int <StaticLifetime>) (Fn [Double Double] Double <StaticLifetime>))
```

eriksvedang (approved):
Looks great!

PR #940, opened 2020-10-07
Add flag to determine binding lookup preferences in the evaluator

### LINKED ISSUE #849: `s-expr` dynamics can currently conflict with interfaces
Right now, since `s-expr` relies on our evaluation strategies, dynamic bindings are always preferred. Consider the case of `inc` which has an interface and dynamic definition, calling `s-expr` will only ever return:

```
鲤 (s-expr inc)
=> (dynamic Dynamic.inc [x] (+ x 1))
```

But users may want the interface definition. 

We need to add a facility for disambiguating such cases.

scolsen:
This change alters the evaluator to do two things:

* Organize lookup order using Control typclasses (alternative, monad) which makes our lookup strategies more modular, making it easy to change the order of lookups should we need to (e.g. global first then dynamic, internal first then global, etc.)
* Adds a flag to eval to let callers specify whether they prefer looking for unqualified bindings `inc` opposed to `Dynamic.inc`, in the Dynamic or Global environment first.

The behavior these changes support is that `s-expr` calls can now retrieve global or type level binding forms when these conflict with bindings in `Dynamic`--previously, Dynamic bindings were always preferred, giving `s-expr` calls no way to retrieve bindings that were actually associated to the global module.

```clojure
;; Before this PR
(s-expr inc)
=> (dynamic Dynamic.inc [x] (+ x 1)) ;; no way to access the interface :(
;; After this PR
(s-expr inc)
=> (definterface inc (Fn [a] a <StaticLifetime>))
(s-expr Dynamic.inc)
=> (dynamic Dynamic.inc [x] (+ x 1)) ;; interface and Dynamic.inc both accessible :)
```

fixes #849

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -39,6 +39,12 @@ import qualified Meta
 
 import Debug.Trace
 
+-- Prefer dynamic bindings
+evalDynamic ctx xobj = eval ctx xobj True
+
+-- Prefer global bindings
+evalStatic ctx xobj = eval ctx xobj False
+
 -- | Dynamic (REPL) evaluation of XObj:s (s-expressions)
 -- Note: You might find a bunch of code of the following form both here and in
 -- macroExpand:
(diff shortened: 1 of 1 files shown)


hellerve (approved):
The lookup logic always bothered me! Nice refactor!

src/Eval.hs
```diff
@@ -52,38 +59,34 @@ import Debug.Trace
 -- it gets real weird with laziness. (Note to the note: this code is mostly a
 -- remnant of us using StateT, and might not be necessary anymore since we
 -- switched to more explicit state-passing.)
-eval :: Context -> XObj -> IO (Context, Either EvalError XObj)
-eval ctx xobj@(XObj o i t) =
+eval :: Context -> XObj -> Bool -> IO (Context, Either EvalError XObj)
```
eriksvedang, line 62: I'd prefer an custom sumtype over a Bool here to avoid "boolean blindness".

src/Eval.hs
```diff
@@ -52,38 +59,34 @@ import Debug.Trace
...
-          else tryLookup path
-      where tryInternalLookup path =
-              case contextInternalEnv ctx of
-                Nothing -> Nothing
-                Just e  ->
-                  case lookupInEnv path e of
-                    Just (_ , Binder _ found) -> Just (ctx, Right (resolveDef found))
-                    Nothing -> Nothing
+      return
+      $ fromMaybe (evalError ctx ("Can't find symbol '" ++ show n ++ "'") i) -- all else failed, error.
```
eriksvedang, line 68: Nice rewrite of this code 👍

eriksvedang (approved):
Great!

src/Eval.hs
scolsen: Good point! Added a simple sumtype to capture this instead.

src/Eval.hs
scolsen: Thanks!

eriksvedang:
Thanks for the fixes @scolsen, looks awesome! Will merge once #900 is in, I don't want to accidentally block that one again :)

scolsen:
> Thanks for the fixes @scolsen, looks awesome! Will merge once #900 is in, I don't want to accidentally block that one again :)

Sounds good, thanks!

PR #942, opened 2020-10-14
Call Random.reseed at program start

hellerve:
This PR ensures that we call `Random.reseed` at program start to ensure randomness.

Cheers

### DIFF
--- a/core/Random.carp
+++ b/core/Random.carp
@@ -26,6 +26,11 @@
     (do
       (set! s (mod (+ (* s a) c) m))
       (/ s m)))
+
+  (private _)
+  (hidden _)
+  (doc _ "forces reseeding based on the current time at program start")
+  (def _ (do (seed) 0))
 )
 
 (defmodule Int
--- a/test/random.carp
+++ b/test/random.carp
@@ -4,11 +4,6 @@
 
 (deftest test
   (assert-op test
-             0.658908
-             (Random.random)
-             "deterministic randomization works as expected"
-             Double.approx)
-  (assert-op test
              0.536041
              (do (Random.seed-from 33333.0) (Random.random))
              "deterministic randomization with seed works as expected"


jacereda:
IMHO the seed should be set explicitly by the user.

hellerve:
I think it makes sense for the system to start with a somewhat random seed. The user can still set their own seed by using `Random.seed-from` should they wish to.

jacereda:
But that's unexpected because all languages I know default to determinism.

hellerve:
Really? I’ve seen a pretty even distribution.

At least JavaScript, Haskell, Clojure, and Python do not default to determinism. I know C and Common Lisp do. What languages are you referring to?

jacereda:
I don't think Haskell falls into the first group:

https://www.stackage.org/haddock/lts-16.18/random-1.1/src/System.Random.html#theStdGen

Besides that, Go, Forth, Zig, Nim, Lua, even Basic, also start with a known seed.

Clojure will probably do whatever Java does underneath.

hellerve:
I don’t know how helpful it is to make lists; the fact is that both approaches exist in popular languages.

I would like to expand on your point as for Haskell, though, because I believe you’re misunderstanding the code you linked to (you can test it by using `ghci -package random` and then using `randomIO`): `theStdGen` is initialized using `mkStdGen`, which in turn uses `getCPUTime` and `getTime`([source](https://www.stackage.org/haddock/lts-16.18/random-1.1/src/System.Random.html#mkStdRNG)).

jacereda:
Oh, I thought `theStdGen` was receiving the seed directly.

I'd say the division is somewhat between system languages and higher level languages. You can expect magic in things like Python, but I wouldn't expect it from a systems language.

We can also consider another viewpoint. This would mean an additional syscall. Yes, some people actually care about the number of syscalls at startup:

https://drewdevault.com/2020/01/04/Slow.html

hellerve:
Yeah, in the end it boils down to a philosophical question: what level of convenience/abstraction does Carp ship with by default? I’m happy to concede either way here, but I’d like to have some input from @carp-lang/main-contributors.

By the way, I just ran the numbers because I was curious: dtruss reports 29 syscalls, 14 of which are unique, which isn’t that bad, actually. What I do not understand is why the count does not seem to change whether I compile using this branch or on master, though the reseeding definitely happens. I’m not quite sure what I’m doing wrong here.

jacereda:
Which program are you tracing? `test/random.carp`?

hellerve:
I was writing a simple "Hello, World!" program to get comparable numbers to the blog post you quoted. It’s of course perfectly possible that the call got optimized away because the variable was never read again!

scolsen:
Maybe it's sort of silly, but since we have a pre and post-load mechanism and macros, could we leave this down to config? What I mean is, could we tuck this call behind a macro that can be pre-loaded if desired? That would give the user choice, but of course it's not as elegant and we'd have to document it somewhere...etc.

I don't have a strong preference either way. There seem to be benefits to both approaches.

hellerve:
At this point, the user could just call `(Random.seed)` themselves, no need for config! The question is just what the default should be. :)

scolsen:
> At this point, the user could just call `(Random.seed)` themselves, no need for config! The question is just what the default should be. :)

Ah, of course 🤦 . 

My personal stance on this is that I expect things to work upon load, so if I'm loading `Random` it should give me randomness--if we need to initialize a seed in order to do that then that seems like the correct path to me. That said, I think the concerns @jacereda raises about syscalls make sense, but it sounds like its being optimized away anyways (at least by the compiler you're using @hellerve?)

I guess if I had to make a definitive choice I'd say we should auto seed simply because it's more aligned with what I'd assume to be the natural expectations of users--I have a feeling most people don't expect to *have* to issue a seed to get some amount of legitimate randomness, but I could be wrong about that.

eriksvedang:
Interesting queston. I think I have left it out half-intentionally so far because it "felt right", kind of siding with @jacereda. Randomness is the kind of thing you should think about when adding to your program anyways, so making the programmer ponder when and what to seed with is beneficial to them, since the design of the program becomes more deliberate. 

I kind of like the usability of doing it automatically too, but it will be harder/messier to opt out of this kind of thing if it's built in.

eriksvedang:
Sateen sleeping on it I think that having it be automatic, but removing the seeding when —no-core might be the best.

hellerve:
@scolsen and I have thought about creating an alternative stdlib over at @carpentry-org (or multiple) for the performance-conscious, or maybe even more embedded-friendly. Since neither of us are experts in that, thoguh, I’m not sure whether it would make sense for us to do that alone.

eriksvedang:
@hellerve OK, that's a cool idea 👍

eriksvedang:
After agonizing a bit more I think we can merge this. If you care about syscalls (which is completely legit, of course) you wouldn't want to load the whole core library anyways. And if you expect a stable seed you're probably interested in controlling that yourself somehow anyways. So that leaves the common case of people just trying out the language and expecting things to "work". Which it will with this PR, so I'll merge it.

PR #946, opened 2020-10-26
Fixes inline-c and makes last argument optional

TimDeve:
It seems at some point inline-c became a noop (I believe when the way macros are evaluated has changed), this PR restores the functionality.

It also makes the last argument optional as, looking through the original PR, it seems that was the original intent.


hellerve (approved):

eriksvedang:
Thanks!

PR #947, opened 2020-10-26
Moves StdInt functions with dependency on String into String.carp

TimDeve:
Moves `StdInt` functions with dependency on String into String.carp, it allows to import `StdInt.carp` by itself in `--no-core` situations.
It is also more consistent as other number type string functions are already present in `String.carp`.


hellerve (approved):
Looks good to me!

eriksvedang:
Excellent!

PR #952, opened 2020-11-07
Removes usage of set-env & add-path from Github Actions workflows

TimDeve:
Both set-env and add-path are being deprecated as they are security vulnaribilities, see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

### DIFF
--- a/.github/workflows/carp.yml
+++ b/.github/workflows/carp.yml
@@ -15,14 +15,11 @@ jobs:
 
     - uses: actions/setup-haskell@v1
 
-    - run:   echo ::set-env name=GITHUB_SHA::$GITHUB_SHA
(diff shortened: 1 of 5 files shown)


TimDeve:
I've also disabled the cache for the MacOS build as it seems it only succeeds when the cache is missed.

From the logs this binary seems to be the culprit:
```
/Users/runner/.stack/setup-exe-cache/x86_64-osx/Cabal-simple_mPHDZzAJ_3.0.1.0_ghc-8.8.2: cannot execute binary file
```

eriksvedang:
Awesome, huge thanks for looking into this more!!!

Issue #851, opened 2020-05-28
Put scripts into a directory?


eriksvedang:
We have a *lot* of bash scripts in the root of the repo right now, perhaps we should do some spring cleaning and put them in their own directory? Would make the Github landing page a bit more welcoming, I think.

hellerve:
Yeah, that sounds good to me. We’d just have to update the CI scripts and such accordingly, but it shouldn’t be too much work.

TimDeve:
@eriksvedang I see you've assigned this to yourself, did you hit any snag? If not do you mind if I pick it up?

eriksvedang:
@TimDeve Oh please do!

TimDeve:
I believe this can be closed.

PR #955, opened 2020-11-09
Moves scripts in scripts folder

TimDeve:
Would close issue #851, should I put the nix files in their own folders as well? I'm not sure who depends on them.


TimDeve:
Kind of a separate topic but should we have `(Debug.sanitize-addresses)` in the benchmarks? Doesn't that make them slower? Or does it just use more memory?

hellerve:
I think @jacereda introduced the nix files.

hellerve:
Also, I would guess that it does make them slower.

jacereda:
What's the proposal? Moving nix stuff to `scripts` as well or move it to another directory? If it's another directory, I could just combine both .nix scripts into a single `default.nix` file. The main directory pollution would be a single file vs a single directory, so I guess it wouldn't be worth it. 

If we want those in the `scripts` directory it should be fine albeit a bit more cumbersome, since entering an environment would no longer be a `nix-shell` from the carp directory, it would require specifying the path to `shell.nix`. Not a big problem though.

Feel free to move them and I'll PR again with whatever fixes are required to make it work again on Nixpkgs.

eriksvedang:
`(Debug.sanitize-addresses)` has found a lot of bugs so far, so I don't think we should remove them...

TimDeve:
@jacereda I meant in their own folder, you're right if it's possible to merge them it would achieve the same outcome.

eriksvedang:
@TimDeve Oh, sorry I missed that you wrote *benchmarks*. Yeah lets remove it from there!

scolsen:
So glad the macOS CI is fixed!

PR #958, opened 2020-11-09
remove Pointer.prn

### LINKED ISSUE #956: Very Small Example doesn't build on MacOS High Sierra (Pointer_prn__SDL_window & Pointer_prn_SDL_Renderer)
When I try to build the very small sdl example on the front page, I get a few errors concerning the following functions:

Pointer_prn__SDL_window

Pointer_prn__SDL_Renderer

clang told me that there were some similar functions, and changing all references to the above functions to:

Pointer_copy_SDL_window
Pointer_copy_SDL_Render

fixed it and I was able to build successfully. I'm not sure if the prn versions are supposed to be there and are missing forward declarations or if the carp compiler is supposed to be generating the copy versions. Either the forward declarations for the prn variants should be added to the output c file or they should be changed to use the copy versions

hellerve:
This PR fixes #956 by removing `Pointer.prn` (please verify that it actually resolves it for you as well)!

Cheers

### DIFF
--- a/core/String.carp
+++ b/core/String.carp
@@ -462,5 +462,4 @@
 (defmodule Pointer
   (register str (Fn [(Ptr a)] String) "Pointer_str")
   (implements str Pointer.str)
-  (defn prn [x] (Pointer.str x))
 )


eriksvedang:
Thanks for looking into this!

I think the problem is that this implementation of `prn` for pointers is too simplistic, it has to be a generic function.

Adding this to the starting env works:

```
templatePointerPrn :: (String, Binder)
templatePointerPrn = defineTemplate
  (SymPath ["Pointer"] "prn")
  (FuncTy [RefTy (PointerTy (VarTy "p")) (VarTy "q")] UnitTy StaticLifetimeTy)
  "prn a pointer `p`."
  (toTemplate "void $NAME ($p** ptrRef)")
  (toTemplate $ unlines ["$DECL {"
                        ,"    return \"ptr\";"
                        ,"}"])
  (const [])
```

But perhaps we can do it with the new(ish) template macro instead?

PR #959, opened 2020-11-09
Run SDL tests on linux CI

### LINKED ISSUE #956: Very Small Example doesn't build on MacOS High Sierra (Pointer_prn__SDL_window & Pointer_prn_SDL_Renderer)

TimDeve:
We seem to have had a regression (#956) without noticing because we don't run the SDL examples/tests in the CI.

This PR enables them in the Linux CI.

#958 fixes #956, will rebase this PR once it is merged.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -15,6 +15,14 @@ jobs:
 
     - uses: actions/setup-haskell@v1
 
+    - name: Install SDL dependency (for SDL tests)
+      run: |
+        sudo apt-get update && sudo apt-get install libsdl2-dev       \
+                                                    libsdl2-gfx-dev   \
+                                                    libsdl2-image-dev \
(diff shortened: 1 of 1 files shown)


TimDeve:
Rebase done, CI looks good.

eriksvedang:
Awesome work @TimDeve !

PR #961, opened 2020-11-10
fix: Make nameOfPolymorphicFunction not add suffix to external functions

eriksvedang:


eriksvedang:
Still trying to make the original `(defn prn [x] (Pointer.str x))` work...

At least this doesn't remove the function completely, and solves *one* particular bug.

eriksvedang:
Looked into it a bit more!

The actual problem is in Deftype.hs, line 301 and 328 where we remove some deps of the deftype:s member `prn` function by filtering away those that are external.

I'll think a bit about whether we actually need to do that in some cases, otherwise I think we can remove that filter predicate.

hellerve (approved):

TimDeve:
@eriksvedang Think you forgot the `implements` call here.

PR #963, opened 2020-11-11
Echoes examples path when running run_carp_tests.sh

TimDeve:
Makes it easier to see which file failed in case of errors.

### DIFF
--- a/scripts/run_carp_tests.sh
+++ b/scripts/run_carp_tests.sh
@@ -10,7 +10,7 @@ fi
 
 ./scripts/build.sh
 
-# Build and run some examples
+echo "Build and run some examples"
 ./test/execute.sh ./examples/basics.carp
 ./test/execute.sh ./examples/functor.carp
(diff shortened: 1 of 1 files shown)


TimDeve:
The Windows failure seems to be caused by [526a5c157](https://github.com/carp-lang/Carp/commit/526a5c1573dc1d94a499ad37a3b1d11c86cd9b84).

jacereda:
Regarding the build failure, I think it would be better to leave the responsibility of generating a shared library to the user.  We could pass `/LD` for `clang-cl`, but that's just the beginning. Do we want to maintain a flags list for the dozens of C compilers out there? I think setting the compiler flags via `Project` is more than acceptable.

eriksvedang:
> I think setting the compiler flags via Project is more than acceptable.

Yeah, that makes a lot of sense – lets do that instead.

eriksvedang:
Nice!

PR #966, opened 2020-11-14
Fix nixpkgs build.

jacereda:


scripts/release.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
eriksvedang, line 1: This change did not work well with github CI when running the relase script (see https://github.com/carp-lang/Carp/runs/1403509009?check_suite_focus=true)

Do you know why @jacereda ?

eriksvedang: Figured it out! needed single `=` in string comparison

PR #967, opened 2020-11-16
Allow Unit to be used as a member type in deftypes

scolsen:
This commit enables support for using values of type () (Unit) in
user-defined types such as product and sumtypes. After this commit,
types such as:

    (deftype Units [action-one () action-two ()])

Are valid, and can be instantiated in the obvious way:

    (Units.init (IO.println "foo") ())

Some important things to note about the implementation:

- The C structs emitted for types containing Unit members *completely
  omit all unit members*. If a type in Carp has N members, the
  corresponding C struct will have (N-U) members where U is the number of
  members with the type `Unit`.

  For example, this type:
   
      (deftype (Foo [one Unit two Int]))

  will produce the following typedef in C:

      typedef struct {
        int two;
      } Foo;

  As a special case, types that *only* have Units as members are represented and
  initialized as completely empty structs:

      (deftype Foo [empty Unit])

      // emits

      typedef struct {
      } Foo;

      Foo Foo_init() {
        Foo instance = {};

        return instance;
      }

   Such a type is merely a container for side effects.

- Side effects are not stored at all in the types that contain Unit
  members. Instead, any side effects will be lifted out of the emitted C
  struct and called prior to initialization.

  For example, initializing `(deftype Foo [empty Unit])` with `(Foo.init
  (IO.println "foo"))` will produce the following C:

      main(...) {
        //...
        static String _10 = "foo";
        String *_10_ref = &_10;
        IO_println(_10_ref);
        Foo _12 = Foo_init();
        //...
      }

- The typical operations on product fields are supported on Unit type
  members, but they have slightly custom semantics. Since we don't
  actually store any values of type Unit in custom types, most
  updaters/getters/setters simply run a side effect.

  This is mostly only supported to make the use of such members more
  intuitive and allow programmers to chain side-effects within some
  context, much like monadic IO in Haskell.

- Match forms also work on Unit types for parity, but again, there is no
  meaningful behavior here, since Unit only has a single type
  inhabitant.

As a bonus, this commit also makes it possible to use `Unit` and `()`
interchangeably in type signatures.

### DIFF
--- a/src/Deftype.hs
+++ b/src/Deftype.hs
@@ -77,23 +77,45 @@ templatesForMembers _ _ _ _ _ = error "Shouldn't reach this case (invalid type d
 -- | Generate the templates for a single member in a deftype declaration.
 templatesForSingleMember :: TypeEnv -> Env -> [String] -> Ty -> (XObj, XObj) -> [((String, Binder), [XObj])]
 templatesForSingleMember typeEnv env insidePath p@(StructTy (ConcreteNameTy typeName) _) (nameXObj, typeXObj) =
-  let Just t = xobjToTy typeXObj
-      memberName = getName nameXObj
-  in [instanceBinderWithDeps (SymPath insidePath memberName) (FuncTy [RefTy p (VarTy "q")] (RefTy t (VarTy "q")) StaticLifetimeTy) (templateGetter (mangle memberName) t) ("gets the `" ++ memberName ++ "` property of a `" ++ typeName ++ "`.")
-     , if isTypeGeneric t
-       then (templateGenericSetter insidePath p t memberName, [])
-       else instanceBinderWithDeps (SymPath insidePath ("set-" ++ memberName)) (FuncTy [p, t] p StaticLifetimeTy) (templateSetter typeEnv env (mangle memberName) t) ("sets the `" ++ memberName ++ "` property of a `" ++ typeName ++ "`.")
-     , if isTypeGeneric t
-       then (templateGenericMutatingSetter insidePath p t memberName, [])
-       else instanceBinderWithDeps (SymPath insidePath ("set-" ++ memberName ++ "!")) (FuncTy [RefTy p (VarTy "q"), t] UnitTy StaticLifetimeTy) (templateMutatingSetter typeEnv env (mangle memberName) t) ("sets the `" ++ memberName ++ "` property of a `" ++ typeName ++ "` in place.")
-     ,instanceBinderWithDeps (SymPath insidePath ("update-" ++ memberName))
-                                                            (FuncTy [p, RefTy (FuncTy [t] t (VarTy "fq")) (VarTy "q")] p StaticLifetimeTy)
-                                                            (templateUpdater (mangle memberName))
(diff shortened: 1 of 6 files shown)


src/Deftype.hs
```diff
@@ -77,23 +77,45 @@ templatesForMembers _ _ _ _ _ = error "Shouldn't reach this case (invalid type d
...
-                                                            (FuncTy [p, RefTy (FuncTy [t] t (VarTy "fq")) (VarTy "q")] p StaticLifetimeTy)
-                                                            (templateUpdater (mangle memberName))
-                                                            ("updates the `" ++ memberName ++ "` property of a `" ++ typeName ++ "` using a function `f`.")
-                                                            ]
+  case t of
+    -- Unit member types are special since we do not represent them in emitted c.
+    -- Instead, members of type Unit are executed for their side effects and silently omitted
+    -- from the produced C structs.
+    UnitTy ->
+      binders (FuncTy [RefTy p (VarTy "q")] UnitTy StaticLifetimeTy)
```
eriksvedang, line 85: How do you feel about actually returning the (super weird!) type `(Ref Unit)` here? Maybe it makes no sense, but I recall having problems with the getters in conjunction with interfaces that rely on the return type being a ref. So that certain structs would not be possible to pass to certain functions requiring the interface, because the getter had a slightly different type. Doesn't seem to have popped up in the tests, so it might be nothing. But worth thinking about one extra time I think...

src/Deftype.hs
```diff
@@ -77,23 +77,45 @@ templatesForMembers _ _ _ _ _ = error "Shouldn't reach this case (invalid type d
...
+              (FuncTy [p, t] p StaticLifetimeTy)
+              (FuncTy [RefTy p (VarTy "q"), t] UnitTy StaticLifetimeTy)
+              (FuncTy [p, RefTy (FuncTy [] UnitTy (VarTy "fq")) (VarTy "q")] p StaticLifetimeTy)
```
eriksvedang, line 88: Perhaps in the same vein it would be fine (or even desirable) to have to pass a function `(Fn [Unit] Unit)` here? For higher level code to work, I mean. In the concrete case it's clearly a nuissance...

eriksvedang (commented):
Looks great, awesome work! I had one concern regarding getters, apart from that I think we're good to go.

src/Deftype.hs
scolsen: Hm, we could do that--however it might prove a bit tricky since `Unit` doesn't implement `copy`--so you wouldn't be able to copy such a ref...but maybe we don't need to? Or maybe we just implement `copy` on `Unit`?

src/Deftype.hs
scolsen: I run into a weird issue when I give the function a `Unit` argument:

```
out/main.c:6532:71: error: use of undeclared identifier 'Fn__LambdaEnv_void'
Foo Foo_update_MINUS_one(Foo p, Lambda *updater) { (*updater).env ? ((Fn__LambdaEnv_void)(*updater).callback)((*updater).env) : ((Fn___void)(*updater).callback)(); return p;}
```

Not sure where I need to update the code to ensure the env is declared

eriksvedang:
This is clearly good enough to merge, we can think about making my proposed changes later if we want to.

PR #976, opened 2020-11-18
Remove shell.nix.

jacereda:


jacereda:
As discussed in #955, this should clean up a bit more the project root.

eriksvedang:
Sweet!

jacereda:
I want to add a workflow for building on Linux+Nixpkgs. Should I just add the file to .workflows and PR? Is that enough to test if it will build?

jacereda:
Nevermind, looks like PRing is enough.

PR #977, opened 2020-11-18
Attempt to run CI jobs using Nixpkgs.

jacereda:


jacereda:
I guess it would be better to address the MacOS failure in a separate request. Maybe it's just a matter of removing the explicit version requests for stack/ghc?

Issue #965, opened 2020-11-12
Unit type in emitted C code is represented as () which is invalid C code


TimDeve:
When nesting two lambda returning () the generated C code contains `()` in the function names which is a compiler error.

Example:
```clojure
(defn main []
  (let [fun (fn [] (println* "Hello"))]
    ((fn [] (~&fun)))))
```
```sh
out/main.c:2252:25: error: expected function body after function declarator
Lambda Function_copy__()_Fn___() (Lambda* ref);
                        ^
out/main.c:2255:23: error: function cannot return function type 'void (Lambda)'
void Function_delete__() (Lambda f);
                      ^
out/main.c:6435:25: error: expected function body after function declarator
Lambda Function_copy__()_Fn___() (Lambda* ref) {
                        ^
3 errors generated.
```

TimDeve:
This works however:
```clojure
(defn main []
  (let [fun &(fn [] (println* "Hello"))]
    ((fn [] (~fun)))))
```

PR #980, opened 2020-11-19
Reify: Add Unit as a special case

scolsen:
Unit's show implementation defaults to the empty list `()`--this
presents problems when validating types, since generic type members are
replaced by the reification (a type literal) of their arguments. In the
`Unit` case, `()` yields some surprising invalid types.

So, we reify UnitTy as `Unit` instead to ensure consistency. This
permits using  Unit as a member in generic types like Result.


eriksvedang (approved):

eriksvedang:
Cool! Was there an issue connected to this?

scolsen:
> Cool! Was there an issue connected to this?

No issue, just a follow-up to #967 -- @TimDeve noticed it wasn't working properly w/ generics, and this was the root issue!

TimDeve:
It did fix https://github.com/carp-lang/Carp/issues/965

eriksvedang:
Awesome!

PR #983, opened 2020-11-20
Updates mutual recursion example

TimDeve:
Updates the example so it can be executed without blowing the stack.

Made it slightly closer to what an actual usage of mutual recursion would look like.


TimDeve:
Linux CI failure is due to what #985 fixes.

PR #984, opened 2020-11-20
Adding more documentation around C Interop

TimDeve:
Started with passing around `String`s across the C boundary.
After that I was gonna do `Array`s and `deftemplate`.
Please chime in with anything that would useful to have in there.

### DIFF
--- a/docs/CInterop.md
+++ b/docs/CInterop.md
@@ -4,6 +4,7 @@ This is an extension of what is covered in the [Language Guide](./LanguageGuide.
 
 ## Content
 
+- [How Carp generates identifiers]("#how-carp-generates-identifiers")
 - [Managed types](#managed-types)
   - [String](#string)
   - [Array](#array)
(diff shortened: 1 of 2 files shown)


eriksvedang:
Looks great to me!

docs/CInterop.md
```diff
@@ -0,0 +1,94 @@
...
+
+```clojure
+(register puts (Fn [(Ptr CChar)] ()))
+
+(let [a-str @"A string."]
+  (puts (String.cstr &a-str)))
+(puts (String.cstr "Hello"))
+```
+
+Might be a good idea to not exposed to the end-user the C type directly:
```
hellerve, line 26: Might I suggest the edit:

`It might be a good idea not to expose the C type to the end user directly:`

docs/CInterop.md
```diff
@@ -0,0 +1,94 @@
...
+  (private puts-c)
+  (register puts-c (Fn [(Ptr CChar)] ()) "puts")
+  (defn puts [str-ref] (puts-c (String.cstr str-ref))))
+
+(let [a-str @"A string."]
+  (MyMod.puts &a-str))
+(MyMod.puts "Hello")
+```
+
+If you are given a `char*` and want to turn it in a managed `String` you can
```
hellerve, line 40: Maybe `turn it into` would be better here?

hellerve (commented):
Looks great to me so far!

docs/CInterop.md
```diff
@@ -0,0 +1,165 @@
+# C Interop
+
+This is an extension of what is was covered in the [Language Guide](./LanguageGuide.md#c-interop).
```
TimDeve, line 3: I've noticed the "is was", will remove.

docs/CInterop.md
wltsmrz: I think that sentence is relatively easily misread. Possible alternatives:

* `You may want to hide the C type from the end-user`
* `It might not be a good idea to expose the C type to the end user directly`

docs/CInterop.md
```diff
@@ -0,0 +1,165 @@
...
+  (println* a-str))
+```
+
+If you are the one writing the C code, you can use the `CARP_MALLOC` macro to
+ensure you are using the same allocator as the Carp compiler:
+
+```c
+char* returns_a_heap_string() {
+  char *hello = "Hello from the heap";
+  char *str = CARP_MALLOC((strlen(hello)+1));
```
wltsmrz, line 93: Wonder if we should use `String` here as Carp does internally https://github.com/carp-lang/Carp/blob/master/core/carp_string.h

wltsmrz: Although for the purposes of this document maybe the current state is better

TimDeve: These functions are meant as examples of functions from an external library, would be unlikely for them to have `String` as the return type.
I might mention it in the next section about wrapping functions in custom C code.

TimDeve:
I think this is ready for review, I want to talk about `inline-c` and wrapping macros but I can do that in a separate PR.

docs/CInterop.md
```diff
@@ -0,0 +1,285 @@
...
+Let's break down what's going on here:  
+The **first** argument to `deftemplate` is the name we'll use to refer to the
+function.  
+The **second** is a type signature and is identical to the one found
+in our previous `register` call.  
+The **third** is our function declaration, it'll be injected at the top of the
+generated C file.  
+The **last** argument represent the function definition.
+
+Two more thing to look at:
```
jacereda, line 226: s/thing/things/

jacereda:
Nice!

scolsen (commented):
Looks great!

docs/CInterop.md
```diff
@@ -0,0 +1,285 @@
...
+(puts (String.cstr "Hello"))
+```
+
+You may want to hide the C type from the end-user:
+
+```clojure
+(defmodule MyMod
+  (hidden puts-c)
+  (private puts-c)
+  (register puts-c (Fn [(Ptr CChar)] ()) "puts")
```
tealeg, line 42: It would be nice to have some commentary on why we have to override the `puts-c` declaration manually here - I got into all kinds of trouble because I missed that out when registering a function like this inside a module (the generated C code ended up referencing non-existent functions).

docs/CInterop.md
```diff
@@ -0,0 +1,285 @@
...
+(relative-include "printall.h")
+
+(register println-all (Fn [(Ptr String) Int] ()) "println_all")
+
+(let [lines [@"One" @"Two" @"Three"]
+      len (Array.length &lines)]
+  (println-all (Array.raw lines) len))
+```
+
+## Embedding C code in Carp
```
tealeg, line 176: It would be useful to have an explanation of how to call or reference a Carp function from C.

docs/CInterop.md
TimDeve: I mean in this specific case it's because I want the function to be called `puts-c` but I'll add a section about how identifiers gets generated.

docs/CInterop.md
TimDeve: Will be covered in the identifiers section.

TimDeve:
@eriksvedang Do you want me to squash the commits?

PR #986, opened 2020-11-20
More Unit type member enhancements

scolsen:
- Specialize generic setter functions against Unit members
- Specialize calls to str/prn against Unit members in StructUtils
  (rather than accessing the struct member, which doesn't exist in the
  Unit case, we simple call Unit's prn function.)
- Don't bind references to Unit values to variables. This fixes an error
  whereby a reference to a Unit would generate invalid C such as:
  `void* _9 = &;`
  Likewise, we omit references to Units from function arguments just as
  we omit Unit values.

- Add `prn`, `copy`, and `zero` implementations for `Unit`

### DIFF
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -471,12 +471,14 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
               do var <- visit indent value
                  let Just t' = t
                      fresh = mangle (freshVar i)
-                 if isNumericLiteral value
-                   then do let literal = freshVar i ++ "_lit";
-                               Just literalTy = ty value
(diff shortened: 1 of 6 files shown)


src/Emit.hs
```diff
@@ -562,7 +564,7 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
               do argListAsC <- createArgList indent (mode == ExternalCode) args
                  let Just (FuncTy _ retTy _) = ty func
                      funcToCall = pathToC path
-                 if retTy == UnitTy
+                 if not (notUnit retTy)
```
jacereda, line 567: I guess it could be a separate PR, but double negation always takes me a while to grok. 

This could just be `isUnit retTy` and filters could use `not . isUnit`

eriksvedang: Very much agree! I’ll merge this now, but let’s make that change.

PR #988, opened 2020-11-21
docs: Remove a lot of duplicated information and refer to webpages instead

### LINKED ISSUE #852: Help (help)
I think that the `(help)` command is a good idea in theory, but it's a bit unwieldy right now. Perhaps someone has ideas for how to make it more useful? And who is it for? (beginner/expert/both?) Should it be more of a collection of links to the docs?

I think that making it possible to write `(help macros)` rather than `(help "macros")` could be a good thing. If it even makes sense to have these "chapters"?

eriksvedang:
This closes #852.


eriksvedang:
Hmm, it seems like the package `open-browser` isn't available in Nix? (ping @jacereda)

eriksvedang:
What would be a reasonable workaround?

jacereda:
It should be there, but default.nix was generated at some point and tweaked afterwards. At some point it would be good to autogenerate it from the cabal file, but while we get there just add `open-browser` to `default.nix` the same way as, say, `ansi-terminal`. Be careful, it's mentioned twice in that file.

eriksvedang:
@jacereda I see, thanks!

PR #989, opened 2020-11-21
chore: Move some examples to test/produces-output

eriksvedang:
Will make it easier for people to find the interesting bits in the 'examples' directory.


TimDeve:
You've got a few .DS_Store files in there.

eriksvedang:
@TimDeve Oh, dang! Thanks 👍

scripts/run_carp_tests.sh
```diff
@@ -43,7 +43,6 @@ echo "Compile-only examples"
 compileOnlyExamples="\
                    examples/mutual_recursion.carp \
                    examples/guessing_game.carp \
-                   examples/unicode.carp \
```
jacereda: We should restore this one if #994 is applied.

eriksvedang: Awesome, I'll put it back!

eriksvedang:
Also closes #991.

eriksvedang:
Oh, snap – we don't have `clock_gettime` on Windows so the benchmarks won't build. Do we have a workaround for that already?

TimDeve:
Could steal a replacement implementation from stack overflow...

jacereda:
> Oh, snap – we don't have `clock_gettime` on Windows so the benchmarks won't build. Do we have a workaround for that already?

I think I want to take a look at that after this gets merged. IMHO `System_nanotime` should return integer nanoseconds. Keep in mind things like the GBA don't have FP hardware and any FP operation there is extremely expensive.

eriksvedang:
> > Oh, snap – we don't have `clock_gettime` on Windows so the benchmarks won't build. Do we have a workaround for that already?
> 
> 
> 
> I think I want to take a look at that after this gets merged. IMHO `System_nanotime` should return integer nanoseconds. Keep in mind things like the GBA don't have FP hardware and any FP operation there is extremely expensive.

Yes please! This was just a quick hack.

PR #995, opened 2020-11-22
Add u32 and u64 literals

wltsmrz:
Thoughts on adding u32/u64 literals?

As the PR is concerned, should add tests and I noticed that due to this rounding thing max u64 isn't representable. I'm not really sure how to proceed with that (I don't actually know Haskell). But if there's interest I'll try to figure it out.

```haskell
             Num UInt64Ty num -> return (show (round num :: Integer) ++ "llu")
```


eriksvedang:
> Thoughts on adding u32/u64 literals?

I guess we might as well :D 

@jacereda are fixing the things needed in the dynamic runtime to cater to this, will merge this once those fixes are in!

eriksvedang:
Now those tests should be possible to write, I hope!

wltsmrz:
After thinking about this a bit more, I convinced myself that I don't really like the idea of only adding u32/u64 (and not for example s32). I also didn't really like the idea of adding `u` and `ul` to accompany regular int literals and `l` although slightly more appealing than first idea. I'll close now, maybe we can think about it more sometime in the future.

eriksvedang:
alright!

Issue #753, opened 2020-05-06
[RFC] Make it possible to initialize structs with just the module name


eriksvedang:
When defining a sumtype you get a bunch of constructors for the various cases, like this:
```clojure
(deftype Sum A B)

(Sum.A)
(Sum.B)
```

When defining a product type you have to use `init` instead:

```clojure
(deftype Prod [x Int])

(Prod.init 123)
```

My suggestion is to allow the following as an alternative syntax:

```clojure
(Prod 123)
```

Right now this gives an error about trying to "call" the module, so hopefully we can just insert the call to init there instead (I haven't really looked into how to do the actual implementation).

As an added bonus, this would make the printing and reading of product types symmetrical also, since right now their string representation lacks the `.init` part:

```clojure
(str (Prod 123))
;;; =>
(Prod 123)
```

TimDeve:
Slightly unrelated but I'm wondering if we should ape Rust for the typedef, so you can declare Prod as: 
```clojure
(deftype Prod [Int])
```
Pair could just become:
```clojure
(deftype Prod [a b])
```

hellerve:
@TimDeve would all members be unnamed then?

TimDeve:
I was thinking it would behave the same as here (numbered fields):
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a70992ebc6734b79332e6205928fbfb7

hellerve:
I don’t really enjoy that, because it forces me to match on the entire thing if I want to change/update one field, and that gets very unwieldy on big types. I enjoy being able to have named getters and setters to update individual fields, and I actually think that the unnamed field pattern in Rust and Haskell is an antipattern. Consider, for instance the `DateTime` datatype underlying my `time` library:

https://github.com/carpentry-org/time/blob/c68425f37e102be10979d1dcfd453eadc2c23141/time.carp#L19-L28

This would become:

```clojure
(deftype Datetime [
  Int,
  Int,
  Int,
  (Maybe Int),
  (Maybe Int),
  (Maybe Int),
  (Maybe Int),
  (Maybe Timezone),
])
```

Except for the timezone, which is which? We can hope that the struct is ordered and the highest unit comes first, but we could also have opted for a US date format (day before month) or something like it. It unnecessarily creates confusion.

[...]

Issue #930, opened 2020-09-24
Support defining types in modules


scolsen:
Currently, all Carp types are global which can lead to odd behavior when defining them in modules. As a basic illustration of this behavior, consider the following example:

```clojure
(defmodule Homotypic (deftype (Pair a a) [fst a snd a]))
鲤 :i Pair
Homotypic.Pair : Type
Defined at line 1, column 32 in 'REPL'
Pair : Module = { ;; seems to just be the original Pair from core. 
    < : (Fn [(Pair a b), (Pair a b)] Bool)
    = : (Fn [(Pair a b), (Pair a b)] Bool)
    > : (Fn [(Pair a b), (Pair a b)] Bool)
    a : (Fn [(Ref (Pair a b) q)] (Ref a q))
    b : (Fn [(Ref (Pair a b) q)] (Ref b q))
    copy : (Fn [(Ref (Pair a b) q)] (Pair a b))
    delete : (Fn [(Pair a b)] ())
    hash : (Fn [(Ref (Pair a b) c)] Int)
    init : (Fn [a, b] (Pair a b))
    init-from-refs : (Fn [(Ref a b), (Ref c d)] (Pair a c))
    prn : (Fn [(Ref (Pair a b) q)] String)
    reverse : (Fn [(Ref (Pair a b) c)] (Pair b a))
    set-a : (Fn [(Pair a b), a] (Pair a b))
    set-a! : (Fn [(Ref (Pair a b) q), a] ())
    set-b : (Fn [(Pair a b), b] (Pair a b))
    set-b! : (Fn [(Ref (Pair a b) q), b] ())
    str : (Fn [(Ref (Pair a b) q)] String)
    update-a : (Fn [(Pair a b), (Ref (Fn [a] a fq) q)] (Pair a b))
    update-b : (Fn [(Pair a b), (Ref (Fn [b] b fq) q)] (Pair a b))
    zero : (Fn [] (Pair a b))
}
Defined at line 91, column 11 in '/Users/scottolsen/dev/Carp/core/Tuples.carp'
Documentation: is a 2-tuple, i.e. a datatype with two members.
鲤 :i Homotypic.Pair
Pair : Module = { ;; original Pair from core + Homotypic Pair?!
    < : (Fn [(Pair a b), (Pair a b)] Bool)
    = : (Fn [(Pair a b), (Pair a b)] Bool)
    > : (Fn [(Pair a b), (Pair a b)] Bool)
[...]

PR #1003, opened 2020-11-22
feat: Implicit struct init

eriksvedang:
Closes https://github.com/carp-lang/Carp/issues/753.

Example of what you can do:

```clojure
(defmodule Wrap
  (deftype A [x Int y Int]))

(use Wrap)

(A 3 4)
```

(no `.init`)


eriksvedang:
This also fixes a bug where deftypes (including sumtypes) would get the global environment as their parent env. Possibly this leads to fixing some old bug regarding lookup..!

jacereda (approved):
Nice!

scolsen:
> This also fixes a bug where deftypes (including sumtypes) would get the global environment as their parent env. Possibly this leads to fixing some old bug regarding lookup..!

ohhh sweet--that will maybe (?) fix a lot of issues with submodules too

scolsen:
Also seems like this fixes #930 based on the test

PR #1004, opened 2020-11-22
Split integer/floating point number representation.

jacereda:
Something like this would help with problems like the one mentioned in #995, since 64bit integers won't be representable with a `Double`.


eriksvedang:
Right, makes sense! 

Do you think this has any implications for the dynamic runtime? Like, is there some situation where we could anticipate having to auto convert between INum and FNum somewhere?

jacereda:
I'm not sure yet, I'm still exploring. I'll send another PR where both are kept as `Num` but the representation is `data Number = Integral Int | Floating Double` so we can get a better idea.

jacereda:
Closed in favour of #1006

PR #1005, opened 2020-11-22
Eval: Add Address to eval

scolsen:
The `address` keyword is useful when one needs to circumvent Carp's
memory management for particular cases. However, the evaluator
previously didn't understand `address` in most contexts, so it proved
difficult to toy around with in the REPL.

This commit adds basic support for calling `address` in the REPL by
adding a case to the evaluator. This allows users to test out functions
that take `Ptr` arguments such as `Pointer.to-value`. The `address`
keyword only accepts a symbol as an argument (since getting the address
of a literal value doesn't always make sense (e.g. &2) and produces
invalid C):

    (def a 1)
    (address a)
    => a
    (type (address a))
    => (Ptr Int)
    (Pointer.to-value (address a))
    => 1

Note: one will still typically need to wrap it in a function in
higher-order contexts.

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -287,6 +287,8 @@ eval ctx xobj@(XObj o i t) preference =
                  Right _ -> eval ctx x preference
        [XObj While _ _, cond, body] ->
          specialCommandWhile ctx cond body
+       [XObj Address _ _, value] ->
+         specialCommandAddress ctx value
        [] -> return (ctx, dynamicNil)
        x -> do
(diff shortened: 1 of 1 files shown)


scolsen:
It'd be nice if calling `address` printed the address instead of returning the symbol, since this is what we do in emitted code (to try it out, set the value of a deftype inhabitant to a pointer)--but I'm not sure how to do that in the evaluator.

eriksvedang:
Neat!

PR #1006, opened 2020-11-22
Split integer/floating point number representation (second alternative)

jacereda:
An alternative to #1004 

I think I prefer this one. Look at `commandPlus` and friends, looks better in this branch IMHO.


eriksvedang:
Yeah this is great!

scolsen:
Nice refactor!

PR #1008, opened 2020-11-23
Meta set fix and refactor

### LINKED ISSUE #842: Setting meta on interfaces in modules fails
As discovered in #833, setting `meta` on `definterface` implementations in modules currently makes the compiler hang. @scolsen seems to have a hunch where this is coming from (for reference, [here is his comment](https://github.com/carp-lang/Carp/pull/833#discussion_r429683191)).

Cheers

scolsen:
## First commit

This commit fixes a subtle bug whereby setting the meta of an existing
binder would cause hangs. Ultimately, this still points to an underlying
issue in our Lookup code that causes such loops, but for now this at
least fixes the hang introduced by the doc call in `core/Tuples.carp`
(see issue #842).

The primary fix seems to be related to setting the parentEnv in a case
in which we failed to do so in `Eval.hs`. Additionally, our meta setting
macros call `eval` which causes them to be evaluated *immediately after*
expansion, causing them to be evaluated in the incorrect context in the
Tuples.carp case.

Additionally:

- Refactored primitiveMetaSet and primitiveMeta to make them cleaner.
- Only set `implements` early when we're certain we won't accidentally
  overwrite the interface.
- Generalize DocStub to `MetaStub` so that it may be used for forward
  meta declarations of any kind.

---
## Second Commit

Calling eval in the body of the meta-set! macros can cause them to be
evaluated before anticipated, possibly setting the meta on a binding in
the incorrect environment. An exemplary case of this issue existed in
`Tuples.carp` (also fixed in this commit) whereby the generated
defmodule for a tuple type called doc, which was evaluated *before* the
emitted module, resulting in overwrites of global name docs instead of
the expected module function.

We retain `evals` in macros that are more useful in the repl, such as
`print-doc`. If a user wants to evaluated one of the meta-set macros in
the REPL, they'll need to add a call to eval.

Fixes #842


scolsen:
Hmm turns out removing the `evals` from macros wasn't as innocuous as I thought--it seems to produce the same results when something is loaded but not when executed. I'll have to think some more about it.

scolsen:
Turns out the `evals` in `Macros.carp` are needed after all... at any rate this PR will still break the hang reported in issue #842 -- I've also changed the doc call in `Tuples` to directly emit `meta-set!` so that we don't eval it too soon.

src/Eval.hs
```diff
@@ -600,10 +600,10 @@ primitiveDefmodule xobj ctx@(Context env i typeEnv pathStrings proj lastInput ex
   (newCtx, result) <-
     case lookupInEnv (SymPath pathStrings moduleName) env of
       Just (_, Binder _ (XObj (Mod innerEnv) _ _)) -> do
-        let ctx' = Context env (Just innerEnv) typeEnv (pathStrings ++ [moduleName]) proj lastInput execMode history -- TODO: use { = } syntax instead
+        let ctx' = Context env (Just innerEnv{envParent=i}) typeEnv (pathStrings ++ [moduleName]) proj lastInput execMode history -- TODO: use { = } syntax instead
```
eriksvedang, line 603: Can we rename `i` to `inner` or something similar here? (since `i` means "info" everywhere else).

eriksvedang (commented):
Looks great! There are some merge errors now after the switch to `pure` everywhere.

src/Eval.hs
scolsen: done!

PR #1009, opened 2020-11-23
Replace return by pure.

jacereda:
OK, this is a monster, but I think it's the way to go...


hellerve (approved):
Did I just go through 1,000+ lines of what is essentially—in most cases anyway—a one word change? Why, yes, I did.

And it looks splendid to me!

src/Commands.hs
```diff
@@ -293,23 +293,203 @@ commandListBindings ctx args =
               putStrLn "\nGlobal environment:\n"
               putStrLn (prettyEnvironment (contextGlobalEnv ctx))
               putStrLn ""
-              return (ctx, dynamicNil)
+              pure (ctx, dynamicNil)
+
+-- | Command for printing help.
```
eriksvedang, line 298: Why was this moved back here? I think `(help)` should be a primitive so that you can access the chapters without quoting their names.

eriksvedang (changes requested):
One word – EPIC!
I detected one problem regarding the recently refactored `help`, apart from that it looks absolutely fantastic.

src/Commands.hs
jacereda: Oh, I messed it with the rebase. Should be fixed now.

eriksvedang (approved):

PR #1010, opened 2020-11-23
docs: Better credits and contribution guide

eriksvedang:
Maintaining the contributors list seemed like a chore when Github already supports that feature in such a nice way.

More ideas for what should go into Contributing.md are most welcome!

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -1,16 +1,15 @@
-Contributing
-============
+## Contributing Guide
 
-Thank you for considering contributing to Carp.
+Thank you for considering contributing to Carp!
 
 This document currently focuses on developer contributions but we welcome all kinds of contributions.
 
-
-Community
----------
+### Community
 The best place to start is to join the Carp Gitter channel over at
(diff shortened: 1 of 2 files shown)


docs/Contributing.md
```diff
@@ -1,16 +1,15 @@
...
-Community
----------
+### Community
 The best place to start is to join the Carp Gitter channel over at
 [https://gitter.im/carp-lang/Carp](https://gitter.im/carp-lang/Carp).
 
+### Understanding the Compiler
+See [Hacking.md](hacking.md) for a gentle introduction to the Carp compiler internals.
 
 License
```
hellerve, line 14: Should we also change this heading, then?

hellerve (approved):

docs/Contributing.md
eriksvedang: You mean Hacking.md becomes UnderstandingTheCompiler.md ?

hellerve: No, this was on the `License` heading which is still the old MD style.

eriksvedang: Now they all use two '#' (if that's what you meant)

PR #1011, opened 2020-11-23
Compile with -Wall.

jacereda:
This builds on top of #1009 

I'll try to address different warning types one by one to see how far we get.

### DIFF
--- a/.github/workflows/nixpkgs.yml
+++ b/.github/workflows/nixpkgs.yml
@@ -12,5 +12,5 @@ jobs:
     - uses: actions/checkout@v2.3.4
     - uses: cachix/install-nix-action@v12
       with:
-        nix_path: nixpkgs=channel:nixos-unstable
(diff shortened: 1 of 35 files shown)


jacereda:
The remaining warnings will probably be more invasive. I guess it'll be better to tackle those after the low hanging fruit is in place.

jacereda:
The first relevant commit in this branch is `Prepare for -Wall`.

eriksvedang:
Very nice! Seems like it got a little confused after merge:ing #1009.

jacereda:
Rebased.

jacereda:
Looks like the Nixpkgs workflow is using the unstable channel and `SDL-mixer` was broken on that channel. Let's try with the stable channel.

hellerve (approved):
Fantastic! So many less code smells!

eriksvedang:
Feels good.

PR #1014, opened 2020-11-24
refactor: Move code out of Macros.carp into other files

### LINKED ISSUE #934: Move macros into domain/semantic modules?
`Macros.carp` has gotten pretty big and has a variety of macros for different purposes. Some of them might fit better into other modules that now exist, for example, control macros such as `for` and `doto` might make more sense in the newer `Control` module. Macros like `windows-only` might be better suited for the `System` module.

However, such a change is not totally negligible as it would probably result in breakages in downstream code (since macros will be defined in a module, instead of at the top level). We'd also have to change the load order of some modules in `Core.carp`. 

Since the change isn't functionally necessary and is a semantic/communication preference, I figured we can open it for discussion here.

eriksvedang:
Closes #934 (at least it's a start).


core/Interfaces.carp
```diff
@@ -1,5 +1,3 @@
-(load-once "Macros.carp")
```
hellerve, line 1: Does this not depend on `Macros.carp` anymore?

core/ArrayMacros.carp
```diff
@@ -0,0 +1,44 @@
+(defmacro doall [f xs]
```
hellerve, line 1: Could this just go into `Array.carp`, maybe?

hellerve (approved):
Looks good to me! Some minor nitpicks, but nothing crucial :)

jacereda:
The nixpkgs build failure is fixed in #1011

core/Interfaces.carp
eriksvedang: It does, but we don't load other core files transitively but rather depend on their order in `Core.carp`. Or at least we probably shouldn't do that – to make it (arguably) easier to write your own prelude. Makes sense?

core/ArrayMacros.carp
eriksvedang: yes!

core/Interfaces.carp
hellerve: Totally; it’s just worth noting that just loading `core/Interfaces.carp` won’t work then because it implicitly expects `Macros.carp`, which is something to keep in mind when writing your own prelude.

eriksvedang: Alternatively we could make sure that each file in Core (transitively) imports everything it needs. That would be good in that you can just import one higher level module (like e.g. "Test.carp") and be sure that it works.

core/ArrayMacros.carp
eriksvedang: Or oh, you mean the whole file? Yeah I tried that at first but there were macros that were used by the stack allocated Array that made it not work. I'll try again though.

core/Cadr.carp
```diff
@@ -0,0 +1,29 @@
+(defmodule Dynamic
```
scolsen, line 1: Should we just add these to `List.carp` as well?

eriksvedang:
@scolsen Does this make sense? I tried to come up with a logical way of dividing this up, not sure if I'm happy with it though. Any suggestions on how to improve upon it are most welcome! The whole endeavor would be a lot easier if we had dynamic `use`, so might be good to do a second pass once that is implemented.

core/Core.carp
```diff
@@ -18,6 +18,12 @@
 (load-once "Interfaces.carp")
 (load-once "Bool.carp")
 (load-once "Macros.carp")
+(load-once "List.carp")
```
scolsen, line 21: Missing `Cadr.carp`? (Or maybe I'm not seeing where it's loaded)

core/Implement.carp
```diff
@@ -0,0 +1,8 @@
+(defndynamic implement-declaration [mod interface]
```
scolsen, line 1: Do you think these could live in the general `Macros.carp` or in `Interface.carp`? Since they're just helpers for implementing interfaces, seems odd to have their own file. But if this makes more sense from a --no_core perspective, it's fine!

scolsen (commented):
Great!

core/Cadr.carp
eriksvedang: There was some dependency that made it hard I think, but I'll try again.

core/Implement.carp
eriksvedang: Heh, sure – maybe I went a little overboard with splitting up stuff :) I just felt that Macros.carp became such a mix of wildly different things, but maybe that's OK.

PR #1018, opened 2020-11-24
Address unused local bind.

jacereda:
This one probably requires closer inspection. I tried to leave debugging helpers commented, but maybe you prefer to leave them uncommented.

### DIFF
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -303,7 +303,6 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
               let indent' = indent + indentAmount
                   retVar = freshVar i
                   isNotVoid = t /= Just UnitTy
-                  sumTypeAsPath = SymPath [] (show exprTy)
                   exprTy = exprTyNotFixed
 
                   tagCondition :: String -> String -> Ty -> XObj -> [String]
@@ -313,8 +312,6 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
                     -- This requires a bunch of extra machinery though, so this will do for now...
                     [var ++ periodOrArrow ++ "_tag == " ++ tagName caseTy (removeSuffix caseName)] ++
                       concat (zipWith (\c i -> tagCondition (var ++ periodOrArrow ++ "u." ++ removeSuffix caseName ++ ".member" ++ show i) "." (forceTy c) c) caseMatchers ([0..] :: [Int]))
-                      where notUnitX (XObj _ _ (Just UnitTy)) = False
-                            notUnitX _ = True
                   tagCondition _ _ _ x =
                     []
                     --error ("tagCondition fell through: " ++ show x)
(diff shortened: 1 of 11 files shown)


src/Emit.hs
```diff
@@ -313,8 +312,6 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
                     -- This requires a bunch of extra machinery though, so this will do for now...
                     [var ++ periodOrArrow ++ "_tag == " ++ tagName caseTy (removeSuffix caseName)] ++
                       concat (zipWith (\c i -> tagCondition (var ++ periodOrArrow ++ "u." ++ removeSuffix caseName ++ ".member" ++ show i) "." (forceTy c) c) caseMatchers ([0..] :: [Int]))
-                      where notUnitX (XObj _ _ (Just UnitTy)) = False
```
scolsen, line 316: Ah, this is possibly a bug--probably we want to filter Units from `caseMatchers` here. Mind adding a todo to double check this?

src/Eval.hs
```diff
@@ -483,9 +483,7 @@ catcher ctx exception =
 
 specialCommandWith :: Context -> XObj -> SymPath -> [XObj] -> IO (Context, Either EvalError XObj)
 specialCommandWith ctx xobj path forms = do
-  let pathStrings = contextPath ctx
```
scolsen, line 486: pathStrings use here is probably also a potential bug/cleanup--I think for almost all if not all of these lookups we should not be using `path` but rather:

`consPath (union (contextPath context) modulePrefixes) path`

where `modulePrefixes` are the list of prefixes in the given `SymPath` argument--this ensures we lookup the right things when we're evaluating something in some local context. I could be wrong about this, though @hellerve @eriksvedang

src/GenerateConstraints.hs
```diff
@@ -89,8 +89,7 @@ genConstraints globalEnv root rootSig = fmap sort (gen root)
                                 trueType <- toEither (ty ifTrue) (ExpressionMissingType ifTrue)
                                 falseType <- toEither (ty ifFalse) (ExpressionMissingType ifFalse)
                                 let expected = XObj (Sym (SymPath [] "Condition in if value") Symbol) (info expr) (Just BoolTy)
-                                let lol = XObj (Sym (SymPath [] "lol") Symbol) (info expr) (Just BoolTy)
```
scolsen, line 92: heh, I always wondered about this :)

scolsen (commented):
Generally looks good  to me!

Left a few comments regarding missing uses of bindings that I think are more likely bugs than not, but we don't need to address those here. 

Awesome cleanup!

src/GenerateConstraints.hs
eriksvedang: 🤔

src/Emit.hs
jacereda: Done. Would it need `isUnit` or the one removed that just checks plain units?

src/GenerateConstraints.hs
jacereda: Probably some test by @hellerve, introduced in 6e147598a.

src/Emit.hs
scolsen: iirc the list was a list of xobjs in this case, so `isUnit` will work but will need to be composed w/ `forceTy`, I think `filter (isUnit . forceTy)` should do it

src/Eval.hs
eriksvedang: Hmm I'm not sure without thinking more about it. Maybe create an issue for it? @scolsen

eriksvedang (approved):

PR #1019, opened 2020-11-24
Address unused matches.

jacereda:
This one could be controversial. For one, it might help locate bugs. On the other hand, it can make some parts a bit more obscure.


jacereda:
Somehow I managed to break Haskell's "if it compiles it works" here. Let me check...

jacereda:
And I messed it in a place that would be caught by `-Wname-shadowing` :)

scolsen:
> Somehow I managed to break Haskell's "if it compiles it works" here. Let me check...

Weird that it typechecked--I'm guessing a variable in a larger scope of the same type shadowed one of the removed names?

scolsen:
> And I messed it in a place that would be caught by `-Wname-shadowing` :)

We probably should turn that on--I have a feeling there's probably a lot of cases of this since we have so many local definitions

eriksvedang (approved):
Awesome!

eriksvedang:
Needs rebase.

jacereda:
Rebased.

PR #1026, opened 2020-11-26
Adds functions to help with callbacks based C interop

TimDeve:
I'm not sure if this the best way to go at it. This PR is more of a request for feedback.

Enables to bind to callback based API, like `pthread_create`:

```clojure
(register create-c (Fn [(Ref Thread) (Ptr ()) (Ptr ()) (Ptr ())] Int) "pthread_create")

(sig create-from-handle (Fn [(Ref Thread) (Ref (Fn [] ()))] (Result () Int)))
(defn create-from-handle [thread fun]
  (let [error-code (create-c thread
                             NULL
                             (Pointer.from-fn fun)
                             (Pointer.env-from-fn fun))]
    (if (= 0 error-code)
        (Success ())
        (Error error-code))))
```

### DIFF
--- a/core/Core.carp
+++ b/core/Core.carp
@@ -21,6 +21,7 @@
 (load-once "Introspect.carp")
 (load-once "Pointer.carp")
 (load-once "Unsafe.carp")
+(load-once "Function.carp")
 (load-once "Generics.carp")
 (load-once "Maybe.carp")
 (load-once "Result.carp")
--- a/core/Function.carp
+++ b/core/Function.carp
@@ -1,0 +1,7 @@
+(defmodule Function
+  (doc unsafe-ptr "returns void pointer to the function passed in."
+                 "This is unsafe as unsafe-ptr can't check the value passed in is a function.")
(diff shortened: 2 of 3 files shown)


eriksvedang (approved):
I like it!

jacereda:
What about something like `Callback.fn` and `Callback.env` for the names?

TimDeve:
> What about something like `Callback.fn` and `Callback.env` for the names?

I've put it in Pointers because it's inherently unsafe like most stuff in that module (actually had it in Unsafe originally). I'm a bit afraid to have it be part of "normal" carp but I'm open to changing it.

eriksvedang:
Perhaps they should be part of the (already existing) `Function` module?

TimDeve:
`Function.unsafe-ptr` & `Function.unsafe-env-ptr` ?

eriksvedang:
> `Function.unsafe-ptr` & `Function.unsafe-env-ptr` ?

Sure, looks good! I think it makes sense with the unsafe prefixes since you really have to know what you're doing :)

jacereda (approved):

PR #1033, opened 2020-11-29
Refactor error reporting and add CARP_DIR check

### LINKED ISSUE #1002: Warn about nonexisting CARP_DIR environment variable at startup
We kind of do warn about it because you can’t load core but a friendly message would be better...

hellerve:
This PR fixes #1002 by adding checking for `CARP_DIR` being set. If it’s not set, it will emit a warning. If it isn’t set and `--no-core` isn’t set either, it will also emit an error and exit.

While @TimDeve and I were at it, we also refactored the error printing to make it a little more ergonomic.

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -2,7 +2,8 @@ module Main where
 
 import qualified System.Environment as SystemEnvironment
 import System.Console.Haskeline (runInputT)
-import Control.Monad (foldM)
+import System.Exit (exitFailure)
+import Control.Monad (foldM, when)
 import GHC.IO.Encoding
 import Data.Maybe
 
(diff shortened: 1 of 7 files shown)


scolsen:
nice! looks like the builds failed because`CARP_DIR` isn’t set in our build environments.

TimDeve:
> nice! looks like the builds failed because`CARP_DIR` isn’t set in our build environments.

We did mention we needed to update the test script...

eriksvedang:
Nice refactor of the printing functions!

eriksvedang (approved):

eriksvedang:
> > nice! looks like the builds failed because`CARP_DIR` isn’t set in our build environments.
> 
> 
> 
> We did mention we needed to update the test script...

Can we solve this in the same PR to make sure there are no issues doing so?

TimDeve:
Yeah, we just forgot to 😅. It's on Veit's branch so he would need to do it.

jacereda (approved):

jacereda:
> We did mention we needed to update the test script...

Probably setting it in `carp.sh` is enough?

TimDeve:
> Probably setting it in `carp.sh` is enough?

Sounds good to me.

hellerve:
I fixed the tests by adding an `export` in `carp.sh`. Sorry for the inconvenience!

eriksvedang:
@TimDeve @hellerve Great teamwork!

hellerve:
It was a lot of fun, thanks @TimDeve!

PR #1045, opened 2020-12-02
refactor: Apply Ormolu auto-formatting

eriksvedang:
So... after working (partly) in Go for half a year I've kind of come around on the auto-formatting issue (sorry for any heated responses back in the summer) and I think that we should just apply [Ormolu](https://github.com/tweag/ormolu) to the code base. I still think the particular style looks kinda silly in places, but overall it will be good. Also, it can't be configured, so we don't need to obsess about that :)

Are you still positive to this change, @carp-lang/maintainers ?

I'd also like suggestions for how to integrate this with CI, or if we should just encourage contributors to add it to their "on save" hooks.


TimDeve:
As a Haskell beginner I can't weight in on if I like this specific formatter, but as I was saying to Veit the other day I almost don't want a work on codebases without auto-formatters anymore, so this seems like a "good thing".

hellerve (approved):
A huge :+1:  from me for this! I think style doesn’t matter so much, but uniformity does!

jacereda:
Yes, please!

scolsen:
No objections here!

TimDeve:
As an addendum, I've been looking into how to install `ormulu` and while it's available on brew, it doesn't seem available in your typical Linux distro repos, one needs to pull the repo and build. If it's something we're gonna mandate that seems to be a bit of hurdle for new-comers.

Any chance we could somehow have it be pulled and available via stack?

eriksvedang:
I installed it with `stack install ormolu` (macOS)

TimDeve:
Ah ok then we can recommend that in the contributor docs.

jacereda:
I'll also PR adding it to the dependencies for nix, so entering `nix-shell` will ensure it's available.

TimDeve:
This is the draft PR for the hook, I'm getting a lot of change when I run format so I'm not sure what's going on here:
https://github.com/carp-lang/Carp/pull/1047

PR #1047, opened 2020-12-02
chore: Creates pre-commit hook to format Haskell code

TimDeve:
This PR will create a `pre-commit` hook to format our Haskell code but I'm seeing a bunch of changes when I run `scripts/format.sh`, could people try it and see if it's local to my machine?


eriksvedang:
The format script works well, no local changes after running it.

I'm not sure about the hooks, how do I know if it works?

TimDeve:
> The format script works well, no local changes after running it.

Hmm, that's a problem, no idea why I have so many changes on mine. Which version of ormolu are you on?

> I'm not sure about the hooks, how do I know if it works?

Need to run `scripts/git-hooks/setup.sh` once then try to commit a Haskell file formatted weird and it should fix itself.

eriksvedang:
```bash
➜  Carp git:(master) ormolu -v    
ormolu 0.0.3.1 UNKNOWN UNKNOWN
using ghc-lib-parser 8.8.3.20200224
```

OK, I did try that but nothing happened. Will try again!

TimDeve:
Ah yeah there has been a major/pre-v1-minor jump since your version.
My version is `0.1.4.1`.

eriksvedang:
I'll try to install latest. Also kind of weird with a lot of changes, I thought the style would be more or less stable...

TimDeve:
It is pre-v1 and that version was even pre-v0.1 so changes are not super surprising, let's hope it's a bit more stable in the future.

eriksvedang:
> It is pre-v1 and that version was even pre-v0.1 so changes are not super surprising, let's hope it's a bit more stable in the future.

Right, that makes sense.

eriksvedang:
Now `master` is formatted using Ormolu `0.1.4.1`.

TimDeve:
Thanks, I've rebased.

@carp-lang/maintainers if people can check that the hook does format (add a few newlines to a Haskell file and then commit it). It won't stop you from doing an empty commit though but preventing that seems more works than it's worth.

eriksvedang:
Everything works for me now, very nice!

PR #1053, opened 2020-12-04
refactor: improve readability of interface functions

scolsen:
Also refactors the `define` function for readability.

The old definitions of these functions were quite unwieldy and difficult
to read. This refactor attempts to make the monadic contexts we're
juggling (often 3, Maybe, IO, Either) easier to spot.


eriksvedang:
This is great! Will read it properly in the weekend.

src/Primitives.hs
```diff
@@ -221,53 +217,63 @@ primitiveImplements x@(XObj _ _ _) ctx args =
...
-                            let fppl = projectFilePathPrintLength (contextProj ctx)
-                             in putStrLn (machineReadableInfoFromXObj fppl annXObj ++ " " ++ err)
-                          _ -> putStrLnWithColor Red err
-                        pure ctx
-                    Right ctx' -> pure (ctx' {contextGlobalEnv = envInsertAt globalEnv (getPath annXObj) (Binder adjustedMeta annXObj)})
-              _ -> pure (ctx {contextGlobalEnv = envInsertAt globalEnv (getPath annXObj) (Binder adjustedMeta annXObj)})
+  pure (hideIt freshBinder)
+    >>= \newBinder -> case annXObj of
+      XObj (Lst (XObj (Defalias _) _ _ : _)) _ _ ->
+        pure (ctx {contextTypeEnv = TypeEnv (envInsertAt (getTypeEnv typeEnv) (getPath annXObj) newBinder)})
```
eriksvedang, line 223: Since all these three matches do the same thing, maybe wrap it up in a function called `defineInTypeEnv` or something? `defineGlobal` could maybe change to `defineInValueEnv` or something to mirror it..?

src/Primitives.hs
```diff
@@ -221,53 +217,63 @@ primitiveImplements x@(XObj _ _ _) ctx args =
...
+      XObj (Lst (XObj (Deftype _) _ _ : _)) _ _ ->
+        pure (ctx {contextTypeEnv = TypeEnv (envInsertAt (getTypeEnv typeEnv) (getPath annXObj) newBinder)})
+      XObj (Lst (XObj (DefSumtype _) _ _ : _)) _ _ ->
+        pure (ctx {contextTypeEnv = TypeEnv (envInsertAt (getTypeEnv typeEnv) (getPath annXObj) newBinder)})
+      _ -> defineGlobal
+  where
+    freshBinder = (Binder emptyMeta annXObj)
+    defineGlobal :: IO Context
+    defineGlobal =
+      when (projectEchoC proj) (putStrLn (toC All (Binder emptyMeta annXObj)))
```
eriksvedang, line 233: I feel like this function is very imperative in it's style (with printing and all), so that `do` notation actually could make more sense. I'm open to leaving it as-is, but please consider it :)

src/Primitives.hs
```diff
@@ -221,53 +217,63 @@ primitiveImplements x@(XObj _ _ _) ctx args =
...
+        >> pure (lookupInEnv (getPath annXObj) globalEnv)
+        >>= \maybebinder ->
+          case maybebinder of
+            Nothing -> pure (hideIt freshBinder)
+              >>= \newBinder -> pure (ctx {contextGlobalEnv = envInsertAt globalEnv (getPath annXObj) newBinder})
```
eriksvedang, line 238: Once again it seems like this is a very common pattern, inserting a binder into an env. So a function for it seems useful.

src/Primitives.hs
```diff
@@ -221,53 +217,63 @@ primitiveImplements x@(XObj _ _ _) ctx args =
...
+        previousType = forceTy (binderXObj binder)
+        warn =
+          emitWarning
+            ( "Definition at " ++ prettyInfoFromXObj annXObj ++ " changed type of '" ++ show (getPath annXObj)
+                ++ "' from "
+                ++ show previousType
+                ++ " to "
+                ++ show (forceTy annXObj)
+            )
+    registerDef :: Binder -> IO Context
```
eriksvedang, line 261: The word "register" makes me think of the `register` for external functions/types, so maybe we can use something less misleading here?

src/Primitives.hs
```diff
@@ -437,9 +443,9 @@ primitiveMembers _ ctx [target] = do
               _
               ( XObj
                   ( Lst
-                      ( XObj (DefSumtype _) Nothing Nothing
-                          : XObj (Sym (SymPath _ _) Symbol) Nothing Nothing
-                          : sumtypeCases
+                      ( XObj (DefSumtype _) Nothing Nothing :
```
eriksvedang, line 446: What version of Ormolu are you on?

src/Primitives.hs
```diff
@@ -712,13 +718,18 @@ primitiveDeftype xobj ctx (name : rest) =
                     )
                in do
                     ctxWithDeps <- liftIO (foldM (define True) ctx' deps)
-                    let ctxWithInterfaceRegistrations =
+                    let fakeImplBinder sympath t = (Binder emptyMeta (XObj (Sym sympath Symbol) (Just dummyInfo) (Just t)))
```
eriksvedang, line 721: Neat!

src/Primitives.hs
```diff
@@ -151,65 +151,61 @@ primitiveColumn x@(XObj _ i t) ctx args =
...
+primitiveImplements _ ctx [x@(XObj (Sym interface@(SymPath _ _) _) _ _), (XObj (Sym (SymPath prefixes name) _) info _)] =
+  do
+    (maybeInterface, maybeImpl) <- pure ((lookupInEnv interface tyEnv), (lookupInEnv (SymPath modules name) global))
+    case (maybeInterface, maybeImpl) of
+      (_, Nothing) ->
+        if null modules
+          then pure (evalError ctx "Can't set the `implements` meta on a global definition before it is declared." info)
+          else updateMeta (Meta.stub (SymPath modules name)) ctx
+      (Nothing, Just (_, implBinder)) ->
+        (warn >> updateMeta implBinder ctx)
```
eriksvedang, line 163: This use of the monadic operators is really nice.

src/Primitives.hs
eriksvedang: Addendum: Noticed it being used above too. My loose convention has been to use "register" for external stuff and "define" for top level forms. So maybe these should be called something with define too.

src/Primitives.hs
```diff
@@ -151,65 +151,61 @@ primitiveColumn x@(XObj _ i t) ctx args =
...
-      checkInterface
-        >> either registerError updateImpls (registerInInterface ctx defobj interface)
+    warn =
+      emitWarning
+        ( "The interface "
+            ++ show (getPath x)
+            ++ " is not defined."
+            ++ " Did you define it using `definterface`?"
+        )
+    registerIt :: Binder -> Binder -> IO (Context, Either EvalError XObj)
```
eriksvedang, line 177: defineIt? See note below.

eriksvedang (approved):
Really great refactorings! I'm happy with it as is, but added some notes for possible improvements. I'll let you address what you see fit, let me know when I should hit that button.

src/Primitives.hs
eriksvedang: (I think they really shine when the whole flow fits on one line...)

src/Primitives.hs
eriksvedang: (below in the code, above in the list of comments... :S)

src/Interfaces.hs
```diff
@@ -48,58 +47,53 @@ instance Show InterfaceError where
 
 -- TODO: This is currently called once outside of this module--try to remove that call and make this internal.
 -- Checks whether a given form's type matches an interface, and if so, registers the form with the interface.
-registerInInterfaceIfNeeded :: Context -> SymPath -> SymPath -> Ty -> Either String Context
-registerInInterfaceIfNeeded ctx path@(SymPath _ _) interface@(SymPath [] name) definitionSignature =
-  maybe (pure ctx) (typeCheck . snd) (lookupInEnv interface typeEnv)
+registerInInterfaceIfNeeded :: Context -> Binder -> Binder -> Ty -> Either String Context
```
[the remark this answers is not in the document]
scolsen: Makes sense to me! Let's tackle that in follow-up PRs?

src/Obj.hs
```diff
@@ -299,6 +299,9 @@ getPath (XObj (Lst (XObj (Primitive _) _ _ : XObj (Sym path _) _ _ : _)) _ _) =
 getPath (XObj (Sym path _) _ _) = path
 getPath x = SymPath [] (pretty x)
 
+getBinderPath :: Binder -> SymPath
+getBinderPath = getPath . binderXObj
```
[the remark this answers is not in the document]
scolsen: I think point free is great for cases like this that are straightforward compositions of existing functions of one argument--in this case I find them really clear and pretty easy to follow. I think for any cases that are more complicated though, point free probably makes things more difficult to understand.

src/Primitives.hs
scolsen: Good to keep in mind, thanks!

src/Primitives.hs
scolsen: Sounds good to me!

src/Interfaces.hs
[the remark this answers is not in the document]
eriksvedang: Yes for sure, that was more a note to self :)

src/Primitives.hs
scolsen: Sure thing!

src/Primitives.hs
scolsen: Totally! I'll add a `Context.hs` and we can build on that

src/Primitives.hs
scolsen: sounds good--it refers to adding the xobj to an interface, which historically was also called registerInInterface...--let's rename it `implementInterface`?

src/Primitives.hs
scolsen: Changed to `addToInterface` I think this is more explicit and accurate, wdyt?

src/Primitives.hs
scolsen: Looks like I've got  0.0.3.1

scolsen: Just updated to 1.4.1

src/Interfaces.hs
```diff
@@ -48,58 +47,53 @@ instance Show InterfaceError where
...
 
--- | Given an XObj and an interface path, ensure that the form is
+-- | Given a binder and an interface path, ensure that the form is
 -- registered with the interface.
-registerInInterface :: Context -> XObj -> SymPath -> Either String Context
-registerInInterface ctx xobj interface =
-  case xobj of
-    XObj (Lst [XObj (Defn _) _ _, XObj (Sym path _) _ _, _, _]) _ (Just t) ->
+registerInInterface :: Context -> Binder -> Binder -> Either String Context
+registerInInterface ctx implementation interface =
```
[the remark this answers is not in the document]
scolsen: Thanks!

scolsen:
@eriksvedang made a couple of small changes based on your feedback--want to take another look?

eriksvedang:
Soooo nice.

PR #1054, opened 2020-12-05
refactor: Make Lookup module more focused and DRY

eriksvedang:


eriksvedang:
@scolsen I started moving things around to make the Lookup module less of a kitchen sink. There's a fair bit of duplication in there right now (and an old comment about that). My current idea is to try to use your more elegant way of looking up things but extend it to also return an (Env, Binder) tuple like the old code. Then we can provide the simpler mean of lookup with just combining the bigger function with `snd . lookupFn`. Sounds reasonable?

eriksvedang:
@scolsen No tests yet, but the Lookup module now has a lot less duplication, and more appropriate names for a bunch if things. Let me know if you can spot any more opportunities for improvement, or if I did something stupid!

scolsen:
@eriksvedang I'll try to take an extensive look tonight, excited about the cleanup!

src/Env.hs
```diff
@@ -0,0 +1,93 @@
...
+      let newInnerEnv = Binder emptyMeta (XObj (Mod (envReplaceEnvAt innerEnv ps replacement)) i t)
+       in env {envBindings = Map.insert p newInnerEnv (envBindings env)}
+    Just _ -> error ("Can't replace non-module: " ++ p)
+    Nothing -> error ("Can't replace non-existing module: " ++ p)
+
+-- | Add a Binder to a specific environment.
+envAddBinding :: Env -> String -> Binder -> Env
+envAddBinding env name binder = env {envBindings = Map.insert name binder (envBindings env)}
+
+{-# ANN addListOfBindings "HLint: ignore Eta reduce" #-}
```
jacereda, line 38: Why? `addListOfBindings` would be quite readable applying Eta reduction...

jacereda (approved):
More readable now definitely!

src/Env.hs
```diff
@@ -0,0 +1,93 @@
...
+  case Map.lookup name (envBindings env) of
+    Just _ ->
+      envAddBinding env name binder
+    Nothing ->
+      case envParent env of
+        Just parent -> env {envParent = Just (envReplaceBinding s binder parent)}
+        Nothing -> env
+envReplaceBinding (SymPath _ _) _ _ = error "TODO: cannot replace qualified bindings"
+
+envBindingNames :: Env -> [String]
```
scolsen, line 75: Nice :)

src/Interfaces.hs
```diff
@@ -95,5 +96,5 @@ retroactivelyRegisterInInterface ctx interface =
   either (\e -> error e) id resultCtx
   where
     env = contextGlobalEnv ctx
-    impls = recursiveLookupAll (getPath (binderXObj interface)) lookupImplementations env
+    impls = lookupMany Everywhere lookupImplementations (getPath (binderXObj interface)) env
```
scolsen, line 99: Love the fact that the more specific arguments (binder, env) comes later--this makes it possible to use partial application effectively in higher order functions.

src/Lookup.hs
```diff
@@ -30,90 +28,56 @@ lookupInEnv path@(SymPath (p : ps) name) env =
...
-          leaves = importsLookup env
-          above = case envParent env of
-            Just parent -> recursiveLookup parent
-            Nothing -> []
-       in --(trace $ "multiLookupInternal '" ++ name ++ "' " ++ show (envModuleName env) ++ ", spine: " ++ show (fmap snd spine) ++ ", leaves: " ++ show (fmap snd leaves) ++ ", above: " ++ show (fmap snd above))
-          spine ++ leaves ++ above
+-- | Like 'lookupBinder' but return the Meta for the binder, or a default empty meta.
+lookupMeta :: SymPath -> Env -> MetaData
+lookupMeta path globalEnv =
+  case lookupBinder path globalEnv of
```
scolsen, line 38: Could also write this as:

```haskell
maybe emptyMeta Meta.fromBinder . lookupBinder
```

Or with the arguments:

```haskell
maybe emptyMeta Meta.fromBinder (lookupBinder path globalEnv)
```

src/Lookup.hs
```diff
@@ -30,90 +28,56 @@ lookupInEnv path@(SymPath (p : ps) name) env =
...
       above = case envParent env of
-        Just parent -> recursiveLookupAll input lookf parent
+        Just parent -> lookupMany lookWhere lookf input parent
         Nothing -> []
    in spine ++ leaves ++ above
 
--- | Lookup binders by name.
-lookupByName :: String -> Env -> [Binder]
+-- | Lookup binders by name in a single Env (no recursion),
+lookupByName :: String -> Env -> [(Env, Binder)]
```
scolsen, line 77: If this only performs a lookup in a single env without recursing, does it need to return the `env`? After all, `env` in this case is precisely the `Env` argument passed to the function, so it doesn't provide any new information to callers.

src/Managed.hs
```diff
@@ -0,0 +1,35 @@
...
+-- | Find out if a type is "external", meaning it is not defined by the user
+--   in this program but instead imported from another C library or similar.
+-- NOTE: Quite possibly this function should be removed and we should rely on 'isManaged' instead?
+isExternalType :: TypeEnv -> Ty -> Bool
+isExternalType typeEnv (PointerTy p) =
+  isExternalType typeEnv p
+isExternalType typeEnv (StructTy (ConcreteNameTy name) _) =
+  case lookupBinder (SymPath [] name) (getTypeEnv typeEnv) of
+    Just (Binder _ (XObj (Lst (XObj (ExternalType _) _ _ : _)) _ _)) -> True
+    Just _ -> False
```
scolsen, line 16: These last two lines can be combined into:

```haskell
_ -> False
```

src/Managed.hs
```diff
@@ -0,0 +1,35 @@
...
+
+-- | Is this type managed - does it need to be freed?
+isManaged :: TypeEnv -> Ty -> Bool
+isManaged typeEnv (StructTy (ConcreteNameTy name) _) =
+  (name == "Array") || (name == "StaticArray") || (name == "Dictionary")
+    || ( case lookupBinder (SymPath [] name) (getTypeEnv typeEnv) of
+           Just (Binder _ (XObj (Lst (XObj (ExternalType _) _ _ : _)) _ _)) -> False
+           Just (Binder _ (XObj (Lst (XObj (Deftype _) _ _ : _)) _ _)) -> True
+           Just (Binder _ (XObj (Lst (XObj (DefSumtype _) _ _ : _)) _ _)) -> True
+           Just (Binder _ (XObj wrong _ _)) -> error ("Invalid XObj in type env: " ++ show wrong)
```
scolsen, line 29: Probably it should be caller's responsibility to handle these last two `error`ing cases? If so, I think the only check that's relevant is the `ExternalType` check, since we are probably ok assuming the `StructTy` passed to this is a valid type in the type env (if not, the error is on the caller). Then we could redefine the struct case:

```haskell
isManaged typeEnv (StructTy (ConcreteNameTy name) _) =
  (name == "Array") || (name == "StaticArray") || (name == "Dictionary")
  || ( case lookupBinder (SymPath [] name) (getTypeEnv typeEnv) of
         Just (Binder _ (XObj (Lst (XObj (ExternalType _) _ _ : _)) _ _)) -> False
         _ -> True -- all other structTys are managed
```

This would make it a pure function proper and push our errors upstream.

src/TypePredicates.hs
```diff
@@ -0,0 +1,30 @@
...
+isTypeGeneric :: Ty -> Bool
+isTypeGeneric (VarTy _) = True
+isTypeGeneric (FuncTy argTys retTy _) = any isTypeGeneric argTys || isTypeGeneric retTy
+isTypeGeneric (StructTy n tyArgs) = isTypeGeneric n || any isTypeGeneric tyArgs
+isTypeGeneric (PointerTy p) = isTypeGeneric p
+isTypeGeneric (RefTy r _) = isTypeGeneric r
+isTypeGeneric _ = False
+
+isFullyGenericType :: Ty -> Bool
+isFullyGenericType (VarTy _) = True
```
scolsen, line 14: We actually need to add a case here:

`(StructTy name members) = isFullyGenericType name && all isFullyGenericType members`

Since structs can now have variable names like `(f a)`

src/Env.hs
eriksvedang: Yeah we added these for all places where HLint wanted to Eta reduce. Might be worth reconsidering.

scolsen (approved):
Fantastic ✨ so glad lookups only return binders now!

I left a few comments but wouldn't consider any of them as necessary.

src/Managed.hs
scolsen: Or better yet:

```haskell
( case lookupBinder (SymPath [] name) (getTypeEnv typeEnv) of
         Just (Binder _ (XObj (Lst (XObj (ExternalType _) _ _ : _)) _ _)) -> False
         _ -> True -- all other structTys are managed
```

(since the last case covers the specific matches on names)

src/Lookup.hs
eriksvedang: The first one unfortunately didn't work, but I'll change to the second one!

src/Lookup.hs
eriksvedang: I think it has to, due to the generality of `lookupMany` (specifically, the signature of `lookupByName` makes it possible to construct the `multiLookup...` functions that all need to return the env where the binding was found)

src/Managed.hs
eriksvedang: Right!

src/Lookup.hs
scolsen: Ahhh, gotcha, thanks for pointing that out!

src/TypePredicates.hs
eriksvedang: Alright, adding it!

src/Managed.hs
eriksvedang: I'll rewrite this function soon anyways.

src/Lookup.hs
eriksvedang: I might be missing something also :)

PR #1055, opened 2020-12-07
fix: ensure type variables kinds are consistent

scolsen:
Previously, we didn't check that type variables had consistent kinds in
type definitions, which could lead to improper types. For example, one
could define:

    (deftype (Foo f a b) [x (f a) y f])

without issues even though this type is invalid since the variable `f`
is assigned two distinct kinds (nullary: * and unary: \*->\*).

This commit adds a check to ensure all of the instances of a type
variable in a type definition have the same kind. It also fixes an issue
whereby only higher-kinded types of one argument were allowed as type
members; now higher-kinded types of any arity are permitted (again
assuming variable kinds are consistent).

When a user writes a type that has variables with inconsistent kinds,
they will be hit with an error:

    (deftype (Foo (f a) b) [x (f a) y f])
    Invalid type definition for 'Foo':

    The type variable `f` is used inconsistently: (f a), f Type
    variables must be applied to the same number of arguments.

    Traceback:
    (deftype (Foo (f a) b) [x (f a) y f]) at REPL:1:1.


scolsen:
If we want, we can also adopt haskell's syntax for kinds. Here's what the same error looks like in haskell:

```haskell
Prelude> foo :: (f a) -> (f a b)

<interactive>:1:18: error:
    • Expected kind ‘k0 -> *’, but ‘f’ has kind ‘*’
    • In an expression type signature: (f a) -> (f a b)
      In the expression: foo :: (f a) -> (f a b)
      In an equation for ‘it’: it = foo :: (f a) -> (f a b)
```

Personally, I always found the `* -> *` notation for kinds a bit obscure, so as long as we think this simplification I've coded up here is clearer I think it's preferable. Consistency of use seemed accurate enough for what we're checking here, but if anyone else has ideas I'd be happy to hear them!

eriksvedang:
Awesome! I like the simplicity of this implementation. Needs rebase.

scolsen:
> Awesome! I like the simplicity of this implementation. Needs rebase.

Done!

hellerve (approved):

Issue #767, opened 2020-05-09
Twice wrapped match does not terminate


hellerve:
Writing `(match)` is not valid, and the compiler will tell you that it isn’t. Writing `((match))`, on the other hand, will not terminate—of course it is equally invalid, but the compiler shouldn’t choke on it.

Cheers

eriksvedang:
Seems like this is solved?!

hellerve:
It is, though the error message is a little weird. It seems like `eval` is getting confused there. I’ll prepare a better error message.

PR #1056, opened 2020-12-07
feat: better error messages for invalid objects

hellerve:
This PR adds better error messages for invalid AST objects by introducing `InvalidObjExample`, which is a type error that carries an example invocation with it. It also improves `InvalidObj` by calling `pretty` on the code inside it instead of `show`.

Cheers


hellerve:
This PR was opened in response to #767.

TimDeve:
Really nice.

eriksvedang:
Wow!

PR #1058, opened 2020-12-07
fix: Handle unit members correctly in array templates

### LINKED ISSUE #1048: incomplete type 'void' is not assignable
Hello!
I have the following code:
```clojure
(deftype Entry [min Int, max Int, char Char, password String])

(defmodule Entry
  (defn is-valid? [t]
    (let [count (String.count-char (Entry.password t) @(Entry.char t))]
      (= count (clamp @(Entry.min t) @(Entry.max t) count))))

  (defn of-string [s]
    (let
      [ar (String.split-by s &[\ ])
      range-string (String.split-by (Array.unsafe-nth &ar 0) &[\-])
      char-string (Array.unsafe-nth &ar 1)
      password @(Array.unsafe-nth &ar 2)
      range (Array.copy-map &(fn [a] (Maybe.unsafe-from (Int.from-string a))) &range-string)
      min @(Array.unsafe-nth &range 0)
      max @(Array.unsafe-nth &range 1)
      char @(Array.unsafe-nth &(String.chars char-string) 0)
      ]
      (IO.println &(Entry.prn &(Entry.init min max char password)))
      )))

(defn main []
  (let-do
    [lines (String.lines &(String.chomp &(IO.read-file "./input.txt")))
     entries (Array.copy-map &Entry.of-string &lines)
     ]
    (IO.println &(Array.prn &entries))
    ))

```

That generates the following error when run:

```
$ carp -x part-1.carp 
out/main.c:1668:49: error: 'void' must be the first and only parameter if specified
void Array_aset_MINUS_uninitialized_BANG___void (Array *aRef, int n, void newValue);
                                                ^
out/main.c:3529:49: error: 'void' must be the first and only parameter if specified
void Array_aset_MINUS_uninitialized_BANG___void (Array *aRef, int n, void newValue) {
                                                ^
out/main.c:3533:24: error: incomplete type 'void' is not assignable
    ((void*)a.data)[n] = newValue;
    ~~~~~~~~~~~~~~~~~~ ^
out/main.c:4004:66: error: too few arguments to function call, expected 3, have 2
                Array_aset_MINUS_uninitialized_BANG___void(_29, i);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       ^
out/main.c:3529:1: note: 'Array_aset_MINUS_uninitialized_BANG___void' declared here
void Array_aset_MINUS_uninitialized_BANG___void (Array *aRef, int n, void newValue) {
^
out/main.c:4692:21: error: argument type 'void' is incomplete
    temp = Unit_prn(((void*)a->data)[i]);
                    ^~~~~~~~~~~~~~~~~~~
out/main.c:4707:21: error: argument type 'void' is incomplete
    temp = Unit_prn(((void*)a->data)[i]);
                    ^~~~~~~~~~~~~~~~~~~
6 errors generated.
carp: callCommand: clang  -o out/Untitled -I /home/charles/build/Carp/core/  -fPIC -g -std=c99 -D_DEFAULT_SOURCE -Wall -Werror -Wno-unused-variable -Wno-self-assign -lm out/main.c (exit 1): failed
```

I'm running Carp off of master: 88cffce62699d9d985cfaacc4af06e7ebbc41b1d

I wasn't able to narrow down what was causing it... anyways thanks for taking a look!

scolsen:
Units may be used as members in arrays just like any other type,
however, the array templates (which are specific to arrays) hadn't been
updated to handle this. This commit makes the necessary updates and adds
some utility functions in efforts to make using templates in the
compiler easier.

After this commit, the following functions should work correctly on
(Array Unit):

- endo-map
- endo-filter
- copy
- aset!
- aset-uninitialized!
- aset
- pop-back!
- pop-back
- push-back
- push-back!

Just like other types with Unit members, the only value that is stored
in the resulting type is the terminal value of Unit `()` or unit. Any
side effects are run during evaluation of arguments and emitted prior to
any structure manipulations.

Because Unit is a terminal type (contains a single value) it makes
`endo-filter` an interesting case to consider. A filter on such an array
can only either retain all elements or drop all elements since there's
no way to predicate over the single terminal object Unit:

```
(endo-filter &(fn [_] false) [() ()])
=> []
(endo-filter &(fn [_] true) [() ()])
=> [() ()]
```

While arrays of units aren't very meaningful in terms of their contents,
their lengths will be expanded and contracted appropriately, so they can
be used for their length:

```
(defn iterate []
  (let-do [x [() () ()]]
    (while (> (Array.length &x) 0)
      (do (println* "foo")
          (Array.pop-back! &x)))))
(iterate)
foo
foo
foo
=> 0
```

Fixes #1048

### DIFF
--- a/src/ArrayTemplates.hs
+++ b/src/ArrayTemplates.hs
@@ -3,7 +3,6 @@
 module ArrayTemplates where
 
 import Concretize
-import Debug.Trace
 import Managed
 import Obj
 import Template
@@ -503,9 +502,7 @@ insideArrayInitLambda :: Ty -> String -> String
 insideArrayInitLambda (FuncTy _ UnitTy _) _ =
   "break;"
 insideArrayInitLambda t indexer =
-  trace
-    (show t)
-    "    Lambda lambda = "
+  "    Lambda lambda = "
     ++ initLambda
     ++ "\n"
     ++ "        (("
--- a/src/ArrayTemplates.hs
+++ b/src/ArrayTemplates.hs
@@ -3,6 +3,7 @@
 module ArrayTemplates where
 
 import Concretize
+import Debug.Trace
 import Managed
 import Obj
 import Template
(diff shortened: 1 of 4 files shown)


eriksvedang:
Wow, that was more involved to add than I imagined! Will do a proper review later.

scolsen:
> Wow, that was more involved to add than I imagined! Will do a proper review later.

Likewise! If we allow units in container types we need to special case for Unit in every individual template...which is kind of unfortunate to say the least. I'm not sure that there's any great way to make it generic either.

One option would be to ensure the explicit templates like the ones here also pass through the more general emissions code, that would remove some need for special casing. For instance, the declarations of functions in general omit void parameters--but for any custom template one has to perform the same omission manually (like we do here). I theory if the custom templates went through a second pass through the more general emitter logic, we wouldn't need the special cases for the declarations at least...but it seems like that would be complicated.

eriksvedang:
> If we allow units in container types we need to special case for Unit in every individual template..

Hmm, what container types do you refer to? (because I thought we handled Unit as a member of struct types now, right?

> One option would be to ensure the explicit templates like the ones here also pass through the more general emissions code [...]

I have been toying with a library for more principled emission where we could represent the code in a more C-agnostic way (which would make it possible to emit code for other languages as well). It's kind of a big task though, and I'm not sure my current implementation is good enough. Could be something to discuss in chat maybe.

eriksvedang:
Another thing to consider is reducing the number of templates. I think we could actually implement the ones that take ownership here (in the Array module) in terms of the ! ones. Not sure why we didn't before, might be a good reason I'm not remembering.

When it comes to sumtypes/structs we could maybe rely more on some deriving macros instead of all the explicit templates in Haskell? And also deftemplate. Not sure how that would work with Unit but maybe could help?

scolsen:
> Hmm, what container types do you refer to? (because I thought we handled Unit as a member of struct types now, right?

Sorry, yes we do. I was writing in a weird tense--basically just stating the fact that our current set up requires separate explicit unit checking for both structs and arrays.

>Another thing to consider is reducing the number of templates. I think we could actually implement the ones that take ownership here (in the Array module) in terms of the ! ones. Not sure why we didn't before, might be a good reason I'm not remembering.
>
>When it comes to sumtypes/structs we could maybe rely more on some deriving macros instead of all the explicit templates in Haskell? And also deftemplate. Not sure how that would work with Unit but maybe could help?

Worth a shot!

src/ArrayTemplates.hs
```diff
@@ -61,23 +75,42 @@ templateEFilter = defineTypeParameterizedTemplate templateCreator path t docs
...
-                      ]
+              multilineTemplate
+                ( let deleter = insideArrayDeletion typeEnv env insideTy
+                   in ( case insideTy of
+                          UnitTy ->
+                            [ "$DECL { ",
+                              "    int insertIndex = 0;",
+                              "    for(int i = 0; i < a.len; ++i) {",
+                              "        if(" ++ templateCodeForCallingLambda "(*predicate)" fTy [elt] ++ ") {",
+                              "            /* ignore () member; just increment length. */",
```
eriksvedang, line 86: This template really looks like it should be possible to write without duplicating the complete body, but rather extract the inner loop part (much like in endoMap above), no?

eriksvedang (approved):
Great job, and some nice refactorings in the templates!

src/ArrayTemplates.hs
scolsen: Good catch, updated!

scolsen: cb7800e

eriksvedang:
Ready to go?

eriksvedang (approved):

scolsen:
> Ready to go?

Yep, I think so! I ran tests locally and they passed -- CI is still running, but I don't expect it should break

PR #1059, opened 2020-12-07
Ormolu in default.nix and emacs before-save hook.

jacereda:


.dir-locals.el
```diff
@@ -0,0 +1 @@
+((haskell-mode . ((before-save-hook . (ormolu-format-buffer)))))
```
eriksvedang, line 1: This gives me a warning every time I visit a file in emacs.

eriksvedang: (I already have my own save hook setup for Ormolu)

jacereda: Right. How did you setup yours? It should be something local to the Carp directory (we don't want to format every codebase under the sun with ormolu).

eriksvedang: True, I did set it up globally (which is not great).
Maybe there's a way to avoid the warning?

jacereda: You can add it to the list of safe variables, but that's unsafe, somebody could add malicious code in a `.dir-locals.el` as a `before-save-hook`. There must be something better.

jacereda: Can you try now? 

Using `0` in .dir-locals should allow disabling it for a directory.

eriksvedang:
Worked well!

PR #1060, opened 2020-12-07
fix: Pointer str & prn

eriksvedang:


hellerve:
The `NULL` pointer actually prints a little differently, which is annoying and breaks the tests.

eriksvedang:
> The `NULL` pointer actually prints a little differently, which is annoying and breaks the tests.

Right, I was afraid of that... Oh well, the important thing is actually that it compiles (which it didn't before) so I'll adjust the test accordingly.

PR #1061, opened 2020-12-08
feat: 'delete' interface (deciding whether a type is managed or not)

eriksvedang:
Example program:

```clojure
(register-type Boo)
(register-type Doo)

(defmodule Boo
  (register init (Fn [] Boo))
  (register delete (Fn [Boo] ()))
  (implements delete Boo.delete))

(defmodule Doo
  (register init (Fn [] Doo)))

(defn f []
  (let [b (Boo.init)
        d (Doo.init)]
    ()))
```

Emitted C for function `f`:

```c
void f() {
    /* let */ {
        Boo _6 = Boo_init();
        Boo b = _6;
        Doo _9 = Doo_init();
        Doo d = _9;
        /* () */
        Boo_delete(b);
    }
}
```


src/Managed.hs
```diff
@@ -1,5 +1,7 @@
 module Managed where
 
+import Data.Maybe (mapMaybe)
+import Debug.Trace
```
hellerve, line 4: This import should probably be removed before merging, right?

src/StartingEnv.hs
```diff
@@ -421,6 +421,11 @@ startingTypeEnv =
...
       Map.fromList
         [ interfaceBinder
+            "delete"
+            (FuncTy [VarTy "a"] UnitTy StaticLifetimeTy)
+            ([SymPath ["Array"] "delete", SymPath ["Pointer"] "delete"] ++ registerFunctionFunctionsWithInterface "delete")
+            builtInSymbolInfo,
+          interfaceBinder
             "copy"
             (FuncTy [RefTy (VarTy "a") (VarTy "q")] (VarTy "a") StaticLifetimeTy)
             ([SymPath ["Array"] "copy", SymPath ["Pointer"] "copy"] ++ registerFunctionFunctionsWithInterface "copy")
```
hellerve, line 431: Maybe these could move into `arrayModule` and `pointerModule`?  This is not necessarily relevant to this PR, though.

hellerve (commented):
This looks great to me! We should probably add a few tests for this behavior and then maybe start writing docs (which can or cannot be part of this PR).

hellerve:
Heh, also the new implementation of `isManaged` confuses the old implementations for `findMemberFunctions*`. That makes sense.

eriksvedang:
NOTE: There's still a bunch of work to be done here, need to actually implement `delete` on all the existing managed types for example.

hellerve:
Right, I didn’t want to say that this is mergeable, just that the direction it's going in looks promising.

src/ArrayTemplates.hs
```diff
@@ -611,15 +611,15 @@ strTy typeEnv env (StructTy _ [innerType]) =
   ]
 strTy _ _ _ = []
 
-takeAddressOrNot :: TypeEnv -> Ty -> String
-takeAddressOrNot typeEnv t = if isManaged typeEnv t then "&" else ""
+takeAddressOrNot :: TypeEnv -> Env -> Ty -> String
```
eriksvedang, line 614: This is actually kind of wrong, should check if it's a primitive type instead?

src/Managed.hs
```diff
@@ -1,5 +1,6 @@
 module Managed where
 
+import Interfaces
 import Lookup
 import Obj
 import Types
```
eriksvedang, line 6: The function isExternalType here should be removed, we don't want to care about that anymore.

src/StructUtils.hs
```diff
@@ -5,15 +5,25 @@ import Obj
...
-memberInfo :: TypeEnv -> Ty -> (Ty, String, Ty)
-memberInfo typeEnv memberTy =
-  let refOrNotRefType = if isManaged typeEnv memberTy then RefTy memberTy (VarTy "w") else memberTy -- OBS! The VarTy "w" here is dubious
-   in (refOrNotRefType, if isManaged typeEnv memberTy then "&" else "", FuncTy [refOrNotRefType] StringTy StaticLifetimeTy)
+-- | The 'str'/'prn' functions for primitive types don't take refs, while other types do
+-- so we need to adjust for that when finding and calling them in compound types
+memberInfo :: TypeEnv -> Env -> Ty -> (Ty, String, Ty)
+memberInfo typeEnv globalEnv memberTy =
+  let refOrNotRefType =
+        if isManaged typeEnv globalEnv memberTy
```
eriksvedang, line 13: Incorrect, should use something like `isPrimitiveType`

test/init_global.carp
```diff
@@ -7,5 +7,6 @@
 ;; The one allocation left after 'carp_init_globals' should be 'g' itself:
 (defn main []
   (do
+    (Debug.log-memory-balance! true)
```
eriksvedang, line 10: Remove.

src/Managed.hs
```diff
@@ -18,10 +18,18 @@ isExternalType typeEnv (StructTy (ConcreteNameTy name) _) =
 isExternalType _ _ =
   False
 
--- | Is this type managed - does it need to be freed?
+-- | Should this type be handled by the memory management system.
+-- Implementation note: This top-level pattern match should be able to just
+-- match on all types and see whether they implement 'delete', but for some
+-- reson that doesn't work. Might need to handle generic types separately?
```
scolsen, line 24: interesting--seems like a good hunch

src/ArrayTemplates.hs
hellerve: Couldn't user-defined types also be primitive (like integral types etc)?

eriksvedang: Yes... we need a special interface for this probably.

eriksvedang: In this case it's actually just a matter of figuring out whether the str function takes a ref or not, so perhaps the code should just query for both and see if any fits.

src/StartingEnv.hs
```diff
@@ -290,6 +290,7 @@ dynamicModule =
         makeVarPrim "line" "returns the line a symbol was defined on." "(line mysymbol)" primitiveLine,
         makeVarPrim "column" "returns the column a symbol was defined on." "(column mysymbol)" primitiveColumn,
         makePrim "info" 1 "prints all information associated with a symbol." "(info mysymbol)" primitiveInfo,
+        makePrim "managed?" 1 "" "" primitiveIsManaged,
```
scolsen, line 295: nice!

src/ArrayTemplates.hs
TimDeve: > Yes... we need a special interface for this probably.

This would be similar to the [Copy trait](https://doc.rust-lang.org/std/marker/trait.Copy.html), no?

eriksvedang: Yup!

eriksvedang: Except the reason things are not passed as ref to str/prn is because they are so small. Copy trait/interface could work on any big structure as long as it's memcopyable, I think.

TimDeve: This is where `derive` helps I guess, have primitives be `Copy` and if you want it on your structure you can `derive` it.

eriksvedang: @TimDeve Yeah, let's aim for that! If you want to write a small proposal at some point, that'd be great.

TimDeve:
With this PR is there any protection against re-implementing `delete` where it already exist? We said it might be unsafe to do so on existing managed type.

scolsen:
> With this PR is there any protection against re-implementing `delete` where it already exist? We said it might be unsafe to do so on existing managed type.

I forget whether or not it was merged yet but a while back I was messing around with making defining two instances of the same interface impl (types are the same) an error at definition time instead of runtime

TimDeve:
That should do it.

src/ArrayTemplates.hs
eriksvedang: Added a small issue for this.

eriksvedang:
Let me know what you think...

src/Sumtypes.hs
```diff
@@ -172,8 +172,8 @@ concreteStr typeEnv env insidePath concreteStructTy@(StructTy (ConcreteNameTy ty
         )
         ( \(FuncTy [RefTy (StructTy _ _) _] StringTy _) ->
             concatMap
-              (depsOfPolymorphicFunction typeEnv env [] "prn" . typesStrFunctionType typeEnv)
-              (filter (\t -> (not . isExternalType typeEnv) t && (not . isFullyGenericType) t) (concatMap caseTys cases))
+              (depsOfPolymorphicFunction typeEnv env [] "prn" . typesStrFunctionType typeEnv env)
+              (filter (not . isFullyGenericType) (concatMap caseTys cases))
```
scolsen, line 176: I think @jacereda added a function in Utils.hs for `filter (not f)`--called `remove`

scolsen (approved):
Awesome! Can't wait to have this merged!

src/StructUtils.hs
```diff
@@ -1,19 +1,33 @@
...
+    then ("", withValueSig)
+    else ("&", withRefSig)
+  where
+    withValueSig = FuncTy [memberTy] StringTy StaticLifetimeTy
+    withRefSig = FuncTy [RefTy memberTy (VarTy "w")] StringTy StaticLifetimeTy -- "w" here is dubious?
+    callWithValue =
+      -- If these interfaces are not implemented we assume ref signature is fine.
+      -- Blitable is required to not accicentally pass a value owned by the struct to
+      -- a non-ref str/prn function implemented on one of its memeber types.
+      let blitable = interfaceImplementedForTy typeEnv globalEnv "blit" (FuncTy [memberTy] memberTy StaticLifetimeTy)
```
scolsen, line 23: It seems somewhat unfortunate that we need to check this.

Is there any way for us to make this more general? As in, could we somehow define `str` differently so that the special case is no longer a special case?

src/StructUtils.hs
```diff
@@ -1,19 +1,33 @@
...
+-- so we need to adjust for that when finding and calling them in compound types.
+-- The returned tuple contains ("" || "&", `str function type`).
+memberStrCallingConvention :: String -> TypeEnv -> Env -> Ty -> (String, Ty)
+memberStrCallingConvention strOrPrn typeEnv globalEnv memberTy =
+  if callWithValue
+    then ("", withValueSig)
+    else ("&", withRefSig)
+  where
+    withValueSig = FuncTy [memberTy] StringTy StaticLifetimeTy
+    withRefSig = FuncTy [RefTy memberTy (VarTy "w")] StringTy StaticLifetimeTy -- "w" here is dubious?
```
scolsen, line 18: Yes, I think using `w` is potentially problematic. We likely need a function here that ensures we don't create a collision w/ the ty vars generated/rewritten by the other parts of the compiler?

src/Sumtypes.hs
eriksvedang: Fixed!

src/StructUtils.hs
eriksvedang: Yes, we could force `str` to only take refs. I actually think using an interface for properties like this is kinda neat, but if we can avoid it all together that's even better of course.

src/StructUtils.hs
eriksvedang: Yeah, usually there's an XObj around that has a piece of unique information we can use for this. 

I'll try to make this function take such an `Info` as an additional argument.

eriksvedang: Hmm, it was harder than expected. The current solution might not actually be a problem since all type varaibles are uniquified before being solved (at the call site). Still, since I might have missed some detail I'll leave the comment in here for now.

eriksvedang:
I'll merge this now. I'm definitely up for changing / removing the `blit` interface, we can continue that discussion in #1079 ..!

PR #1062, opened 2020-12-08
Address incomplete patterns

jacereda:


jacereda:
This is a first step, IMHO we should try to follow the 'make illegal states unrepresentable' path. Ideally we should get rid of most of these error paths over time by providing strong warrantees (by providing `NonEmpty` lists to commands, for example).

jacereda:
The horrible error messages are deliberate, so we have some motivation to get rid of them ;)

src/Deftype.hs
```diff
@@ -428,6 +434,7 @@ binderForStrOrPrn typeEnv env insidePath structTy@(StructTy (ConcreteNameTy type
             (concreteStr typeEnv env structTy (memberXObjsToPairs membersXObjs) strOrPrn)
             ("converts a `" ++ typeName ++ "` to a string.")
         )
+binderForStrOrPrn _ _ _ _ _ _ = error "binderforstrorprn"
```
eriksvedang, line 437: I want to see the face of a person who gets "binderforstrorprn" as an error message :D

eriksvedang (approved):
Really nice, I see it has already caught a couple of lurking bugs / mistakes!

eriksvedang:
> The horrible error messages are deliberate, so we have some motivation to get rid of them ;)

I like your thinking!

eriksvedang:
> This is a first step, IMHO we should try to follow the 'make illegal states unrepresentable' path. Ideally we should get rid of most of these error paths over time by providing strong warrantees (by providing `NonEmpty` lists to commands, for example).

Yes, please – sounds awesome!

PR #1063, opened 2020-12-10
Use different command types for each arity

jacereda:

Issue #902, opened 2020-07-01
Unify languages


jacereda:
I was wondering if it would be possible to replace the dynamic language with plain Carp. 

AFAICS we could have something like a compilation cache to reduce the amount of compilation for dynamic functions and an execution cache to reduce the amount of "shelling out". With that in place, I guess it could be speedy enough. What do you think?

jacereda:
This isn't a novel idea, `zig` does use a compilation cache and I'm not sure, but probably does use an execution cache as well.

eriksvedang:
Interesting idea. How would the (binary) functions modify the REPL environment?

jacereda:
Via primitives?

jacereda:
Let's take https://github.com/carp-lang/Carp/blob/134d9a5b02ce91c53ad61262d3758d85443239a4/core/Dynamic.carp#L56 as an example. 

`relative-to` could be instead a `defncomptime`. `current-file` and `add-c` could remain dynamic functions. 

The point here is we have lots of useful stuff in carp that we can't use from dynamic functions and have to end up having duplicate functionality in the dynamic world.

At a later stage, and if it proves useful, we could consider having a protocol to communicate with the `comptime` executables (maybe some sort of Interpreter monad) and get rid of dynamic functions completely.

jacereda:
And looking at the issue title, it wasn't very fortunate, since what I'm proposing now is adding another mode instead of removing one :)

hellerve:
I don‘t think I understand the proposition completely just yet: would all code then be essentially static, in the sense of a JIT-like environment rather than an interpreter (but leveraging static compilers rather than a JIT framework)?

jacereda:
I've been thinking a bit more about this. 

Let's start with the mechanism for evaluating static functions at the REPL.  Let's say we want to evaluate:

`(Int.+ 1 2)`

I would propose separating the function from its arguments. The function `Int.+` would then be compiled into an executable in $HOME/.cache/carp/fn/<hash>. The hash can be computed via `Data.Hashable` easily. 

Now, the generated executable would need to `read` the arguments passed via the command line, apply the function and `repr` the result (I think we really need a `repr`, looks like the current approach is to use `println*`).

If we now evaluate at the REPL:

`(Int.+ 10 20)`

It will find the previously compiled executable and call it with the new arguments.

With that mechanism in place, I think `Eval.hs` could use it in most places where we error out with `HasStaticCall`.

Does that make sense?

[...]

PR #1066, opened 2020-12-13
Add more variants for running external commands

jacereda:
At some point this should help with #902.

Just look at the last 2 commits, the ones before are from #1063.


jacereda:
Also, it gets rid of `spawnCommand`, which requires quotation and is generally a bad thing.

eriksvedang:
I'll review this properly after rebase.

jacereda:
Rebased.

src/StartingEnv.hs
```diff
@@ -279,6 +279,8 @@ dynamicModule =
     variadics =
       let f = addVariadicCommand . spath
        in [ f "array" commandArray "creates an array from a collection of elements." "(array 1 2 3) ; => [1 2 3]",
+            f "run-with-args" commandRunWithArgs "runs the built executable with arguments." "(run-with-args 1 2 3)",
+            f "run-exe-with-args" commandRunExeWithArgs "runs an executable with arguments." "(run-exe-with-args \"path-to-executable\" 1 2 3)",
```
hellerve, line 283: For consistency, should there also be a `run-exe`? But I guess that would just be `run-exe-with-args` without args, so that we could define

```clojure
(defndynamic run-exe [exe]
  (run-exe-with-args exe))
```

jacereda: Maybe even better would be to get rid of all the variants except `run-exe-with-args` and implement the rest in the carp side?

hellerve: That makes sense, since they’re all strictly less powerful. I guess the only thing we need to do for that is to find the compiler that should be used by default, then we’re good to go.

jacereda: Do you mean finding the resulting executable path?

hellerve: Yeah, exactly.

PR #1067, opened 2020-12-13
Allow for repository files to be loaded selectively in load

hellerve:
This PR extends `load` by allowing a second argument to selectively load a specific file from a repository, instead of requiring `main.carp` or `<repo-name>.carp`.

It probably conflicts with #1063, and I’d happily amend my PR once it is merged.

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -809,20 +809,36 @@ primitiveDefmodule _ ctx [] =
 -- | "NORMAL" COMMANDS (just like the ones in Command.hs, but these need access to 'eval', etc.)
 
 -- | Command for loading a Carp file.
-commandLoad :: UnaryCommandCallback
-commandLoad ctx xobj@(XObj (Str path) i _) =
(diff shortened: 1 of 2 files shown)


jacereda (approved):

jacereda:
I've noticed you used the same `load` instead of a separate `load-file` or something like that.

I took the other approach in #1066. Is there any preference?

eriksvedang:
As was foreseen, has conflicts.

hellerve:
> I've noticed you used the same `load` instead of a separate `load-file` or something like that.
> 
> I took the other approach in #1066. Is there any preference?

You mean whether a command’s arity should be overloaded or a new command be introduced? I think for `load` it makes sense to keep it the same command, and generally this is a case-by-case thing, but I don’t have super strong preferences generally.

TimDeve:
To me when in Dynamic and you have a command that's like `CommandX + some config` it makes sense to overload it.

PR #1068, opened 2020-12-13
fix: handle unhandled case in isCaptured

scolsen:
Previously, this function missed a case, namely, one where the head of a
type definition contains a concrete struct, like `Maybe` with a variable
argument, `a`. That is,

```
(deftype (Foo (Maybe a)) [x (Maybe a)])
```

Would throw a type error. This commit fixes that issue.

### DIFF
--- a/src/Validate.hs
+++ b/src/Validate.hs
@@ -138,7 +138,8 @@ canBeUsedAsMemberType typeEnv typeVariables ty xobj =
         -- `a` may be used as a member, sans `f`, but `f` may not appear
         -- without `a`.
         isCaptured :: Ty -> Ty -> Bool
-        isCaptured t v@(VarTy _) = t == v
+        isCaptured v@(VarTy _) t@(VarTy _) = t == v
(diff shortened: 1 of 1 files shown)


scolsen:
Before we merge this, it's not clear to me that we should even allow type heads such as `(Foo (Maybe a))`--as it doesn't seem to make sense?

In such a case, isn't it likely the author meant `(Foo a) [x (Maybe a)]` instead?

TimDeve:
Yeah, I don't see what `(deftype (Foo (Maybe a)) [x (Maybe a)])` would mean.

scolsen:
> Yeah, I don't see what `(deftype (Foo (Maybe a)) [x (Maybe a)])` would mean. Should this just output a clearer error message instead?

sounds good! I think we should keep this case for exhaustiveness reasons, but let me add an additional check for type heads that ensures all their types are variables.

TimDeve:
@scolsen That was merged too early no?

scolsen:
> @scolsen That was merged too early, no?

Yeah, it's ok though, I can push the other changes as a separate PR

PR #1069, opened 2020-12-13
Add Dynamic.hash

jacereda:
Builds on top of #1066.

This can be useful if at some point we want to go in the direction mentioned in #902.


eriksvedang:
Nice!

eriksvedang:
Got some conflicts after mergeing other PR:s.

jacereda:
Rebased.

eriksvedang (commented):
It looks good, but I would appreciate a better explanation for what this PR does before merge:ing.

More specifically:

What's the reason for the "reimplementation" of `Map` and `Set`?
Why is deriving `Generic` used in some places?
Why do types that never occur as values in the code (like `ExecutionMode`) need to change – my understanding is that we only want to hash "Carp values", no?

jacereda:
For `Map` and `Set` to be `Hashable` they need to provide an instance, but those types are defined elsewhere and we can't write the instance there. The wrappers are used to avoid orphan instances https://wiki.haskell.org/Orphan_instance

`Generic` is used to derive automatically the Hashable instances. Generic is very useful, it can provide automatic instances for other things like `Functor`.

We only want to hash Carp stuff and I hashed everything "on demand". The reason `ExecutionMode` ended up being hashed is because someone (maybe `ClosureContext`?) ended up requiring it. The solution would be to identify all the "wrong" instances and write the offending instances manually, so they obviate the unneeded values (in this case, probably `Context` should be rewritten manually).

If you help me spot those unneeded instances I can rewrite those.

eriksvedang:
@jacereda OK, thanks a lot for the explanations! I'll see if I can identify those...

jacereda:
There's also a good reason to have our own `Map` and `Set`. We can experiment easily with `Data.Map.Strict` or `Data.HashMap(.Strict)`.

eriksvedang:
> There's also a good reason to have our own `Map` and `Set`. We can experiment easily with `Data.Map.Strict` or `Data.HashMap(.Strict)`.

OK, yeah that sounds good!

src/Info.hs
```diff
@@ -61,6 +62,9 @@ data FilePathPrintLength
   | ShortPath
   deriving (Eq)
 
+instance Hashable FilePathPrintLength where
```
eriksvedang, line 65: Not needed?

src/Obj.hs
```diff
@@ -885,7 +930,9 @@ forceTy :: XObj -> Ty
 forceTy xobj = fromMaybe (error ("No type in " ++ show xobj)) (xobjTy xobj)
 
 -- | How should the compiler be run? Interactively or just build / build & run and then quit?
-data ExecutionMode = Repl | Build | BuildAndRun | Install String | Check deriving (Show, Eq)
+data ExecutionMode = Repl | Build | BuildAndRun | Install String | Check deriving (Show, Eq, Generic)
+
+instance Hashable ExecutionMode
```
eriksvedang, line 935: This one?

src/Project.hs
```diff
@@ -1,11 +1,16 @@
...
 
+import Data.Hashable
+import GHC.Generics (Generic)
 import Info
 import Util
 
-data Target = Native | Target String
+data Target = Native | Target String deriving (Generic)
+
+instance Hashable Target
```
eriksvedang, line 13: This one?

src/Project.hs
```diff
@@ -45,6 +50,9 @@ data Project = Project
     projectCModules :: [FilePath],
     projectLoadStack :: [FilePath]
   }
+  deriving (Generic)
+
+instance Hashable Project
```
eriksvedang, line 55: This one?

src/Project.hs
```diff
@@ -92,14 +100,18 @@ instance Show Project where
...
 
 instance Show Includer where
   show (SystemInclude file) = "<" ++ file ++ ">"
   show (RelativeInclude file) = "\"" ++ file ++ "\""
 
 -- | This flag is used on Carp source files to decide wether to reload them or not when calling `(reload)` / `:r`
-data ReloadMode = DoesReload | Frozen deriving (Show)
+data ReloadMode = DoesReload | Frozen deriving (Show, Generic)
+
+instance Hashable ReloadMode
```
eriksvedang, line 114: This one?

eriksvedang (commented):
Found some places where it might be possible to remove the `Hashable` instance. I might have forgotten about some detail that makes it impossible of course...

jacereda:
Would it make sense to go farther and given:

```
(defn double [x] (* x x))
(defn times2 [a] (* a a))
```

ensure both hashes are the same?

eriksvedang:
> Would it make sense to go farther and given:
> 
> ```
> (defn double [x] (* x x))
> (defn times2 [a] (* a a))
> ```
> 
> ensure both hashes are the same?

I don't think so, seems like we'd get into some pretty serious (open?) computer science problems if we try to...

jacereda:
Well, I don't pretend to enter the realm of function equality, something like `(defn t2 [a] (* 2 a))` would yield a different hash. Just "structural equality" or whatever the term is, meaning that both would generate the same code modulo variable renaming.

eriksvedang:
> Well, I don't pretend to enter the realm of function equality, something like `(defn t2 [a] (* 2 a))` would yield a different hash. Just "structural equality" or whatever the term is, meaning that both would generate the same code modulo variable renaming.

Alright, yeah I guess if you rename all variables according to some predictable scheme it would work. Not sure it's worth it though :)

hellerve:
I think this is what Unison does? Maybe we could look at how they do it, but I think they just hash the entire AST, and variable names do not count, just their identities.

jacereda:
Damn, I entered a git shitstorm, let's see if I can fix it...

jacereda:
@hellerve Yes, something like that is what I had in mind. I guess I can experiment with that later.

eriksvedang (approved):

jacereda:
Oops, the hash type should be `Long`.

jacereda:
Changed to `Long`.

jacereda:
Maybe it would be better to use `Uint64`?

eriksvedang:
> Maybe it would be better to use `Uint64`?

No idea what the tradeoffs are...

jacereda:
Either there's a problem with evaluation or I'm missing something. 

I would expect:

```
(def foo 1)
(= (hash foo) (hash 1))
```

to yield true.

jacereda:
> > Maybe it would be better to use `Uint64`?
> 
> No idea what the tradeoffs are...

At least printing a `Uint64` wouldn't yield negative values.

jacereda:
> Either there's a problem with evaluation or I'm missing something.
> 
> I would expect:
> 
> ```
> (def foo 1)
> (= (hash foo) (hash 1))
> ```
> 
> to yield true.

I think the problem lies in `Eval.hs`, in `resolveDef`. If I add a case there for `Def` I get what I expect. Is that a bug?

jacereda:
> I think the problem lies in `Eval.hs`, in `resolveDef`. If I add a case there for `Def` I get what I expect. Is that a bug?

That's only the beginning. This:

```clojure
(def foo 1)
(def bar foo)
( = (hash foo) (hash bar))
```

still fails.

eriksvedang:
Oh no. Should we still merge this to avoid conflicts?

jacereda:
Sure, whatever is needed to fix the evaluation would probably be better in a separate PR.

eriksvedang:
> Also related, as convenient as it is, I don't like the fact that typing `Int.+` in the REPL ends up printing the documentation. I would expect the evaluation of that form. `ghci` and several other REPLs handle this via something like `:i Int.+` IIRC. Should I open an issue to discuss that?

Yeah, please do.

eriksvedang:
> I'm thinking about evaluation, not that we have a Generic instance for the AST, would a fold be a good way to handle evaluation?

Not exaclty sure how the generic instance plays into that, would this be like some kind of dynamic dispatch basically?

jacereda:
> Not exaclty sure how the generic instance plays into that, would this be like some kind of dynamic dispatch basically?

I'm just thinking that having a `Foldable` (that AFAIK can be implemented for free based on `Generic`) could help simplify `eval`.

eriksvedang:
> > Not exaclty sure how the generic instance plays into that, would this be like some kind of dynamic dispatch basically?
> 
> I'm just thinking that having a `Foldable` (that AFAIK can be implemented for free based on `Generic`) could help simplify `eval`.

For sure.

PR #1070, opened 2020-12-14
Primitive refactors

scolsen:
This commit is the first in what will hopefully be a series of helpful
primitive refactors. To start, we:

- Move some inline `evalError` strings into a `PrimitiveError` module,
  (similar to the `TypeError`/`Types` module relationship
- Add `Reifiable` instances for String and Int types to take these
  types to their XObj representation.
- Add info utility functions for converting Info data to an XObj
- Refactor the `info` primitive:

  - Use monadic combinators + `maybe` instead of nested cases.
  - Use helper lookup functions that take a *context*--nearly *all*
    lookup calls currently extract some env, typically without doing
    anything to it, to pass it to lookup. This is a sign the boundary is
    incorrect and lookups should take the context instead--this will allow
    us to eliminate a ton of local `globalEnv`, `typeEnv`, etc. bindings.
  - Don't print hidden bindings
  - Indent printed meta information.
  - Color bindings blue

## Some examples of info output changes

**Binders are now printed blue (hopefully easier to read on black/white bgs?)**
Screen Shot 2020-12-14 at 5 25 01 PM
**When signatures and privacy meta are present, they are printed.**
Screen Shot 2020-12-14 at 5 25 08 PM
**More binders**
Screen Shot 2020-12-14 at 5 25 16 PM

**Hidden binders are no longer printed**

Screen Shot 2020-12-14 at 5 25 59 PM

### DIFF
--- a/src/Lookup.hs
+++ b/src/Lookup.hs
@@ -2,10 +2,10 @@ module Lookup where
 
 import qualified Data.Map as Map
 import Data.Maybe (catMaybes, mapMaybe)
+import Env
 import qualified Meta
 import Obj
 import Types
-import Env
 
 -- | The type of generic lookup functions.
 type LookupFunc a b = a -> Env -> [b]
@@ -32,17 +32,17 @@ lookupInEnv path@(SymPath (p : ps) name) env =
 lookupInterfaceOrType :: Context -> SymPath -> Maybe Binder
 lookupInterfaceOrType ctx path =
   let typeEnv = (getTypeEnv (contextTypeEnv ctx))
-  in  lookupBinder path typeEnv
+   in lookupBinder path typeEnv
 
 lookupGlobalBinding :: Context -> SymPath -> Maybe Binder
 lookupGlobalBinding ctx path =
   let global = (contextGlobalEnv ctx)
-  in lookupBinder path global
+   in lookupBinder path global
 
 lookupContextualBinding :: Context -> SymPath -> Maybe Binder
 lookupContextualBinding ctx path =
   let ctxEnv = contextEnv ctx
-  in lookupBinder path ctxEnv
+   in lookupBinder path ctxEnv
 
 lookupEverywhere :: Context -> SymPath -> Maybe [Binder]
 lookupEverywhere ctx (SymPath _ name) =
--- a/src/Lookup.hs
+++ b/src/Lookup.hs
@@ -5,6 +5,7 @@ import Data.Maybe (catMaybes, mapMaybe)
 import qualified Meta
 import Obj
 import Types
+import Env
 
 -- | The type of generic lookup functions.
 type LookupFunc a b = a -> Env -> [b]
@@ -27,6 +28,27 @@ lookupInEnv path@(SymPath (p : ps) name) env =
       case envParent env of
         Just parent -> lookupInEnv path parent
         Nothing -> Nothing
+
+lookupInterfaceOrType :: Context -> SymPath -> Maybe Binder
+lookupInterfaceOrType ctx path =
+  let typeEnv = (getTypeEnv (contextTypeEnv ctx))
+  in  lookupBinder path typeEnv
+
+lookupGlobalBinding :: Context -> SymPath -> Maybe Binder
+lookupGlobalBinding ctx path =
+  let global = (contextGlobalEnv ctx)
+  in lookupBinder path global
+
+lookupContextualBinding :: Context -> SymPath -> Maybe Binder
+lookupContextualBinding ctx path =
+  let ctxEnv = contextEnv ctx
+  in lookupBinder path ctxEnv
+
+lookupEverywhere :: Context -> SymPath -> Maybe [Binder]
+lookupEverywhere ctx (SymPath _ name) =
+  case map snd (multiLookupEverywhere name (contextEnv ctx)) of
+    [] -> Nothing
+    xs -> Just xs
 
 -- | Like 'lookupInEnv' but only returns the Binder (no Env)
 lookupBinder :: SymPath -> Env -> Maybe Binder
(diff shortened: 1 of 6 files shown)


src/Lookup.hs
```diff
@@ -28,6 +29,27 @@ lookupInEnv path@(SymPath (p : ps) name) env =
         Just parent -> lookupInEnv path parent
         Nothing -> Nothing
 
+lookupInterfaceOrType :: Context -> SymPath -> Maybe Binder
```
scolsen, line 32: Generally speaking, I think we should move toward defining lookups against `Context` instead of `Env`.

There's tons of instances, currently, of 

```haskell
let global = contextGlobalEnv ctx
      typeE = contextTypeEnv ctx
```

etc., but we almost *never* (if ever?) manipulate the `env` before performing a lookup. Almost every function that performs a lookup just grabs an env as is to pass to a lookup function.

Instead, I think we should handle such env delving internally in `Lookup` and simply indicate what we're doing by well-named functions, the details of which remain internal:

`lookupGlobalBinding :: Context -> ...`
`lookupInternalBinding :: Context -> ...`
`lookupType :: Context -> ...`

This would enable us to remove all those noisy local bindings to envs. WDYT? @carp-lang/maintainers

scolsen: We can also perform further filtering internally. e.g. the function here, `lookupInterfaceOrType` could also have a specific version `lookupInterface` that peek into the XObj and only return it if it is in fact an interface.

```
lookupInterface :: Context -> Maybe Binder
lookupInterface ctx path = 
  case lookupInterfaceOrType ctx path of
      Just binder@(Binder _ <InterfaceXObj>) -> Just binder
      _ -> Nothing
```

Defining such functions in Lookup would eliminate tons of inline matches on Binder XObj contents in other functions as well.

scolsen: Basically, it'd make us be able to leverage `Maybe`, its monad, functor, applicative, etc. as an interface/context for our computations in more places and eliminate tons of `lets` and `case`s

jacereda (approved):
Very nice!

src/Lookup.hs
jacereda: I'd say yes, yes and yes.

eriksvedang: Sounds great!

src/Lookup.hs
```diff
@@ -28,6 +29,27 @@ lookupInEnv path@(SymPath (p : ps) name) env =
...
+lookupInterfaceOrType ctx path =
+  let typeEnv = (getTypeEnv (contextTypeEnv ctx))
+   in lookupBinder path typeEnv
+
+lookupGlobalBinding :: Context -> SymPath -> Maybe Binder
```
eriksvedang, line 37: Wouldn't it make sense to prefix these with something that makes them a bit different from the "old" lookup functions (or add something to the old ones, like a "InEnv" suffix or similarly) ? Right now it will be very hard to anticipate what type each lookup function expects when reading the code.

src/Lookup.hs
```diff
@@ -28,6 +29,27 @@ lookupInEnv path@(SymPath (p : ps) name) env =
...
+lookupGlobalBinding ctx path =
+  let global = (contextGlobalEnv ctx)
+   in lookupBinder path global
+
+lookupContextualBinding :: Context -> SymPath -> Maybe Binder
+lookupContextualBinding ctx path =
+  let ctxEnv = contextEnv ctx
+   in lookupBinder path ctxEnv
+
+lookupEverywhere :: Context -> SymPath -> Maybe [Binder]
```
eriksvedang, line 47: I think this one should still be called something with "multiLookup" since it returns many Binders.

src/PrimitiveError.hs
```diff
@@ -0,0 +1,59 @@
+module PrimitiveError where
```
eriksvedang, line 1: Nice module 👍

eriksvedang (changes requested):
Really good improvements. The insight regarding `Context` is great, will surely make things a lot nicer! I had some issues with the new names in `Lookup` though, I think those need to be very clear if we're going to have so many different ones.

eriksvedang:
One more comment: The non-printing of info for hidden bindings makes little sense to me – if you're trying to get info about something you don't care if it's hidden or not, right? 

One special case that I worry about particularly is when debugging concretized functions, since those are always hidden. This change will make that extra hard. But I also think that someone who is making an encapsulated module with some hidden helper functions might want to get their info.

When getting info for many things it's probably what you want though.

scolsen:
> One more comment: The non-printing of info for hidden bindings makes little sense to me – if you're trying to get info about something you don't care if it's hidden or not, right?
> 
> One special case that I worry about particularly is when debugging concretized functions, since those are always hidden. This change will make that extra hard. But I also think that someone who is making an encapsulated module with some hidden helper functions might want to get their info.
> 
> When getting info for many things it's probably what you want though.

We call `show Binder` to render a binder in info which already doesn't print hidden things--so you'd only see metadata for an invisible binder. 

But I think your reasoning here makes sense, so I've made a change to make it work w/ hidden binders too.

scolsen:
@eriksvedang made the changes you requested! Here's how hidden binders print now:

Screen Shot 2020-12-15 at 11 34 49 AM

(but only when calling `info`)

src/PrimitiveError.hs
scolsen: thanks!

src/Lookup.hs
scolsen: 4d30c24

src/Lookup.hs
scolsen: 4d30c24

eriksvedang (approved):
Awesome!

PR #1071, opened 2020-12-15
Add automatic map resizing

hellerve:
This PR adds automatic resizing to the `Map` and `Set` datatypes to deal with collisions under load. Basically, the number of buckets is recomputed when the load falls under/above a certain threshold, to avoid having too many collisions. This means that some inserts will take longer (because the map will be recomputed entirely), but hopefully this will amortized by avoiding unnecessary collisions.

This removes `Map.create-with-len` and `Set.create-with-len`.

Cheers


hellerve:
I would also like to refactor the docstrings in the `Map` and `Set` modules a little, but this PR is big enough as it is.

eriksvedang (changes requested):
Sorry for dragging my feet on this one, I'm still a little scared of this code :) Looks great though, have you compared benchmarks with before?

There seems to be a generated doc included in the changes, otherwise I think we're good to go!

hellerve:
> Sorry for dragging my feet on this one, I'm still a little scared of this code :) Looks great though, have you compared benchmarks with before?

I have, the benchmarks don’t show huge changes, thought the collision benchmarks do actually show some improvement.

> There seems to be a generated doc included in the changes, otherwise I think we're good to go!

I’ll fix that, thanks for the heads-up!

eriksvedang (approved):

Issue #875, opened 2020-06-13
Can't get env when submodules share names with types


scolsen:
```
carp: Can't get env.
CallStack (from HasCallStack):
  error, called at src/Lookup.hs:217:34 in CarpHask-0.3.0.0-JeTRwcTheISKNh9w0jc5KF:Lookup

```

**EDIT**: This seems to happen when a type shares a name with a submodule:

```
(deftype Foo Bar)

(defmodule Baz 
  (defmodule Foo
   ........
))
```

hellerve:
This is interesting. I can’t reproduce this bug with the following code:

```clojure
(deftype Foo Bar)

(defmodule Baz
  (defmodule Foo
    (defn x [] 1)))
```

But there is in fact a bug there, maybe different. If this is just put in the REPL, it seems to work at first. But then, if we try to `(build)`, we get:

```
carp: Can't insert into non-existing module: Foo
CallStack (from HasCallStack):
  error, called at src/Lookup.hs:192:16 in CarpHask-0.3.0.0-JeTRwcTheISKNh9w0jc5KF:Lookup
```

Which I presume is the same bug rearing a different, equally ugly head.

hellerve:
Actually, this looks like it’s related to #865 as well.

scolsen:
> This is interesting. I can’t reproduce this bug with the following code:
> 
> 
> 
> ```clojure
> 
> (deftype Foo Bar)
> 
> 
> 
> (defmodule Baz
> 
>   (defmodule Foo
> 
>     (defn x [] 1)))
> 
> ```
> 
> 
> 
> But there is in fact a bug there, maybe different. If this is just put in the REPL, it seems to work at first. But then, if we try to `(build)`, we get:
> 
> 
> 
> ```
> 
> carp: Can't insert into non-existing module: Foo
> 
> CallStack (from HasCallStack):
> 
>   error, called at src/Lookup.hs:192:16 in CarpHask-0.3.0.0-JeTRwcTheISKNh9w0jc5KF:Lookup
> 
> ```
> 
> 
> 
> Which I presume is the same bug rearing a different, equally ugly head.

oh wow! yeah I should have mentioned the bug in the op occurs when you `load` a source with the submodule setup

hellerve:
Oh, this is getting quite complicated. I think it boils down to modules that already exist in a parent environment being problematic.

eriksvedang:
I fail to reproduce this...

scolsen:
[...]

Issue #865, opened 2020-06-02
`Dynamic` cannot be used as a submodule name


scolsen:
Defining a submodule named `Dynamic` results in a compiler error.

Given the source:
```
(defmodule Q                                                                                                                                                                                             
  (defmodule Dynamic                                                                                                                                                                                     
    (defn a [] "a")                                                                                                                                                                                      
  )                                                                                                                                                                                                      
)  
```

The carp compiler throws an error when attempting to evaluate the Dynamic submodule:

```
鲤 (load "d.carp")
鲤 :i Q
carp: Can't insert into non-existing module: Dynamic
CallStack (from HasCallStack):
  error, called at src/Lookup.hs:190:16 in CarpHask-0.3.0.0-JeTRwcTheISKNh9w0jc5KF:Lookup
```

Renaming the submodule solves the issue:

```
鲤 :i Q
Q : Module = {
    D : Module = {
        a : (Fn [] (Ref String a))
    }
}
Defined at line 0, column 0 in 'dummy-file'

```

scolsen:
Note that it may be used as a top-level module name without issue.

eriksvedang:
Hmm, may be something with the magic lookup in the Dynamic module.

scolsen:
fixed in #1078

Issue #792, opened 2020-05-12
`hidden` in submodule confuses the compiler


TimDeve:
The following code:
```clojure
(defmodule Modone
 (defmodule Modtwo
  (hidden funone)
  (defn funone [] @"Hello")
  (defn funtwo [] (funone))))

(use Modone.Modtwo)

(defn main []
 (println* (the String (funtwo))))
```
Produces this error:
```
I encountered an error when emitting code:

I found an ambiguous symbol `funone` at line 5, column 20 in '/mnt/c/Users/tim/dev/sandbox/carpsudmod/main.carp'

Possibilities:
  Modone.Modtwo.funone
  Modone.funone

All possibilities have the correct type.

Traceback:
(build) at Compiler (Build & Run):1:5.
```

It only happens if the function is labeled with `hidden`.

Edit: Same with `private` and `sig` so probably related to `meta`

eriksvedang:
I think this is solved now?

TimDeve:
Not on latest master for me.

scolsen:
It's still an issue--it's proving a little tricky to solve. For some reason, the `MetaSet` primitive doesn't get the full stack context-- it only gets the top level module:

> Somehow pathstrings for submodules aren't captured in the pathstrings that primitiveMetaSet recieves
so when it's called before a form exists it only adds a DocStub to the top-level module

```
鲤 :i Sub
Sub : Module = {
    Mod : Module = {
        Ule : Module = {
            a-func : (Fn [a] a)
            hey : (Fn [(Ref String a)] String)
        }
    }
    a-func : a
}
```

> I tried altering primitiveDefModule...but it pretty muich breaks everything
what's odd is that the module itself/defining binding is correct
only MetaSet doesn't get the full story for some reason

> So, I think the issue underlying #823 is that we pop off a module name everytime we define a brand new module:

```
鲤 (load "tricky.carp")
pathstrings:["Sub","Mod","Ule"]
pathstrings:["Sub","Mod"]
pathstrings:["Sub"]
鲤
```

> On the face of it, this isn't problematic in eval because we dive down to eval the other expressions, but when it comes to meta-set of a sig it's receiving only the last path string, with the interior modules popped off, which assigns the sig to the wrong module.

For some reason, when metaset  forward declarations are evaluated it only gets `[Sub]` as its context paths, even though the paths should be `[Sub, Mod, Ule]` -- weirdly this issue only happens for metaSet

TimDeve:
[...]

PR #1078, opened 2020-12-17
Various submodule fixes

### LINKED ISSUE #823: Calling a function with a sig from within a submodule results in a compiler error
Calling a function with a sig from within a submodule results in a compiler error.
Example:
```clojure
(defmodule Sub
 (defmodule Mod
  (defmodule Ule
    (sig a-func (Fn [String] String))
    (defn a-func [a] a)

    (defn hey [a] (String.append a &(a-func @a))))))

(defn main []
  (println* (Sub.Mod.Ule.hey "Wow")))
```
Result in error:
```

I found an ambiguous symbol `a-func` at line 7, column 38 in '/Users/tim/example/main.carp'

Possibilities:
  Sub.Mod.Ule.a-func
  Sub.a-func

All possibilities have the correct type.

Traceback:
(build) at Compiler (Build & Run):1:5.
```

Example of working code for context:
```clojure
(defmodule Sub
    (sig a-func (Fn [String] String))
    (defn a-func [a] a)

    (defn hey [a] (String.append a &(a-func @a))))

(defn main []
  (println* (Sub.hey "Wow")))
```
```clojure
(defmodule Sub
 (defmodule Mod
  (defmodule Ule
    ;(sig a-func (Fn [String] String))
    (defn a-func [a] a)

    (defn hey [a] (String.append a &(a-func @a))))))

(defn main []
  (println* (Sub.Mod.Ule.hey "Wow")))
```
```clojure
(defmodule Sub
 (defmodule Mod
   (defmodule Ule
     (defn a-func [a] a)

     (sig hey (Fn [(Ref String)] String))
     (defn hey [a] (String.append a &(a-func @a))))))

(defn main []
 (println* (Sub.Mod.Ule.hey "Wow")))
```

scolsen:
This PR addresses a number of issues with submodule definition/resolution:

In particular it:

- Ensures calling a macro form that generates a binding within a submodule adds the bindings to the correct module (previously they were erroneously hoisted to whatever module was at the toplevel)
- Allows users to define submodules with names of global modules/types without causing a conflict.

fixes #823, #875, #865, #792

Notably, it *doesn't* handle the issue with `use`ing submodules reported in https://github.com/carp-lang/Carp/issues/792#issuecomment-729587870

This also gets us part of the way to #930 -- but we still need to prevent overwriting types in the type environment to get all the way there.


scolsen:
whew, this one was a huge pain to debug and finally understand what exactly was going on, but it feels good to have figured it out.

I think in general we need to be more explicit w/ lookups in certain contexts. The main problem here was that `lookupInEnv` *always* checks the parent environment, which is typically what we want, but in this case it wasn't since that's what led to the compiler confusing a submodule like `Foo.Dynamic` for the already defined global module `Dynamic`.

Likewise we should make an effort to break down the massive functions w/ tons of local definitions/cases into smaller independent functions. In Eval it was really hard to keep track of what context is being passed where since many of the branches in the main loop just use the original context passed at the top level of the function. If possible, we should try to separate the recursive part of the code from the actual transformations.

eriksvedang:
Wow, closing 4 bugs with one PR! Great job figuring all of this out. 

> If possible, we should try to separate the recursive part of the code from the actual transformations.

Yes, very much agree. Some kind of Context-monad and `traverse` like suggested elsewhere should help a lot, right? 

I've been thinking that the lookup functions should contain a set of already looked up paths to avoid recursion, what do you think about that?

scolsen:
> I've been thinking that the lookup functions should contain a set of already looked up paths to avoid recursion, what do you think about that?

That sounds good!

> Yes, very much agree. Some kind of Context-monad and `traverse` like suggested elsewhere should help a lot, right?

Yes, I think so. Tuples actually already have a monad instance when their left component is a Semigroup. I experimented with defining a semigroup over `Conrtext` so that we can use the monad instance of `(Context, Either EvalError XObj)` but this was kind of awkward. I think our best bet would be to define:

```
deftype EvaluationResult a = Result Context (Either EvalError a)

instance Functor (EvaluationResult a) where
  fmap f (Result ctx res) = Result ctx (fmap f res) 
  
instance Monad (EvalauationResult a) where
  r@(Result ctx res) >>= f = case res of Left _ -> r Right x -> f x
```

etc., where most instances just call the same instance on the interior Either. We can the define custom combinators for running these w/ functions that also want to manipulate ctx.

test/vector3.carp
```diff
@@ -14,7 +14,7 @@
   (assert-op test
              &(init 1.0 2.0 3.0) &(init 1.0 1.0 3.0)
              "/= operator works"
-             Vector3./=)
```
eriksvedang, line 17: Were these changes necessary now with the new env resolution, or is it just left over from testing something?

eriksvedang (approved):

eriksvedang:
Thanks for providing of the new type you have in mind, I've been pondering how that one should look too but I wasn't sure. If we were to move to a single mutable ref for storing the two root envs, would that change anything. Seems like more stuff would have to be in IO, no?

test/vector3.carp
scolsen: Yes, it is necessary now because of the small lookup resolution changes. I think it’s actually what we want too. Before, because of the way we handled the module lookups, this line found the globally defined `/=` ( from Generics.carp) even though the author here was really trying to reference a function defined in a particular module—I think we want errors in this case otherwise there will be situations in which the resulting behavior is unexpected.

scolsen:
> Thanks for providing an example of the new type you have in mind, I've been pondering how that one should look too, but I wasn't sure. If we were to move to a single mutable ref for storing the two root envs, would that change anything? Seems like more stuff would have to be in IO, no?

It depends exactly what type we use to contain the mutability--it's definitely going to be a monad, but we have some options as to whether we want to "simulate" mutability using a state monad or use IoRef for actual mutability

PR #1084, opened 2020-12-20
Support defining types in modules (BREAKING)

### LINKED ISSUE #930: Support defining types in modules

scolsen:
This commit adds support for defining types (using deftype) in modules.
Previously, all types were hoisted to the top level of the type
environment. After this commit, the type environment supports defining
nested modules just like the value env, so, calling the following:

```
(defmodule Foo (deftype Bar Baz))
```

Adds the following to the type env:

```
Foo : Module = {
    Bar : Type
}
```

and the following to the value env:

```
Foo : Module = {
    Bar : Module = {
        Baz : (Fn [] Foo.Bar)
        copy : (Fn [(Ref Foo.Bar q)] Foo.Bar)
        delete : (Fn [Foo.Bar] ())
        get-tag : (Fn [(Ref Foo.Bar q)] Int)
        prn : (Fn [(Ref Foo.Bar q)] String)
        str : (Fn [(Ref Foo.Bar q)] String)
    }
}

```

Such a type is *distinct* from any type defined at the top level that
happens to also have the name `Bar`.

This commit also updates info and tests to account for types in modules.

BREAKING CHANGE: This change is breaking since it alters the names of
types that were previously defined in modules. A good example of this is
the `Id` type in the `Color` module. Previously, one could refer to this
type by simply typing `Id` since it was hoisted to the top level. Now it
*must* be referred to by `Color.Id` since `Id` at the top level of the
type env and `Color.Id` (Id in the color module) are considered to be
distinct types.

fixes #930


scolsen:
NOTE: This might break SDL--the machine I ran tests on doesn't have a copy of SDL, so I couldn't check them. If any of the SDL types are defined in a module it will break.

src/Scoring.hs
```diff
@@ -17,36 +18,39 @@ scoreTypeBinder typeEnv b@(Binder _ (XObj (Lst (XObj x _ _ : XObj (Sym _ _) _ _
...
 
 depthOfDeftype :: TypeEnv -> Set.Set Ty -> XObj -> [Ty] -> Int
-depthOfDeftype typeEnv visited (XObj (Lst (_ : XObj (Sym (SymPath _ selfName) _) _ _ : rest)) _ _) varTys =
+depthOfDeftype typeEnv visited (XObj (Lst (_ : XObj (Sym (SymPath path selfName) _) _ _ : rest)) _ _) varTys =
   case concatMap expandCase rest of
     [] -> 100
     xs -> maximum xs
   where
-    depthsFromVarTys = map (depthOfType typeEnv visited selfName) varTys
+    depthsFromVarTys = map (depthOfType typeEnv visited (intercalate "" (path ++ [selfName]))) varTys
```
jacereda, line 46: `concat`?

src/Primitives.hs
```diff
@@ -303,9 +304,11 @@ primitiveInfo _ ctx [target@(XObj (Sym path@(SymPath _ _) _) _ _)] = do
...
-      case lookupBinderInContextEnv ctx path of
-        Nothing -> notFound ctx target path
-        Just found -> printer found
+      (printIfFound (lookupBinderInTypeEnv ctx path))
+        >> ( case lookupBinderInContextEnv ctx path of
+               Nothing -> notFound ctx target path
+               Just found -> printer found
+           )
   where
     printIfFound :: Maybe Binder -> IO (Context, Either EvalError XObj)
```
jacereda, line 313: Perhaps not related to this PR, but shouldn't this return `IO ()`?

src/Primitives.hs
```diff
@@ -676,7 +683,7 @@ primitiveDeftype _ _ _ = error "primitivedeftype"
 
 primitiveUse :: Primitive
 primitiveUse xobj ctx [XObj (Sym path _) _ _] =
-  pure $ maybe lookupInGlobal useModule (lookupInEnv path e)
+  (pure $ maybe lookupInGlobal useModule (lookupInEnv path e))
```
jacereda, line 686: The excess parenthesis will probably cause a warning in `HLint`

jacereda (commented):
Nice!

eriksvedang:
Woo, Christmas is coming early!

src/Types.hs
```diff
@@ -328,3 +334,19 @@ typesDeleterFunctionType memberType = FuncTy [memberType] UnitTy StaticLifetimeT
...
+
+createStructName :: [String] -> String -> String
+createStructName path name = (intercalate "." (path ++ [name]))
+
+getStructName :: Ty -> String
+getStructName (StructTy (ConcreteNameTy name) _) = name
+getStructName (StructTy (VarTy name) _) = name
+getStructName _ = ""
+
+getPathFromStructName :: String -> [String]
```
eriksvedang, line 346: Does this mean that the struct type is stored with it's enclosing environments as part of the name for the type?

(e.g. `(defmodule A (defmodule B (deftype T)))` would create a type with the name "A.B.T")

eriksvedang:
Got some conflicts now that I merged #1061 , I hope it's nothing too tricky. Let me know if you have any questions about the new code.

src/Types.hs
scolsen: Yep, that's precisely correct.

I think ideally we wouldn't need to do this, but right now there are some functions that take a struct name as an argument and *then* use the name to perform an additional lookup--(see Scoring.hs) in particular. Possibly these can be refactored and eventually the name of the type can drop the module prefixes and we can rely on lookups alone to disambiguate.

src/Scoring.hs
scolsen: thanks! bb777d4

src/Primitives.hs
scolsen: 4bbc6bc

src/Primitives.hs
scolsen: Definitely! Added a TODO about it: d963569

src/Primitives.hs
```diff
@@ -642,10 +646,14 @@ primitiveDeftype xobj ctx (name : rest) =
                       )
                       i
                       (Just TypeTy)
+                  holderEnv = \name' prev -> Env (Map.fromList []) (Just prev) (Just name') [] ExternalEnv 0
```
eriksvedang, line 649: What does "holder" mean here?

src/Types.hs
eriksvedang: Seems like `ConcreteNameTy` maybe should contain a `SymPath`..?

src/TypesToC.hs
```diff
@@ -27,28 +29,29 @@ tyToCManglePtr :: Bool -> Ty -> String
...
+    f ByteTy = "uint8_t"
+    f StringTy = "String"
+    f PatternTy = "Pattern"
+    f CharTy = "Char"
+    f UnitTy = "void"
+    f (VarTy x) = x
+    f (FuncTy argTys retTy _) = "Fn__" ++ joinWithUnderscore (map (tyToCManglePtr True) argTys) ++ "_" ++ tyToCManglePtr True retTy
+    f (StructTy s []) = tyToCManglePtr False s
+    f (StructTy s typeArgs) = tyToCManglePtr False s ++ "__" ++ joinWithUnderscore (map (tyToCManglePtr True) typeArgs)
+    f (ConcreteNameTy name) = mangle (intercalate "" (map unpack (splitOn (pack ".") (pack name))))
```
eriksvedang, line 47: `concat` here too?

src/Types.hs
scolsen: Yes, that seems reasonable!

PR #1085, opened 2020-12-20
chore: Simplify default.nix

jacereda:


jacereda:
This way it will call `cabal2nix`, so when adding a new module you don't need to modify `default.nix`.

eriksvedang:
I trust that this works and is better :)

Issue #1079, opened 2020-12-19
"Blit":able interface


eriksvedang:
Could be used to copy values using memcpy instead of walking the datastructure (e.g. an array of ints).

Similar to Rust:s `copy` trait (https://doc.rust-lang.org/std/marker/trait.Copy.html) ?

jacereda:
By blittable do you mean "combine chunks of memory applying logical operations"?

jacereda:
I see you used that term in https://github.com/carp-lang/Carp/commit/9e820df02a24e3700d178ecefc2d9ed894666f9b

Maybe something like `pod?` would be more appropriate?

jacereda:
as in Plain Old Data...

eriksvedang:
Yeah I was considering that too. The idea is that it's freely copyable (the deleter does nothing for the type) and "self contained", no pointers.

eriksvedang:
Does anyone have any more ideas for names? I guess mirroring Rust would make a lot of sense here, to avoid too much confusion.

scolsen:
> freely copyable

What does this mean precisely? That there's no need to copy individual members of the type? 

I left a comment on the delete PR, but I think the best case would be if we found some way to make this function not needed. Instead, `copy` could remain the sole interface, and implementers can decide how to do it, `memcpy` or whatever.

eriksvedang:
> What does this mean precisely? That there's no need to copy individual members of the type?

With "freely copyable" I mean anything that only lives on the stack and where copying or deleting is taken care of by the C value type semantics.

There's currently no way to know if a type has these properties inside Carp. We had some special casing for it but it depended on knowing which ones of the built-in types are special. With the changes in https://github.com/carp-lang/Carp/pull/1061 there's very little difference between built-in and external types.

Now, in regards to `str`/`prn` we could just decide that they always take a `Ref`, since then we ensure that they won't delete anything passed to them (like `StringCopy.str` does, for example).

> Instead, copy could remain the sole interface

I agree that this is desirable, but I'm not convinced it's enough. For example, `Array` is not "freely copyable" but it's implementation of `copy` could be made a lot more efficient for the situations where it contains a type that is (via `memcpy`). I can't see how a single `copy` interface can provide enough information to implement this properly though – let me know if I'm missing something.

scolsen:
[...]

PR #1087, opened 2020-12-21
Document delete

hellerve:
This PR documents the behavior introduced in #1061, except for `blit`, since we’re still discussing that in #1079.

Cheers


eriksvedang (approved):
Very nice!

PR #1088, opened 2020-12-21
feat: Add drop interface

### LINKED ISSUE #954: Support implementing `drop` | `delete` interface (custom automatic resource management)
#737 is related but I wanted to make a clearer feature request.

It could be quite useful an equivalent to [Rust's drop trait](https://doc.rust-lang.org/rust-by-example/trait/drop.html) that would allow to write custom logic to be run before (or instead) `delete` is called. 

There are two use cases I specifically have in mind:
- freeing non-memory related resources, like socket or file handles (avoiding work-arounds like [this](https://github.com/carpentry-org/sockets/blob/edbb459fb909c51be907abca753a3ae8ea80f242/sockets.carp#L116))
- freeing memory for opaque type by calling custom destructor as seen in #737 

Ideally it would be as simple as calling `(implements MyType.drop drop)` but I don't think there is anyway at the moment to enforce the fact that `MyType.drop` takes MyType as an argument.

hellerve:
his PR adds support for a `drop` interface, i.e. an interface that gets run before a value is deleted. It should not do any memory cleanup, but other cleanup, like closing OS handles, can be done here. It fixes #954.

Cheers

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -6,7 +6,7 @@ import AssignTypes
 import Constraints
 import Control.Monad.State
 import Data.List (foldl')
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, catMaybes)
 import Debug.Trace
 import Env
 import Info
@@ -1015,9 +1015,7 @@ manageMemory typeEnv globalEnv root =
                     _ <- unmanage value -- The assigned value can't be used anymore
                     MemState managed _ _ <- get
                     -- Delete the value previously stored in the variable, if it's still alive
-                    let deleters = case createDeleter okCorrectVariable of
-                          Just d -> Set.fromList [d]
-                          Nothing -> Set.empty
+                    let deleters = createDeleterAndDrop okCorrectVariable
                         newVariable =
                           case okMode of
                             Symbol -> error "How to handle this?"
@@ -1138,9 +1136,7 @@ manageMemory typeEnv globalEnv root =
                 deletedInBoth = Set.intersection deletedInTrue deletedInFalse
                 createdInTrue = memStateDeleters stillAliveTrue \\ preDeleters
                 createdInFalse = memStateDeleters stillAliveFalse \\ preDeleters
-                selfDeleter = case createDeleter xobj of
-                  Just ok -> Set.fromList [ok]
-                  Nothing -> Set.empty
+                selfDeleter = createDeleterAndDrop xobj
                 createdAndDeletedInTrue = createdInTrue \\ selfDeleter
                 createdAndDeletedInFalse = createdInFalse \\ selfDeleter
                 delsTrue = Set.union (deletedInFalse \\ deletedInBoth) createdAndDeletedInTrue
@@ -1353,6 +1349,7 @@ manageMemory typeEnv globalEnv root =
                               FakeDeleter {deleterVariable = dv} -> dv == deleterName
                               PrimDeleter {aliveVariable = dv} -> dv == deleterName
                               RefDeleter {refVariable = dv} -> dv == deleterName
+                              Drop {} -> False
                           )
                           deleters
                  in case matchingDeleters of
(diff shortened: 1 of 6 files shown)


hellerve:
This is a work in progress since variable movement currently generates too many drops. See for instance this generated C for the example I added in this PR:

```c
int main(int argc, char** argv) {
    carp_init_globals(argc, argv);
    /* let */ {
        A _6 = A_init();
        A a = _6;
        /* () */
        A_drop(&_6);
        A_drop(&a);
        A_delete(a);
    }
    return 0;
}
```

As you can see, `drop` is emitted twice, once for the autogenerated variable. This is of course not okay.

TimDeve:
There is some context on why implementing drop might conflict with the way we do `copy` in https://github.com/carp-lang/Carp/issues/954. To summarize if someone implement a custom `drop` they would also need to write a custom `copy` to ensure the resource is not dropped twice.

hellerve:
I don’t think it necessarily conflicts, but it’s true: someone who implements `drop` probably also wants to make sure that on a copy special code is run as well. I’d happily add that interface as well.

eriksvedang:
I think the current approach is a bit off. Each deleter in the set of deleters represent one value being alive in that scope, so adding a second deleter for any such value seems like it could be confusing. 

Instead I think it's the `ProperDeleter` (which is the sumtype case that will actually emit a call to `delete`, see Emit.hs:775) that should contain the name of the drop function to call. That way, when emitting the call to `delete`, it can also emit the call to `drop` right before. 

Should be a less intrusive approach to the already complicated `manageMemory` function also, where we ensure that only things that has deleters calls drop.

hellerve:
But shouldn’t `drop` also be available when we don’t have a proper deleter?

eriksvedang:
> But shouldn’t `drop` also be available when we don’t have a proper deleter?

That could probably be added to the other deleters then. Having more than one deleter per value is not going to fly, I'm afraid.

hellerve:
It does work, though?

hellerve:
Good news: I made `drop` part of existing deleters instead. It’s now part of `ProperDeleter`. This simplifies the implementation and should ease any concerns about it.

If this is satisfying, I’ll add more documentation on it.

hellerve:
It also removes the double drop problem.

TimDeve:
Putting here some of the thought I had originally posted on the Gitter:

A potential pitfall with `drop` is that if you `copy` a value you might potentially release your resource twice. It would be good to have a way to mark a type as non-copyable and provided a `clone` that ensure that the resource is also duplicated. It would also mean that anytype that contains that non-copyable type can't be copyable as well, not sure how doable that is to enforce...

scolsen:
> Putting here some of the thought I had originally posted on the Gitter:
> 
> A potential pitfall with `drop` is that if you `copy` a value you might potentially release your resource twice. It would be good to have a way to mark a type as non-copyable and provided a `clone` that ensure that the resource is also duplicated. It would also mean that anytype that contains that non-copyable type can't be copyable as well, not sure how doable that is to enforce...

This is actually I think the third time this use-case of "making sure a thing *doesn't* implement some interface (in this case copy)* has come up. I'm more and more convinced it'd be worthwhile to offer it as a general mechanism, just not sure what to call it.

eriksvedang:
@hellerve Nice, this is exactly what I had in mind.

@scolsen Should be pretty easy to implement such a feature, exactly what to call it might not be super important as long as it's internal? We could add it in a later PR also, since we can just try to avoid making that error in the meantime :)

TimDeve:
@scolsen Yes, the problem is related to `blit`. Rust solves this by only allowing `blit` on types comprised of other `blit` type, and everything else needs to handcraft `clone` I believe. And you can't implement `drop` on a `blit` type.

I used the Carp names here to avoid confusion as `blit` is called `copy` in Rust.

hellerve:
To continue with this, can we agree on necessary next steps for this to be a feature?

As a must do I have:
- [ ] Write documentation

As optionals I have:
- [ ] Change semantics and enforcing of `copy` and `blit` on  a type that has `drop`, and
- [ ] Adding a mechanism for enforcing that something doesn’t implement an interface.

Both optionals seem reasonable, but also like all of this together would make for a _massive_ PR.

scolsen:
> To continue with this, can we agree on necessary next steps for this to be a feature?
> 
> As a must do I have:
> 
> * [ ]  Write documentation
> 
> As optionals I have:
> 
> * [ ]  Change semantics and enforcing of `copy` and `blit` on  a type that has `drop`, and
> * [ ]  Adding a mechanism for enforcing that something doesn’t implement an interface.
> 
> Both optionals seem reasonable, but also like all of this together would make for a _massive_ PR.

Agreed. I think this PR should be scoped to adding drop only and imo we should add a general mechanism for enforcing  interface dependencies (in addition to this, we also chatted about it in #1095) in a separate PR.

hellerve:
I added (fairly short) documentation for `drop`, and I’ll convert this from a draft to a real PR. Let me know if anything is missing!

src/Concretize.hs
```diff
@@ -1388,7 +1388,7 @@ manageMemory typeEnv globalEnv root =
         else pure (Right xobj)
     unmanageArg xobj@XObj {} =
       pure (Right xobj)
-    createDeleter :: XObj -> Maybe Deleter
+    getDropFunc t = nameOfPolymorphicFunction typeEnv globalEnv (FuncTy [(RefTy t (VarTy "w"))] UnitTy StaticLifetimeTy) "drop" -- TODO: var is not sound
```
eriksvedang, line 1391: You can create a lifetime from the xobj here by using `makeTypeVariableNameFromInfo`, it will be ugly but unique.

eriksvedang (commented):
Everything looks great, but I had one idea for how to get rid of the hard coded life time.

eriksvedang (approved):

eriksvedang:
Oh, that made it fail..?

src/Concretize.hs
hellerve: Thanks for that hint, I didn’t know that! Fixed.

PR #1089, opened 2020-12-21
Make sure output directory is right in test

hellerve:
This PR uses `--eval-preload` to make sure the output directory in test is as expected when running the test script. This makes sure that having a different config set in `profile.carp` doesn’t conflict with running the tests.

Cheers

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -136,7 +136,7 @@ main = do
   _ <-
     ifCarpDirSet
       ( pure startingContext
-          >>= load [carpProfile | hasProfile]
+          >>= load [carpProfile | hasProfile && profile]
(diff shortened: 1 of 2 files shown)


jacereda:
I'd say there're several things that can cause impurities, perhaps tests should just disable loading `profile.carp`.

hellerve:
Mhm, that makes sense too. Do we have a mechanism for that?

jacereda:
`--no-profile`

jacereda:
Maybe tests could set `CARP_OPTS` (used in `carp.sh`).

hellerve:
Good call, I fixed all of that. It turns out `--no-profile` was also broken, which I was able to fix.

Issue #1076, opened 2020-12-16
RFC: Explicit commands for documentation


jacereda:
As convenient as it is, I don't like the fact that typing `Int.+` in the REPL ends up printing the documentation. I would expect that to yield the evaluation of that form. ghci and several other REPLs would handle this via something like `:i Int.+`. We could try to replicate that or have a primitive. What do you think?

scolsen:
We do have `:i` in fact, that's what you see when you type `Int.+` by itself. I'm fine with either behavior.

eriksvedang:
I'd prefer evaluation too, to be able to inspect the value of global variables (`def`).

hellerve:
Since this is just a special case, “fixing” this is as simple as removing a special case from `executeCommand` in `Eval`. This would yield:

```
> Int.+
=> (external Int.+ (λ [Int Int] Int))
> (defdynamic x 1)
> x
=> 1
> (def y 2)
> y
=> (def y 2)
```

I suppose we should then replace that with a special case for static values?

eriksvedang:
Shouldn't it just compile `def`:s and run those just like functions?

hellerve:
That would be a special case, since we don’t automatically build and run unless it’s a call expression. We can make it so that not just call expressions get run, but that would potentially lead to very weird behaviors, especially if it’s not just `def`s, but also functions like `Int.+` we include. Either way it wouldn’t be hard implementation-wise, but I’m not sure what the right approach would be.

eriksvedang:
We can print functions though (as "λ"). So they would evaluate just fine.

eriksvedang:
The idea would be to completely rely on `HasStaticCall` to figure out whether to build-and-run or not.

hellerve:
We already do that, but `HasStaticCall` is not emitted on symbol lookup in `eval`.

eriksvedang:
> We already do that, but `HasStaticCall` is not emitted on symbol lookup in `eval`.

I see. But maybe it could?

hellerve:
I’m experimenting with this approach as we speak, and will report back on my findings!

hellerve:
This approach seems to work, although it is a little clunky (I’m sure it can be made prettier, though). This is the output now:

```
> Int.+
Compiled to '/Users/veitheller/.carp/out/Untitled' (executable)
λ
=> 0
> +
=> (command Dynamic.+ [[x] [y]])
> (def x 1)
> x
[...]

PR #1090, opened 2020-12-21
feat: Evaluate symbols for statics in REPL

hellerve:
This PR changes the default behavior in the REPL from printing symbol info to evaluating that symbol. In order to do this, we needed to add a few special cases to the symbol lookup in `eval` for `def`s, `defn`s, and `external`s.

Please note that that makes a weird problem that we have even weirder. Consider this program:

```clojure
(def x 1)

x

(defn main [] x)
```

Currently, this will print the information of `x` during compilation. Now, this compiles another program and prints `1` before compiling the actual `main` program. This is a somewhat unrelated bug that should probably be fixed—global symbols should probably just be an error in a file—, but it is existing behavior that changes.

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -65,17 +65,24 @@ eval ctx xobj@(XObj o info ty) preference =
     Lst body -> eval' body
     Sym spath@(SymPath p n) _ ->
       pure $
-        fromMaybe
-          (evalError ctx ("Can't find symbol '" ++ show n ++ "'") info) -- all else failed, error.
-          -- Certain contexts prefer looking up bindings in the dynamic environment (e.g. defdyanmic) while others
-          -- prefer the static global environment.
-          ( ( case preference of
+        case tryAllLookups of
+          Just (_, (Right (XObj (Lst ((XObj Def _ _) : _)) _ _))) ->
+            (ctx, Left (HasStaticCall xobj info))
+          Just (_, (Right (XObj (Lst ((XObj (Defn _) _ _) : _)) _ _))) ->
+            (ctx, Left (HasStaticCall xobj info))
+          Just (_, (Right (XObj (Lst ((XObj (External _) _ _) : _)) _ _))) ->
+            (ctx, Left (HasStaticCall xobj info))
+          val ->
+            fromMaybe
+              (evalError ctx ("Can't find symbol '" ++ show n ++ "'") info) -- all else failed, error.
+              val
+      where
+        tryAllLookups =
+          ( case preference of
                 PreferDynamic -> tryDynamicLookup
                 PreferGlobal -> (tryLookup spath <|> tryDynamicLookup)
-            )
-              <|> (if null p then tryInternalLookup spath else tryLookup spath)
           )
-      where
+            <|> (if null p then tryInternalLookup spath else tryLookup spath)
         tryDynamicLookup =
           ( lookupBinder (SymPath ("Dynamic" : p) n) (contextGlobalEnv ctx)
               >>= \(Binder _ found) -> pure (ctx, Right (resolveDef found))
@@ -576,18 +583,6 @@ folder context xobj = do
 
 -- | Take a repl command and execute it.
 executeCommand :: Context -> XObj -> IO (XObj, Context)
-executeCommand ctx s@(XObj (Sym _ _) _ _) =
-  executeCommand
-    ctx
-    ( XObj
-        ( Lst
-            [ XObj (Sym (SymPath [] "info") Symbol) (Just dummyInfo) Nothing,
-              s
-            ]
-        )
-        (Just dummyInfo)
-        Nothing
-    )
 executeCommand ctx@(Context env _ _ _ _ _ _ _) xobj =
   do
     when (isJust (envModuleName env)) $
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -64,19 +64,18 @@ eval ctx xobj@(XObj o info ty) preference shouldResolve =
   case o of
     Lst body -> eval' body
     Sym spath@(SymPath p n) _ ->
-      if shouldResolve
-      then
-        pure $
-          case tryAllLookups of
-            Just (_, (Right (XObj (Lst ((XObj Def _ _) : _)) _ _))) ->
-              (ctx, Left (HasStaticCall xobj info))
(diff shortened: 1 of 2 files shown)


hellerve:
Oh, this is a possible fix for #1076, if we decide on it.

src/Eval.hs
```diff
@@ -65,17 +65,24 @@ eval ctx xobj@(XObj o info ty) preference =
...
-          -- Certain contexts prefer looking up bindings in the dynamic environment (e.g. defdyanmic) while others
-          -- prefer the static global environment.
-          ( ( case preference of
+        case tryAllLookups of
+          Just (_, (Right (XObj (Lst ((XObj Def _ _) : _)) _ _))) ->
+            (ctx, Left (HasStaticCall xobj info))
+          Just (_, (Right (XObj (Lst ((XObj (Defn _) _ _) : _)) _ _))) ->
+            (ctx, Left (HasStaticCall xobj info))
+          Just (_, (Right (XObj (Lst ((XObj (External _) _ _) : _)) _ _))) ->
+            (ctx, Left (HasStaticCall xobj info))
```
hellerve, line 74: This is somewhat clunky, I’d be happy to hear alternatives.

scolsen: What about moving this into a function called `checkStatic` or `isStatic` and call it in the Maybe Monad? Then you can selectively call it for lookups that matter as well (unless it matters for all of them, in which case tack it at the end of the maybe chain instead)

```haskell
checkStatic :: (Context, Either EvalError XObj)-> Maybe (Context, Either EvalError XObj)
checkStatic ctx (Right XObj (Lst ((XObj Def ...)))) = Just (ctx, Left HasStaticCall....)
...
checkStatic ctx anything = pure (ctx, anything)

--- assuming it only applies for certain lookups:
(case preference of 
  Dyna -> tryDynamicLookup >>= checkStatic
  Global -> (tryLookup <|> tryDynamicLookup) >>= checkStatic
)

--- or just tack it to the end of the chain
(case preference of 
  Dyna -> tryDynamicLookup
  Global -> (tryLookup <|> tryDynamicLookup)
) <|> (if null p then tryInternalLookup spath else tryLookup spath)
  >>= checkStatic
```

This way it's a lot easier to pull this out or change it later.

scolsen (approved):
I'd recommend moving the explicit casing over the lookup results into a local function you can conveniently add to the chain of Maybe monad ops--this makes it quite a bit more modular and I think better separates concerns.

Otherwise, looks great!

scolsen:
Ahhh, looks like this makes `Introspect` work a bit differently.

hellerve:
@scolsen That was actually a bug on my side. I think it’s important to make it so that `HasStaticCall` is emitted only on the top level; otherwise it’d lead to weirdness inside macros.

scolsen:
> @scolsen That was actually a bug on my side. I think it’s important to make it so that `HasStaticCall` is emitted only on the top level; otherwise it’d lead to weirdness inside macros.

Ah gotcha, thanks!

src/Eval.hs
```diff
@@ -59,23 +59,34 @@ evalStatic ctx xobj = eval ctx xobj PreferGlobal
...
-      pure $
-        fromMaybe
-          (evalError ctx ("Can't find symbol '" ++ show n ++ "'") info) -- all else failed, error.
-          -- Certain contexts prefer looking up bindings in the dynamic environment (e.g. defdyanmic) while others
-          -- prefer the static global environment.
-          ( ( case preference of
+      if shouldResolve
+      then
+        pure $
+          case tryAllLookups of
```
scolsen, line 70: I would still prefer stylistically to have this check as a call in the monad instead of an explicit case statement; should we need to change where this happens it's a lot more arduous to move case statements around than to swap out function calls in a monad:

```haskell
if shouldResolve 
then unwrapLookup (tryAllLookups >>= checkStatic)
else unwrapLookup tryAllLookups
....
where checkStatic :: (Context, Either EvalError XObj) -> Maybe (Context, Either EvalError XObj)
checkStatic ctx (Right (XObj....))= pure (ctx, Left HasStatic...)
checkStatic ctx val = pure ctx val
```

scolsen (approved):

src/Eval.hs
scolsen: Or, maybe better:

```
(tryAllLookups >>= pure . checkStatic)
```

hellerve: That looks great, I like it! Will update.

scolsen (approved):

src/Eval.hs
```diff
@@ -38,12 +38,12 @@ data LookupPreference
   | PreferGlobal
 
 -- Prefer dynamic bindings
-evalDynamic :: Context -> XObj -> IO (Context, Either EvalError XObj)
-evalDynamic ctx xobj = eval ctx xobj PreferDynamic
+evalDynamic :: Bool -> Context -> XObj -> IO (Context, Either EvalError XObj)
```
eriksvedang, line 41: I think replacing the bool with a well named type would make this new code a lot more readable.

eriksvedang (changes requested):

eriksvedang:
This is great!

There are even more cases that create static functions though, like `deftemplate` etc. I like @scolsen:s idea of breaking it out into a special predicate function. Perhaps it could even be implemented by checking for the cases where we don't want static call, if that's shorter.

hellerve:
There is already a predicate function, but I guess it is inlined. I can pull it out and move it into `Obj`.

eriksvedang:
Oh sorry maybe I just read sloppily. In-line is fine.

hellerve:
I rewrote the code using a predicate and added a special type. Ready for another look!

src/Eval.hs
```diff
@@ -37,13 +37,17 @@ data LookupPreference
...
 
+data Resolver
+  = ResolveGlobal
+  | ResolveLocal
+
 -- Prefer dynamic bindings
-evalDynamic :: Context -> XObj -> IO (Context, Either EvalError XObj)
-evalDynamic ctx xobj = eval ctx xobj PreferDynamic
+evalDynamic :: Resolver -> Context -> XObj -> IO (Context, Either EvalError XObj)
+evalDynamic shouldResolve ctx xobj = eval ctx xobj PreferDynamic shouldResolve
```
jacereda, line 46: `shouldResolve` should probably be named `resolver`

eriksvedang (approved):
Awesome, thanks!

PR #1092, opened 2020-12-22
feat: Add parse command

hellerve:
This PR adds the command `parse` which takes in a string and returns an expression. Please note that empty strings and strings containing multiple expressions will result in an error.

Cheers

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -15,6 +15,7 @@ import Lookup
 import qualified Map
 import qualified Meta
 import Obj
+import Parsing (parse)
 import Path
 import Project
 import Reify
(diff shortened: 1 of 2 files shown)


src/Commands.hs
```diff
@@ -841,3 +842,13 @@ toSymbols x = x
 
 commandHash :: UnaryCommandCallback
 commandHash ctx v = pure (ctx, Right (XObj (Num LongTy (Integral (hash v))) (Just dummyInfo) (Just LongTy)))
+
+commandParse :: UnaryCommandCallback
+commandParse ctx (XObj (Str s) i _) =
+  case parse s "command:parse" of
```
jacereda, line 848: `pure $ case ...` so it's easier to notice that all cases are pure?

jacereda (approved):

src/Commands.hs
hellerve: Thanks, I updated this portion!

eriksvedang:
Yay!

PR #1093, opened 2020-12-22
refactor: Remove deps functions

eriksvedang:
Experimenting with removing a bunch of code 😈


hellerve:
I think I need context here: is this not needed at all anymore? Where does e.g. array deletion get member deletion from, then?

eriksvedang:
@hellerve `concretizeType` in Concretize.hs. Since all types in the program get passed through there, any type passed to `Array` will get instantiated. Does this make sense?

hellerve:
Okay, that makes sense, cool!

PR #1094, opened 2020-12-22
feat: overwrite existing interface implementations

scolsen:
This PR alters the behavior of interfaces so that implementations
with the same type signature will overwrite previous implementations
with that signature--before this was a runtime error.

Previously, if a user defined two distinctly named implementations of an
interface that shared a type, Carp would panic and error at runtime if
the interface was called and resolved to the type, since it couldn't
decide which implementation to use from the type alone. After this
commit, we instead issue a warning and overwrite existing
implementations of the same type, so that defining:

```clojure
(defn foo [] 0)
(implements zero foo)
```

will replace `Int.zero` in the `zero` interface's implementation path
list and won't result in a runtime error--instead `foo` will be called
when `zero` is called in a context in which it returns an int:

```clojure
[WARNING] An implementation of the interface zero with type (Fn [] Int)
already exists: Int.zero. It will be replaced by the implementation:
foo.
This may break a bunch of upstream code!
```

test/interface.carp also has a concrete illustration of this case.


TimDeve:
I think the message is a bit vague, maybe something like "This may be unsafe if overriding functions supplied by Carp".

eriksvedang:
This is great!

I don't quite understand how to trigger the following behaviour (in `master`) that you describe though:

> Carp would panic and error at runtime if the interface was called and resolved to the type

Not that it matters much, this new behaviour seems much better.

test/interface.carp
```diff
@@ -29,6 +29,13 @@
 (defn monster [scary?] (if scary? @"RAWR" @"meow"))
 (implements monster monster)
 
+;; If multiple implementations of the same concrete type are provided,
+;; one overwrites the other.
+(defn laugh-monster [times] (String.repeat times "LOL"))
```
eriksvedang, line 34: 👹

src/Primitives.hs
```diff
@@ -234,10 +231,10 @@ define hidden ctx@(Context globalEnv _ typeEnv _ proj _ _ _) annXObj =
         >>= \maybeinterfaces ->
           pure (mapMaybe (`lookupBinder` getTypeEnv typeEnv) (fromMaybe [] maybeinterfaces))
             >>= \interfaceBinders ->
-              pure (foldM (`registerInInterface` binder) ctx interfaceBinders)
-                >>= \case
-                  Left e -> printError (contextExecMode ctx) e >> pure ctx
-                  Right newCtx -> pure newCtx
+              pure (foldl (\(ctx', _) interface -> registerInInterface ctx' binder interface) (ctx, Nothing) interfaceBinders)
+                >>= \(newCtx, err) -> case err of
+                  Just e -> ((printError (contextExecMode ctx) (show e)) >> pure ctx)
```
eriksvedang, line 236: Have you run HLint on this? :)

TimDeve:
> Carp would panic and error at runtime if the interface was called and resolved to the type

I believe Scott means you would get an error when you try to use the interface (at build time) rather than when defining it, which is something I've hit before.

eriksvedang:
OK, gotcha

scolsen:
> > Carp would panic and error at runtime if the interface was called and resolved to the type
> 
> I believe Scott means you would get an error when you try to use the interface (at build time) rather than when defining it, which is something I've hit before.

Yep, to trigger the old behavior at HEAD:

```clojure
(defn foo [] 0)
(implements zero foo)
(defn bar [] (the Int (foo)))
There are several exact matches for the interface `zero` of type `(Fn [] Int)` at line 1, column 24 in 'REPL'
Possibilities:
    Int.zero : (Fn [] Int)
    foo : (Fn [] Int) at REPL:1:1.
```

eriksvedang:
@scolsen Thank you!

src/Primitives.hs
scolsen: Done!

scolsen:
Small update to make sure the overriden binders also update their implements meta and aren't printed when inspecting interfaces:

```clojure
鲤 (defn foo [] 0)
鲤 (implements zero foo)
[WARNING] An implementation of the interface zero with type (Fn [] Int) already exists: Int.zero. It will be replaced by the implementation: foo.
This may break a bunch of upstream code!
鲤 (Introspect.implements? 'zero Int.zero)
=> false
鲤 :i Int.zero
Int.zero : (Fn [] Int)
  Defined at line 64, column 3 in 'Carp/core/Int.carp'
  Implements: ()
```

Also updated the error message, tried to keep it generic though, thanks for the suggestion @TimDeve!

eriksvedang:
Really nice, awesome work!

PR #1097, opened 2020-12-23
feat: Unify aupdate and aupdate! with other update functions

### LINKED ISSUE #471: Unify signatures for `Array.aupdate` and struct update functions
At the moment, the signatures for array updaters and struct updaters are different:

```
鲮 (load "Vector.carp")
鲮 (type Vector2.update-x)
Vector2.update-x : (λ [(Vector2 a), (Ref (λ [a] a))] (Vector2 a))
鲮 (type Array.aupdate)
Array.aupdate : (λ [(Array a), Int, (Ref (λ [&a] a))] (Array a))
```

Specifically the function being passed at the end. The function could take ownership over the element without any problems (since the element will be replaced anyway), but it only does in struct updaters. This could and should be unified.

Cheers

eriksvedang:
Closes #471 

What we need to not have to resort to a template here is something like `Array.unsafe-nth-value` that returns the actual element in the array (but does not copy it). I think I'll switch to that implementation, seems nicer.


eriksvedang:
Probably should update `UnsafeArray.aupdate!` at the same time...

hellerve (approved):
this seems like a great change!

core/Array.carp
```diff
@@ -220,14 +220,20 @@ If the element is not found, returns `Nothing`")
           (set! c (Int.inc c))))
       c))
 
+  (doc unsafe-nth-value "returns the value at index `i` of an array `a` (just like unsafe-nth) but does not take its reference, and does *not* copy the value. Should only be used for optimizations and when you know what you're doing, circumvents the borrow checker!")
```
hellerve, line 223: Could we make the reference to `unsafe-nth` a link here (using `[unsafe-nth](#unsafe-nth)`)?

eriksvedang:
I'll make a new PR for this.

PR #1099, opened 2020-12-23
feat: Add "rest" function to Array

Efimero:
This is a very common idiom in Clojure that gets used a lot in reducers. It's nice to have.

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -183,7 +183,7 @@ If the array is empty, returns `Nothing`")
 
   (doc rest "gets all but the first element from the array.")
   (defn rest [xs]
-    (slice xs 1 (length xs)))
+    (suffix xs 1))
 
   (doc reverse "reverses an array.")
   (defn reverse [a]
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -181,6 +181,10 @@ If the array is empty, returns `Nothing`")
   (defn suffix [xs start-index]
     (slice xs start-index (length xs)))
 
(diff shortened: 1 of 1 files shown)


core/Array.carp
```diff
@@ -181,6 +181,10 @@ If the array is empty, returns `Nothing`")
   (defn suffix [xs start-index]
     (slice xs start-index (length xs)))
 
+  (doc rest "gets all but the first element from the array.")
+  (defn suffix [xs]
```
Efimero, line 185:
```suggestion
  (defn rest [xs]
```

Efimero:
```suggestion
  (defn rest [xs]
```

core/Array.carp
```diff
@@ -181,6 +181,10 @@ If the array is empty, returns `Nothing`")
   (defn suffix [xs start-index]
     (slice xs start-index (length xs)))
 
+  (doc rest "gets all but the first element from the array.")
+  (defn rest [xs]
+    (slice xs 1 (length xs)))
```
hellerve, line 186: We could use `(suffix xs 1)` here to simplify it a little.

hellerve (approved):
Wow, I thought we had that function already! Good addition!

I left a little style nitpick, but I think we’re good to go.

core/Array.carp
Efimero, line 186:
```suggestion
    (suffix xs 1))
```

PR #1100, opened 2020-12-23
fix: fix errors with set!

### LINKED ISSUE #1098: Dynamic `set!` just returns the value if no binder is found
If no variable is found that can be reset by `set!`, it just returns the value and does not emit an error or warning. This should be an error instead, in my opinion.

Cheers

scolsen:
Notably, don't type check dynamic bindings (which can be set to
whatever) and eliminate a hang that resulted from not handling an error
at the end of the `set!` call. Also refactors some of the code in
efforts to make it a bit cleaner.

Also adds an error when `set!` can't find the variable one calls set!
on.

fixes #1098


scolsen:
Here's the new error:

```clojure
鲤 (set! x 'foo)
I couldn't find the variable x, did you define it using `def` or `defdynamic`? at REPL:1:7.
```

PR #1102, opened 2020-12-26
fix: manage nested modules in the type environment

scolsen:
A misplaced type env reference caused a bug in which types could not be
placed in modules more than a single level deep. Similarly, existing
modules in the type env were always overwritten when new types were
added to modules that were not disjoint. This commit fixes both issues.

In the future, we should likely make module management functions more
general and call them to manage modules in both the value and type envs.

e.g.:

```
鲤 :i Foo
Foo : Module = {
    Bar : Module = {
        Baz : Module = {
            Boo : Type
        }
        Boo : Type
    }
}
  Defined at line 0, column 0 in 'dummy-file'
Foo : Module = {
    Bar : Module = {
        Baz : Module = {
            Boo : Module = {
                copy : (Fn [(Ref Foo.Bar.Baz.Boo q)] Foo.Bar.Baz.Boo)
                delete : (Fn [Foo.Bar.Baz.Boo] ())
                get-tag : (Fn [(Ref Foo.Bar.Baz.Boo q)] Int)
                prn : (Fn [(Ref Foo.Bar.Baz.Boo q)] String)
                str : (Fn [(Ref Foo.Bar.Baz.Boo q)] String)
            }
        }
        Boo : Module = {
            copy : (Fn [(Ref Foo.Bar.Boo q)] Foo.Bar.Boo)
            delete : (Fn [Foo.Bar.Boo] ())
            get-tag : (Fn [(Ref Foo.Bar.Boo q)] Int)
            prn : (Fn [(Ref Foo.Bar.Boo q)] String)
            str : (Fn [(Ref Foo.Bar.Boo q)] String)
        }
    }
}
  Defined at line 0, column 0 in 'dummy-file'
```

Is possible now--before one of the `Boo` types would be overwritten.

### DIFF
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -11,7 +11,7 @@ import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Either (rights)
 import Data.Functor ((<&>))
 import Data.List (union)
-import Data.Maybe (fromJust, fromMaybe, mapMaybe, maybeToList)
+import Data.Maybe (fromJust, fromMaybe, mapMaybe)
 import Deftype
 import Emit
 import Env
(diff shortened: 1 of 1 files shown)


src/Primitives.hs
```diff
@@ -672,11 +672,15 @@ primitiveDeftype xobj ctx (name : rest) =
...
-                  folder (contx, prev) pathstring = (contx {contextTypeEnv = TypeEnv $ envInsertAt (getTypeEnv typeEnv) (SymPath (maybeToList (envModuleName prev)) pathstring) (holderModule pathstring prev)}, holderEnv pathstring prev)
-                  wHolders = fst (foldl folder (ctx, getTypeEnv typeEnv) pathStrings)
+                  holderModule name'' prevEnv priorPaths tyenv =
+                    case lookupBinder (SymPath priorPaths name'') tyenv of
+                      Just existing@(Binder _ (XObj (Mod _) _ _)) -> existing
+                      _ -> Binder emptyMeta (XObj (Mod (holderEnv name'' prevEnv)) (Just dummyInfo) (Just ModuleTy))
+                  folder (contx, prev, priorPaths) pathstring =
+                    (contx {contextTypeEnv = TypeEnv $ envInsertAt (getTypeEnv (contextTypeEnv contx)) (SymPath priorPaths pathstring) (holderModule pathstring prev priorPaths (getTypeEnv (contextTypeEnv contx)))}, holderEnv pathstring prev, priorPaths ++ [pathstring])
+                  (wHolders, _, _) = (foldl folder (ctx, getTypeEnv typeEnv, []) pathStrings)
```
jacereda, line 681: Maybe it would be better to use `foldl'` ?

jacereda: Nevermind, I opened a separate PR for that.

PR #1103, opened 2020-12-27
refactor: Use different primitive types for each arity

jacereda:


src/Eval.hs
```diff
@@ -361,7 +361,13 @@ eval ctx xobj@(XObj o info ty) preference resolver =
...
         XObj (Lst [XObj (Command _) _ _, sym, XObj (Arr params) _ _]) i _ : args ->
           badArity (getName sym) params args i
-        x@(XObj (Lst [XObj (Primitive prim) _ _, _, _]) _ _) : args -> getPrimitive prim x ctx args
+        [e@(XObj (Lst [XObj (Primitive (NullaryPrimitive nullary)) _ _, _, _]) _ _)] -> nullary e ctx
+        [e@(XObj (Lst [XObj (Primitive (UnaryPrimitive unary)) _ _, _, _]) _ _), x] -> unary e ctx x
+        [e@(XObj (Lst [XObj (Primitive (BinaryPrimitive binary)) _ _, _, _]) _ _), x, y] -> binary e ctx x y
+        [e@(XObj (Lst [XObj (Primitive (TernaryPrimitive ternary)) _ _, _, _]) _ _), x, y, z] -> ternary e ctx x y z
+        [e@(XObj (Lst [XObj (Primitive (QuaternaryPrimitive quaternary)) _ _, _, _]) _ _), x, y, z, w] -> quaternary e ctx x y z w
+        e@(XObj (Lst [XObj (Primitive (VariadicPrimitive variadic)) _ _, _, _]) _ _) : args -> variadic e ctx args
+        XObj (Lst [XObj (Primitive _) _ _, sym, XObj (Arr params) _ _]) i _ : args -> badArity (getName sym) params args i
```
hellerve, line 370: The usage of `badArity` leads to errors like this here:

```
You’ll have to provide [x], [y] as well.
```

I suppose that’s because of how we provide the arguments in `makePrim'` below.

hellerve (commented):
I think the error message problem I pointed out should be fixed, but other than that, this is an important quality of life improvement!

src/Eval.hs
```diff
@@ -1234,54 +1238,52 @@ dynamicOrMacro ctx pat ty name params body = do
...
-          (newCtx', evaledB) <- evalDynamic ResolveLocal newCtx b
-          pure $ case evaledB of
-            Left e -> (newCtx, Left e)
-            Right (XObj (Bol bb) _ _) ->
-              (newCtx', Right (boolToXObj bb))
-            Right b' -> evalError ctx ("Can’t call `or` on " ++ pretty b') (xobjInfo b')
-        else pure (newCtx, Right falseXObj)
-    Right a' -> pure (evalError ctx ("Can’t call `or` on " ++ pretty a') (xobjInfo a'))
-primitiveAnd _ _ _ = error "primitiveand"
+-- primitiveAnd :: Primitive
```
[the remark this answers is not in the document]
jacereda: Done.

src/Eval.hs
jacereda: Oops, that was also broken for commands. Fixed.

hellerve (approved):

PR #1104, opened 2020-12-27
refactor: Replace foldl uses with foldl'

jacereda:


jacereda:
`foldl` is most often a bad thing leading to space leaks or increased memory consumption. I haven't reviewed properly all the instances, but I would just assume it's equivalent. The biggest danger would be getting a false exception, but that's usually easier to catch than a space leak.

eriksvedang (approved):

eriksvedang:
Looks great, needs rebasing.

jacereda:
Rebased.

PR #1108, opened 2020-12-30
fix members primitive

hellerve:
This PR fixes the `members` primitive, which has to implement custom lookup logic to get to the type by looking at modules.

Cheers

### DIFF
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -369,7 +369,7 @@ dynamicOrMacroWith ctx producer ty name body = do
 primitiveMembers :: UnaryPrimitiveCallback
 primitiveMembers _ ctx target = do
   let typeEnv = contextTypeEnv ctx
-  case bottomedTarget of
+  case bottomedTarget target of
     XObj (Sym path@(SymPath _ name) _) _ _ ->
       case lookupBinder path (getTypeEnv typeEnv) of
         Just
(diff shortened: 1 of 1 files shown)


src/Primitives.hs
```diff
@@ -415,16 +415,16 @@ primitiveMembers _ ctx target = do
...
+    bottomedTarget t =
+      case t of
         XObj (Sym targetPath _) _ _ ->
           case lookupBinder targetPath env of
             -- this is a trick: every type generates a module in the env;
             -- we’re special-casing here because we need the parent of the
             -- module
-            Just (Binder _ (XObj (Mod _) _ _)) -> target
+            Just (Binder _ (XObj (Mod _) _ _)) -> t
             -- if we’re recursing into a non-sym, we’ll stop one level down
```
scolsen, line 426: Seems like this comment no longer holds since we recurse again into the contents of the Binder (`x`) now

scolsen (approved):

src/Primitives.hs
hellerve: It actually does, hence the comment about "one level down"—otherwise we should stop right here.

scolsen: Ah, you're right! -- I missed the catch all case below this line

scolsen (approved):

PR #1109, opened 2020-12-30
feat: Recognize the NetBSD platform.

dressupgeekout:
These were the only changes I needed to make to Carp's code itself in order to get it to build and run on NetBSD/amd64 9.0.

### DIFF
--- a/core/Platform.carp
+++ b/core/Platform.carp
@@ -36,6 +36,9 @@
 (doc freebsd-target? "are we targeting FreeBSD?")
 (defdynamic freebsd-target? (target-os? "freebsd"))
 
+(doc netbsd-target? "are we targeting NetBSD?")
+(defdynamic netbsd-target? (target-os? "netbsd"))
+
 (doc posix-target? "are we targeting a POSIX platform?")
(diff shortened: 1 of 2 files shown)


hellerve:
This looks great! Semi-related to this PR: do Github actions currently support NetBSD as a host OS? (Note: I know nothing of Github actions, so this might be a really stupid question)

dressupgeekout:
@hellerve I honestly have no clue. My guess is no, but maybe I could look into it? I'm not that familiar with GitHub actions either.

dressupgeekout:
Actually, now that I think about it -- even if GitHub actions did support NetBSD, it wouldn't successfully perform CI for _this_ PR, because I needed to update and/or invent several Haskell packages in the first place

hellerve (approved):

hellerve:
It looks like only Windows, MacOS, and Linux are supported out of the box, but you can host your own runners? For now this doesn’t seem to be worth the effort, but good to know!

In any case, I think this PR looks fine :)

eriksvedang:
Cool, nice addition!

Issue #1095, opened 2020-12-23
RFC: Implement `delete` for `Ptr` and add `Leak` type


scolsen:
Now that `delete` is an interface, we can easily implement it for `Ptr` and allow Carp to also manage this type, all we'd need to add is:

`(implements delete Pointer.free)`

since `Pointer.free` already has the right signature. Doing so would limit the purpose of `Ptr`, making it a type primarily for C interop, that still leverages Carp's memory management in carp code--previously it served a double function, being an unmanaged type also intended for interop.

Implementing `delete` would remove this "unmanaged memory" use case for `Ptr` and make it strictly an interop type, after doing so, we could neatly reintroduce this "unmanaged type" functionality with a new type `Leak` which is precisely an "unmanaged" pointer (core will implement `delete` on it as a no-op). The type name also communicates that the memory needs to be managed manually. `Unsafe.leak` could also return this type, indicating at the type level precisely where something was leaked in the program (the sig of this function would change from `[a] ()` to `[a] (Leak a)`). 

Since `Leak`'s delete implementation would be a no-op, destructors on this type *must* be called manually--in other words, `Leak` (in core) is the type of things whose destructors *must* be called manually (unless one defines an implementation of `delete` for it downstream).

Other names could be:

`Memory` (this type is just raw, unmanged memory to something)
`Unmanged` (this type is not managed)
`Eternal` (for eternal lifetime, unless manually killed)

## Why?

I think the nice thing about having a type like `Leak` is that it'd grant us a dedicated type in Core that communicates at the type level precisely when something is not going to be managed by Carp's borrow system. One can define interfaces against this and communicate the "unmanaged" nature of values to the user, e.g. as a library user, seeing a function with signature `[b] (Leak c)` would immediately indicate that I'm getting a long-lived, borrow-checker immune value that I'll have to delete/clean-up manually by calling some function on it in my Carp code--that is, I can be confident `delete` emissions won't free it. Or, as another example, if the body of some function in a library manually calls a special deleter on an argument, this can be partially communicated (or more so hinted at) `[(Leak Foo)] b`.

scolsen:
@carp-lang/maintainers

hellerve:
I think this is a great proposal!

[...]

PR #1111, opened 2021-01-02
chore: Abuse deftemplate to get rid of Address

jacereda:
I don't remember if this has been discussed before. If you think it's worth it I can polish the patch (it emits a double semicolon right now and I don't know where to place the definition).

### DIFF
--- a/core/Blitable.carp
+++ b/core/Blitable.carp
@@ -57,3 +57,7 @@
 (defmodule Uint64
   (defn blit [x] (the Uint64 x))
   (implements blit Uint64.blit))
+
+(defmodule Pointer
+  (defn blit [x] (the (Ptr a) x))
+  (implements blit Pointer.blit))
--- a/core/Core.carp
+++ b/core/Core.carp
@@ -15,7 +15,7 @@
 (system-include "core.h")
 (system-include "carp_memory.h")
 
-(deftemplate address (Fn [a] (Ptr a)) "#define $NAME(a) (&(a))" "")
+(deftemplate address (Fn [(Ref a)] (Ptr a)) "$a * $NAME($a * a)" "$DECL { return a; }")
 
 (load-once "Interfaces.carp")
 (load-once "Blitable.carp")
--- a/core/Core.carp
+++ b/core/Core.carp
@@ -15,6 +15,8 @@
 (system-include "core.h")
 (system-include "carp_memory.h")
 
+(deftemplate address (Fn [a] (Ptr a)) "#define $NAME(a) (&(a))" "")
+
 (load-once "Interfaces.carp")
 (load-once "Blitable.carp")
 (load-once "Bool.carp")
--- a/core/Pointer.carp
+++ b/core/Pointer.carp
@@ -1,6 +1,10 @@
 (defmodule Pointer
   (doc eq "checks two pointers for equality.")
   (deftemplate eq (Fn [(Ptr p) (Ptr p)] Bool) "bool $NAME($p *p1, $p *p2)" " $DECL { return p1 == p2; }")
+  (implements = Pointer.eq)
+  (doc eq "checks two pointer references for equality.")
+  (deftemplate ref-eq (Fn [(Ref (Ptr p)) (Ref (Ptr p))] Bool) "bool $NAME($p **p1, $p **p2)" " $DECL { return *p1 == *p2; }")
+  (implements = Pointer.ref-eq)
   (doc to-ref "converts a pointer to a reference type.
 
 The user will have to ensure themselves that this is a safe operation.")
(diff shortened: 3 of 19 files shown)


hellerve:
That’s pretty smart!

eriksvedang:
I actually with something similar in #1097, but for dereferencing pointers (adding `*` in front, instead). I think it's worth pursuing, making the compiler smaller and all.

eriksvedang:
I think we need some good tests for it though, since the address operator is not actually used much in the tests code, iirc.

jacereda:
Where should the `deftemplate` go?

jacereda:
Any idea on how to test it?

TimDeve:
Being a C macro it will break if you try to pass address as a ref callback, no? Doesn't address have to be a built-in instead of a function?

hellerve:
Sidenote: currently `address` will also break then, because it is a special form.

TimDeve:
Do you not get a compiler error if you try to use a built-in/special form as a function?

hellerve:
Yes, but a rather unfortunate one:

```
> (Array.copy-map &address &[1 2 3])
I didn’t understand the form `address` at line 1, column 18 in 'REPL'.

Is it valid? at REPL:1:1.
```

I guess this is better than the error you get when using the `deftemplate` version, which generates invalid C. Sadly just making a template function might not work either, due to value semantics.

scolsen:
> Sadly just making a template function might not work either, due to value semantics.

What do you mean by value semantics in this context--are you referring to how the borrow checker might handle this? I think a template function should work ok based on some of the templates in `Pointer.carp`, but I'm probably missing a nuance to borrowing/ref management--I suppose the Ptr's in the Array would be unowned, but I think that's ok?

jacereda:
We could change `address` to receive a `Ref`, that way it could be a plain function instead of a macro.

scolsen:
Looks like there needs to be some changes to the SDL sources too, but otherwise this should work and I like this approach! `address` only taking a Ref makes sense to me.

hellerve:
> > Sadly just making a template function might not work either, due to value semantics.
> 
> What do you mean by value semantics in this context--are you referring to how the borrow checker might handle this? I think a template function should work ok based on some of the templates in `Pointer.carp`, but I'm probably missing a nuance to borrowing/ref management--I suppose the Ptr's in the Array would be unowned, but I think that's ok?

The problem is with the value semantics of C: you can’t take the address of a stack variable; `Int`, for instance, would be one of those.

scolsen:
> > > Sadly just making a template function might not work either, due to value semantics.
> > 
> > 
> > What do you mean by value semantics in this context--are you referring to how the borrow checker might handle this? I think a template function should work ok based on some of the templates in `Pointer.carp`, but I'm probably missing a nuance to borrowing/ref management--I suppose the Ptr's in the Array would be unowned, but I think that's ok?
> 
> The problem is with the value semantics of C: you can’t take the address of a stack variable; `Int`, for instance, would be one of those.

Gotcha thanks! I guess that should be taken care of via `blit` and checking that whatever is passed to `address` does not implement `blit`--but right now that would again limit us to handling `address` in the compiler unless we add some general mechanism for expressing "assert this interface isn't implemented" as discussed in #1095

jacereda:
I assumed the use case for `address` was always related to interfacing with C and the current state of the PR (plain function taking a `Ref`) seems to handle that. Do you have examples where this wouldn't work?

scolsen:
> I assumed the use case for `address` was always related to interfacing with C and the current state of the PR (plain function taking a `Ref`) seems to handle that. Do you have examples where this wouldn't work?

That's my interpretation too. I think `Unsafe.leak` is to make something unmanaged.

TimDeve:
Ok, works for me, hadn't seen the update . Are we saying it's alright to make Ptr `blit` because it's unmanaged?

core/Blitable.carp
```diff
@@ -57,3 +57,7 @@
 (defmodule Uint64
   (defn blit [x] (the Uint64 x))
   (implements blit Uint64.blit))
+
+(defmodule Pointer
```
eriksvedang, line 61: Why is this instance needed?

jacereda: `test/address.carp` won't build unless we have this in place IIRC.

eriksvedang:
I had some idea for a tricky situation where this approach maybe wouldn't work, but I can't for my life remember what it was. Probably it wasn't a problem after all, so let's put it in. You mentioned you wanting to clean it up @jacereda, though it looks totally fine to me. Should we just merge it as is?

jacereda:
IIRC the cleanup part was referring to removing the double semicolon and I think https://github.com/carp-lang/Carp/pull/1111/commits/6e917d5dae5f8c44449a7dbe02353d4638a3661d handled that.

PR #1112, opened 2021-01-02
feat: Reduce binary size with more compact exhaustiveness error repor…

jacereda:
…ting

### DIFF
--- a/core/core.h
+++ b/core/core.h
@@ -34,6 +34,12 @@ typedef char CChar;
     } while (0)
 #endif
 
+#define UNHANDLED(file, line)                                                  \
+    do {                                                                       \
+        printf("Unhandled case in 'match' expression at %s:%d\n", file, line); \
+        abort();                                                               \
+    } while (0)
+
 // Array
 typedef struct {
     size_t len;
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -20,6 +20,7 @@ import Info
 import qualified Map
 import qualified Meta
 import Obj
+import Path (takeFileName)
 import Project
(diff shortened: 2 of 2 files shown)


eriksvedang:
Can you please elaborate on how this reduces binary size? (honest question)

Cheers!

jacereda:
Each distinct string literal with the form "Unhandled case in ...<file/line/col info>" is embedded in the binary. This replaces those strings with a constant string (the format string) which happens to be the same in all cases plus and will be hopefully present only once in the binary, plus some formatting code that will hopefully be smaller than the string it replaces.

jacereda:
Sizes for `test/result.carp` with `-O2 -flto':

Before:
```
   text	   data	    bss	    dec	    hex	filename
  34320	    984	     80	  35384	   8a38	out/Untitled
```

After:
```
    text	   data	    bss	    dec	    hex	filename
  33624	    976	     72	  34672	   8770	out/Untitled
```

Of course, the real win would be to have exhastiveness checks, but while that arrives it might be a good thing.

jacereda:
Also, if we wrap the UNHANDLED macro with `#ifndef UNHANDLED`, someone really concerned about code size could just replace that with a single `abort` call adding a define to `profile.carp`.

hellerve:
The `ifndef` wrap is not added yet; should we do that in a separate PR? It seems like a worthy addition.

jacereda:
An option not requiring the `ifndef`: Just include a header doing:

```
#undef UNHANDLED
#define UNHANDLED(f,l) abort()
```

PR #1118, opened 2021-01-11
fix: account for interfaces in function lookups

scolsen:
An older version of Carp only allowed users to implement interfaces by
defining a function of the same name. Since PR 769 and roughly commit
040e9e4 however, this was no longer the case, since we added a primitive
for implementing interfaces, allowing users to implement interfaces
  using functions with whatever name they want.

When this happened, the function lookups that happen during
concretization were never updated to reflect the fact that interface
implementations might have different names than their corresponding
interfaces. It's rare, but this can lead to errors for template
functions in the compiler.

For example, registered types automatically get an implementation of
`prn` with the same name. This function takes a ref to the registered
type as an argument. If a user wants to use such a type as an array
member however, they must provide a prn that takes a value argument
instead of a ref. If the user still wants to keep the ref variant handy,
however, they need to give this new prn implementation a different name,
otherwise Carp will think they are redefining the existing function.
However, using an arbitrary name would result in this second prn
instance not being found during the lookup fixed in this commit, since
it did not match the name of the interface exactly.

After this commit, this issue will no longer crop up, as the
concretize.hs function lookup now accounts for the fact that, if the
function it is looking for is an interface, it may need to find
functions with names that differ from the name it received as an
argument; e.g. a lookup for `prn` could result in finding a function
named `foo` that is a valid implementation of `prn`.

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -671,8 +671,12 @@ concretizeDefinition allowAmbiguity typeEnv globalEnv visitedDefinitions definit
           Left $ CannotConcretize definition
 
 -- | Find ALL functions with a certain name, matching a type signature.
-allFunctionsWithNameAndSignature :: TypeEnv -> Env -> String -> Ty -> [(Env, Binder)]
-allFunctionsWithNameAndSignature typeEnv env functionName functionType =
+-- When the functionName argument denotes an interface, the name will match iff either:
+--   1. The name of the binding matches functionName exactly OR
+--   2. The name of the binding matches one of the names in the interface's implementation paths
+-- For all other functions, the name must match exactly, and in all cases, the signature must match.
+allImplementations :: TypeEnv -> Env -> String -> Ty -> [(Env, Binder)]
+allImplementations typeEnv env functionName functionType =
   filter (predicate . xobjTy . binderXObj . snd) foundBindings
   where
     predicate (Just t) =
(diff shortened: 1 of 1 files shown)


scolsen:
Hopefully the explanation here is at least a little bit clear--it's on the later side here and a good explanation wasn't coming to me.

The chat has some more background-context on a concrete issue that sparked this.

scolsen:
I didn't do any through investigation, but this is the same code path that emits those "too many delete" etc. warnings--likely a similar approach could fix those--finding somewhere that we rely on the old implementation name == interface name assumption and adjusting it.

src/Concretize.hs
```diff
@@ -671,19 +671,33 @@ concretizeDefinition allowAmbiguity typeEnv globalEnv visitedDefinitions definit
           Left $ CannotConcretize definition
 
 -- | Find ALL functions with a certain name, matching a type signature.
-allFunctionsWithNameAndSignature :: Env -> String -> Ty -> [(Env, Binder)]
-allFunctionsWithNameAndSignature env functionName functionType =
-  filter (predicate . xobjTy . binderXObj . snd) (multiLookupEverywhere functionName env)
+allFunctionsWithNameAndSignature :: TypeEnv -> Env -> String -> Ty -> [(Env, Binder)]
```
hellerve, line 674: Is this name still accurate?

hellerve (approved):
This looks good to me!

Should we open an issue for the behavior of `lookupImplementations`?

src/Concretize.hs
eriksvedang: Agree that is seems a little weird now with this fix...

eriksvedang:
This problem makes sense, didn't see it coming though!

src/Concretize.hs
scolsen: Good point! What do we think of simplifying to `allImplementations` where an implementation is meant to indicate a function that:

1. has a matching type signature.
2. has a matching name, either through membership in the set of paths when `name` denotes an interface, otherwise an exact match.

As an alternative, we could be a bit verbose but explicit and call it `allFunctionsWithNameAndSignatureOrAllImplementations`

eriksvedang: `allImplementations` 👍

scolsen: renamed in 3f7c96d!

scolsen:
> This looks good to me!
> 
> Should we open an issue for the behavior of `lookupImplementations`?

Good idea, added! #1123

PR #1119, opened 2021-01-11
Fix closure history on dynamic eval

hellerve:
This PR fixes the closure history for dynamic evaluation. Since the history is only used for tracebacks, it wouldn’t make much sense to use the original history—which tells us where the closure was created—to report errors, but rather the history of the call, to tell us where the closure was actually used.

NB: Maybe the history of the closure creation could also be interesting in some cases, though, to trace back where it’s coming from?

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -302,7 +302,7 @@ eval ctx xobj@(XObj o info ty) preference resolver =
                 (newCtx, evaledArgs) <- foldlM successiveEval (ctx, Right []) args
                 case evaledArgs of
                   Right okArgs -> do
-                    (_, res) <- apply c body params okArgs
+                    (_, res) <- apply c{contextHistory=contextHistory ctx} body params okArgs
                     pure (newCtx, res)
                   Left err -> pure (newCtx, Left err)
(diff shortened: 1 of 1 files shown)


eriksvedang:
Nice, this seems very helpful! But like you say, sometimes maybe you want the other information, I'm not sure.

eriksvedang (approved):

eriksvedang:
(I'm fine with merge:ing this though)

hellerve:
I think if we have  to choose, this is strictly more useful. Consider this example:

```clojure
(defndynamic x []
  (fn [] (macro-error "oops")))
  
 (defndynamic y []
   ((x)))
   ```
   
   In the old code, the source location of the error would be in `x`, not `y`, even though `y` produces the error. This is misleading, since we lose the info of who called the closure in `x`.  Does that make sense?

eriksvedang:
> I think if we have to choose, this is strictly more useful. Consider this example:
> 
> ```clojure
> (defndynamic x []
>   (fn [] (macro-error "oops")))
>   
>  (defndynamic y []
>    ((x)))
> ```
> 
> In the old code, the source location of the error would be in `x`, not `y`, even though `y` produces the error. This is misleading, since we lose the info of who called the closure in `x`. Does that make sense?

Thanks for the example, yes it makes a lot of sense!

Issue #1101, opened 2020-12-24
Closures cache all environments


hellerve:
When closures are created, they save the state of the environments around them so they can lookup values later. This works fine in most cases, but since the environment is persistent, updates to the global environment etc will not be saved. Even types and interface implementations will not be present.

Consider this use case:

```clojure
(defdynamic t-member-names ; could be a defndynamic here, but bear with me
  (fn [] (map car (members T))))
  
(deftype T [x Int])

(t-member-names) ; will not be able to find T, because it wasn’t defined when the fn was
```

This is a problem that cropped up during the implementation of a generic `derive`, which heavily relies on dynamic `fn`s.

Cheers

scolsen:
Ahhh so that’s the cause of “fixed definition order” I’ve seen this before too.

Could we solve this by just setting the parent of the parent env of the closure to the current context env structure art call time?

hellerve:
> Could we solve this by just setting the parent of the parent env of the closure to the current context env structure at call time?

This would only work if we didn’t want to support arbitrary nesting of `fn`, I think? But it might be worth trying out!

eriksvedang:
I've been trying to come up with some ideas for how to fix this but it's quite hard. How important is it for implementing your code? (Quite important I assume...) 

We could capture global variables as paths rather than envs but that doesn't work for local vars. Maybe we need mutable cells of some sort when capturing dynamic local vars. They could be referred to via a unique ID or something (and stored in the Context). 

Ideally we can just solve this with the existing envs but I dont quite see how...

hellerve:
Well, the problem is that you won’t be able to read from or modify global variables from within a closure, which is a pretty big problem IMO.

I think we might have to go with `IORef`s to environments in the mid-to-long term to rectify these issues. That’s pretty annoying, but I think necessary?

eriksvedang:
Yeah that's probably the most complete solution.

eriksvedang:
> Ahhh so that’s the cause of “fixed definition order” I’ve seen this before too.
> 
> 
> 
> Could we solve this by just setting the parent of the parent env of the closure to the current context env structure at call time?

I've pondered this some more and I think this approach can work too. We just need to be careful with nested lambdas and such. I think the algorithm should be something like traversing upwards and replacing each named env with the latest one?

scolsen:
> > Ahhh so that’s the cause of “fixed definition order” I’ve seen this before too.
[...]

PR #1120, opened 2021-01-11
feat: Derive

hellerve:
This PR adds `derive`, a macro for deriving arbitrary interfaces for structs based on their members. It is also pluggable, meaning that you can write your own derivers.

This PR is a draft, because it depends on #1101 being fixed ~~and because documentation has to be written on the feature~~.

Cheers

EDIT: Documentation has been written.


hellerve:
For documentation purposes I wanted to put `derive` into a module, but it seems like #1121 prevents me from doing so.

hellerve:
This is ready for review!

docs/Derive.md
```diff
@@ -0,0 +1,107 @@
+# Derive
+
+`derive` is a mechanism that automatically determines how to implement
+interfaces for datatypes based on their members. It also allows you to write
+your own rules for `derive`, called a `deriver`, but most users will likely
+not use that feature.
+
+If you’d like to learn how to `derive` interfaces for your types, read the
+first section of this document. If you’d like to provide a deriver for an
```
scolsen, line 9: Maybe link to the sections here using heading anchors?` [first section](#i-using-derive)`

core/Derive.carp
```diff
@@ -0,0 +1,92 @@
...
+  (defndynamic derivable? [f]
+    (let [deriver (get-deriver f Derive.derivers)]
+      (not (empty? deriver))))
+
+  (doc derivables "returns the list of currently derivable interfaces.")
+  (defndynamic derivables []
+    (map car Derive.derivers))
+
+  (doc derive "derives an interface function `f` for a type `t`.")
+  (defmacro derive [t f]
```
scolsen, line 61: I don't think it's totally necessary, but I feel this might read a bit better if `t` is the second argument:

`(derive zero Point)`

This also would make it follow the order of `implements`

`(implements zero foo)`

docs/Derive.md
```diff
@@ -0,0 +1,107 @@
...
+```clojure
+(deftype Point [
+  x Int
+  y Int
+])
+(derive Point zero)
+(derive Point =)
+```
+
+The code above will enrich the type `Point` by providing implementations for
```
scolsen, line 26: How about:

"The code above will provide implementations of `zero` and `=` for the type `Point` based on its members."

docs/Derive.md
```diff
@@ -0,0 +1,107 @@
...
+interface name needs to be quoted.
+
+If either of the preconditions above is not met, you will have to write your
+own version of these functions, and may not use `derive`.
+
+Some users might want to be able to derive update interfaces that take a type,
+do the same thing to all its members, and return it. A good example for this
+in the context of `Point` is `inc`.
+
+While generally this might require you to write your own deriver—see section II
```
scolsen, line 47: Link to section 2

docs/Derive.md
```diff
@@ -0,0 +1,107 @@
...
+of this document to learn how to do that—, Carp provides you with a special
```
scolsen, line 48: I would delete "you with", so "Carp provides a special dynamic..."

scolsen (approved):
Amazing! Such a useful feature.

I left a few comments, mostly nits.

core/Derive.carp
```diff
@@ -0,0 +1,92 @@
...
+generate a body when given a type `body`.
+
+Example:
+```
+(make-deriver 'zero []
+  (fn [t]
+    (cons 'init
+      (map (fn [_] '(zero)) (members t)))))
+```")
+  (defndynamic make-deriver [f args body]
```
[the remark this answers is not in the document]
hellerve: That makes sense! Resolved in `0b8ae4b6`.

core/Derive.carp
hellerve: Fair enough; what’s the consensus here, @carp-lang/maintainers?

hellerve:
Thanks for the edits, @scolsen! I implemented them (except for the order of arguments in `derive`, which I’m waiting for feedback on).

core/Derive.carp
eriksvedang: I think it makes a lot of sense to have the same order for derive & implements. Since we already have a lot of code that relies on the `implements` ordering it seems reasonable to go with that.

TimDeve: Always felt that the order in `implements` should be the other way around, especially if you read it "infix": `zero implements foo`. But there is certainly something to be said for consistency.

scolsen: > Always felt that the order in implements should be the other way around, especially if you read it "infix": zero implements foo.
But there is certainly something to be said for consistency.

That's a fair point, if others feel this way too, I'm ok with making a change on the `implements` side. The reason it has the order is because when considering prefix notation, `(implements zero foo)` made more sense to me--basically imagine there's a comma or colon when reading it off as english-- `implements zero: foo`--it also allows the form to take varargs in the future `(implements zero foo...)`. 

`(implements foo zero)`, putting the implementation first, felt odd and seemed more likely to create friction when skimming code even if it corresponds to the correct infix order. At the very least, I'd personally assume `foo` was the interface in that case, why precisely I don't know but it's how my brain wants to naturally parse it (this may well be a personal problem, so I'm happy to go along with consensus).

Another argument for changing the order on the `implements` side would be that it's more consistent, I think, with some of our other forms, `meta` for example.

eriksvedang: It's definitely not super clear which one is most natural. I usually don't try to make it "read naturally" in the natural language kind of way. Instead (even when now programming Haskell) I use some kind of "currying" mindset where the bigger or more stable argument comes first to the function. This is of course kind of vague, but most obvious example is maybe a setter; the thing being modified comes first and then comes the new value.

Applying the same idea to the functions we're discussing seem to suggest that `implements` should switch, since we're "modifying that type". One could maybe also say that we're modifying the interface, but I don't think that's how people generally think about it (even though it's kind of what's happening in the compiler internals.)

All this is to say that I kind of agree that this PR has the superior order, but it's a close call and I didn't want to suggest changing all the other code in the same PR. 

Maybe we can just merge this as-is and fix the ordering later? That will give us more time to think about it, without blocking people from trying out `derive`.

hellerve: I see merits in either ordering, and I don’t mind either. What would be interesting would be to enable people to derive multiple interfaces in one call:

```clojure
(derive [= zero str] T)
; vs
(derive T [= zero str])
```

Where, for me, the second ordering seems more natural; this of course is entirely subjective as well.

I’d argue that for now keeping concistency and not breaking existent behavior probably trumps any need to change, meaning that I should go ahead and switch the order for `derive` and we can revisit this another time. How does that sound?

eriksvedang: Deriving multiple ones sounds great. Could be a `*` macro also, of course (or just rest args?)

> How does that sound?

Perfect!

hellerve: While changing the order I realized that this would also change overrides to look like this:

```clojure
(derive = T eq) ; => derive = for T as eq
```

I don’t know whether I like that very much.

eriksvedang: ugh, no that's really confusing

TimDeve: `(derive T [= zero str])` Cool 🙂

scolsen: Cool, good discussion! Let's keep this order and change the order of `implements` instead. I can take that on.

hellerve: Okay, I changed the order back for now then :) I’ll resolve this discussion for now!

docs/Derive.md
```diff
@@ -0,0 +1,113 @@
...
+(derive Point zero)
+(derive Point =)
+
+; if you’d like to generate a different function name
+; pass it as a third argument. This is useful to avoid
+; name collisions
+(derive Point str my-str)
+```
+
+The code above wil provide implementations of `zero` and `=` for the type
```
eriksvedang, line 31: typo: "wil"

eriksvedang (commented):
Found a typo.

scolsen:
Thanks @hellerve sorry for all the back and forth

hellerve:
No worries, that’s how design discussion (should) go!

eriksvedang:
Is this ready to merge now? It looks great to me, but there was so much discussion that I might have missed something.

hellerve:
No objections from me, I think we are ready! 🚀

PR #1122, opened 2021-01-11
Respect use in dynamic lookup

### LINKED ISSUE #1121: Dynamics are not exported into the global space when module is `use`d
Dynamic symbols are not exported in the global space when their module is used.

Example:

```clojure
> (defmodule X (defndynamic x [] 1))
> (use X)
> x
Can't find symbol '"x"' at REPL:1:1.
> (x)
Can't find symbol '"x"' at REPL:1:2.
```

Cheers

hellerve:
This PR fixes #1121 by respecting `use` in dynamic lookup.

Cheers


eriksvedang:
This is great! Can we please add some tests around this? (since this kind of lookup has often been a very problematic to get right)

hellerve:
I added a test case! I hope that’s enough.

PR #1124, opened 2021-01-12
fix: Closure context

scolsen:
This PR does two things:

1. Adds `Semigroup` instances for `Env` and `Context` to offer a way to combine each of these types, respectively. Envs are combined using the left-biased union of the maps containing their bindings.
2. Updates our handling of closures to evaluate them under the *semigroup combination* of the context captured during their creation and the context under which they are evaluated, biasing toward the creation-context. This allows us to ensure closures have access to the latest global definitions, fixing #1101, while also ensuring that, in the case of conflicts, they prefer the local definitions captured during the creation context.

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -302,7 +302,7 @@ eval ctx xobj@(XObj o info ty) preference resolver =
                 (newCtx, evaledArgs) <- foldlM successiveEval (ctx, Right []) args
                 case evaledArgs of
                   Right okArgs -> do
-                    (_, res) <- apply c{contextHistory=contextHistory ctx} body params okArgs
+                    (_, res) <- apply (c {contextHistory = contextHistory ctx} <> ctx) body params okArgs
                     pure (newCtx, res)
                   Left err -> pure (newCtx, Left err)
         XObj (Lst [XObj Dynamic _ _, sym, XObj (Arr params) _ _, body]) i _ : args ->
--- a/src/Map.hs
+++ b/src/Map.hs
@@ -40,3 +40,6 @@ keys (Map m) = M.keys m
 
 map :: (a -> b) -> Map k a -> Map k b
 map f (Map m) = Map $ M.map f m
+
+union :: Ord k => Map k v -> Map k v -> Map k v
+union (Map m) (Map m') = (Map (M.union m m'))
--- a/src/Obj.hs
+++ b/src/Obj.hs
@@ -3,6 +3,7 @@
 
 module Obj where
 
+import Control.Applicative
 import Control.Monad.State
 import Data.Char
 import Data.Hashable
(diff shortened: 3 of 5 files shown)


scolsen:
@hellerve hopefully this unblocks you for `derive`! I used the example you included in the issue and it works after this change.

scolsen:
This fixes #1101 as is, but thinking about this a bit more, the semigroup for `Env` likely has to be a bit more nuanced.

We probably want to left-bias for *internal Envs* but right bias for other envs--that is, any `let` bindings in a closure context should win over global bindings of the same name, but any global bindings in the closure context should lose against global bindings of the same name in the latter evaluation/application context.

In other words, the semigroup for `Env` should right-bias global envs and type envs and left-bias internal envs.

eriksvedang:
> This fixes #1101 as is, but thinking about this a bit more, the semigroup for `Env` likely has to be a bit more nuanced.
> 
> We probably want to left-bias for _internal Envs_ but right bias for other envs--that is, any `let` bindings in a closure context should win over global bindings of the same name, but any global bindings in the closure context should lose against global bindings of the same name in the latter evaluation/application context.
> 
> In other words, the semigroup for `Env` should right-bias global envs and type envs and left-bias internal envs.

I haven't thought hard about this but given that you're correct it seems worth fixing before merge:ing, right?

src/Eval.hs
```diff
@@ -302,7 +302,7 @@ eval ctx xobj@(XObj o info ty) preference resolver =
                 (newCtx, evaledArgs) <- foldlM successiveEval (ctx, Right []) args
                 case evaledArgs of
                   Right okArgs -> do
-                    (_, res) <- apply c{contextHistory=contextHistory ctx} body params okArgs
+                    (_, res) <- apply (c {contextHistory = contextHistory ctx} <> ctx) body params okArgs
```
eriksvedang, line 305: So clean, I'm impressed!

scolsen:
> > This fixes #1101 as is, but thinking about this a bit more, the semigroup for `Env` likely has to be a bit more nuanced.
> > We probably want to left-bias for _internal Envs_ but right bias for other envs--that is, any `let` bindings in a closure context should win over global bindings of the same name, but any global bindings in the closure context should lose against global bindings of the same name in the latter evaluation/application context.
> > In other words, the semigroup for `Env` should right-bias global envs and type envs and left-bias internal envs.
> 
> I haven't thought hard about this but given that you're correct it seems worth fixing before merge:ing, right?

Yes, if I am right I think we should wait until this is updated! Let me run a couple checks to see if my thinking is right here, and I'll let you know

scolsen:
My assumption was right--made the fix and added some tests to check some basic closure behaviors:

```
./test/closure.carp
Test 'closures over global variables get the global variable's latest state.' passed
Test 'closures with arguments use argument values' passed
Test 'closures with let bindings prefer let bindings over global names' passed
Test 'dynamic closures can refer to global bindings delcared after the closure' passed
Test 'nested closures prefer closed-over bindings' passed
Results:
	|=====|
	Passed: 5	Failed: 0
```

scolsen:
@eriksvedang should be ready now assuming all tests pass (passed locally)

src/Eval.hs
scolsen: Thanks!

eriksvedang:
Really nice! I'm a little confused though, the tests seem to be mainly for statically compiled closures, no? (which is not what the patch addresses)

scolsen:
> Really nice! I'm a little confused though, the tests seem to be mainly for statically compiled closures, no? (which is not what the patch addresses)

Doh! Thanks for catching that. Looks like I got confused this morning. Added some commits to change it to dynamics

PR #1125, opened 2021-01-12
fix: replace qualified dynamic bindings

scolsen:
Previously, we couldn't replace qualified bindings in Envs. This change
updates envReplaceBindings to accept qualified paths and to replace the
binding denoted by the path, updating the module environments the binder
belongs to accordingly.

If the qualified path does not denote an existing module, it simply
returns the environment as is (just like the unqualified case).

### DIFF
--- a/src/Env.hs
+++ b/src/Env.hs
@@ -1,6 +1,7 @@
 module Env where
 
 import Data.List (foldl')
+import Data.Maybe (fromMaybe)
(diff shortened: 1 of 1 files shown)


hellerve (approved):
Looks great, thanks for looking into this so quickly!

eriksvedang:
Awesome!

PR #1126, opened 2021-01-12
fix: don't combine internal envs in closures

scolsen:
This fixes a bug whereby closures would mistakenly capture bindings of
other scopes because they'd combine internal environments.

In particular:

```
(defndynamic sum-with [f]
 (fn [] (reduce (fn [acc m] (+ acc (* m f))) 0 [1 2 3])))
((sum-with 10))
```

Would lead to `f` in the function passed to reduce being bound to the
function itself, instead of the expected `10` since reduce *also*
defines an `f` as an argument, and these would collide when combining
the internal envs.

We now keep internal envs separate, which prevents this issue. The
combination of type and global envs is retained, ensuring closures have
access to the latest global values.


eriksvedang:
Subtle, but makes sense!

PR #1127, opened 2021-01-12
DRAFT: Closure side effects

scolsen:
This PR depends on #1126.

This commit enables dynamic closures to update the global and type
environments, an example use case is using a closure to define a new
type:

```
> ((fn [] (eval '(deftype X []))))
> :i X
> X : Type
  Defined at line 1, column 25 in 'REPL'
  X : Module = {
        copy : (Fn [(Ref X q)] X)
            delete : (Fn [X] ())
                init : (Fn [] X)
                    prn : (Fn [(Ref X q)] String)
                        str : (Fn [(Ref X q)] String)
  }
    Defined at line 1, column 25 in 'REPL'
```

Previously, we ignored the context resulting from the evaluation of a
closure, preventing such environment manipulations. This gives users more
flexibility in dynamic code.

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -313,7 +313,9 @@ eval ctx xobj@(XObj o info ty) preference resolver =
                 (newCtx, evaledArgs) <- foldlM successiveEval (ctx, Right []) args
                 case evaledArgs of
                   Right okArgs -> do
-                    (_, res) <- apply (c {contextHistory = contextHistory ctx} <> ctx) body params okArgs
+                    let newGlobals = (contextGlobalEnv newCtx) <> (contextGlobalEnv c)
+                        newTypes = TypeEnv $ (getTypeEnv (contextTypeEnv newCtx)) <> (getTypeEnv (contextTypeEnv c))
+                    (_, res) <- apply (c {contextHistory = contextHistory ctx, contextGlobalEnv = newGlobals, contextTypeEnv = newTypes}) body params okArgs
                     pure (newCtx, res)
                   Left err -> pure (newCtx, Left err)
         XObj (Lst [XObj Dynamic _ _, sym, XObj (Arr params) _ _, body]) i _ : args ->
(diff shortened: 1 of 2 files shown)


scolsen:
@carp-lang/maintainers @hellerve rose this behavior in the chat. I opened this as a draft for now, since I'm not sure if we want to permit this sort of global effect in a closure--I personally think we should. I would make the dynamic language more powerful and I don't think it really breaks expectations since it's supposed to be a metaprogramming facility, and I don't see any major issues with it.

eriksvedang:
I like it!

scolsen:
Thanks! I think @hellerve had some reasons we might want to hold off on implementing this, though -- @carp-lang/maintainers for their opinions on this.

hellerve:
I think it’s weird to breaking the scoping rules in this one case—if I open up a new scope (like `fn`) I’d typically expect a new environment, and not being able to update the global scope. But maybe I’m just misguided.

I think in the long term, I’d rather we expose more of the context and environment to the dynamic scopes, so that stuff like this can be implemented in Carp directly. That, of course, is a lot of work, though.

I talked about this before in the context of reification of data types. This PR would fix the reification, but in a different way than I imagined when dabbling with it. I imagined a new primitive `hoist` that would hoist any call inside it into the global scope. This, too, is something we could implement in Carp if we had dynamic access to environment and context instead, of course.

scolsen:
> I think it’s weird to breaking the scoping rules in this one case—if I open up a new scope (like `fn`) I’d typically expect a new environment, and not being able to update the global scope. But maybe I’m just misguided.
> 
> I think in the long term, I’d rather we expose more of the context and environment to the dynamic scopes, so that stuff like this can be implemented in Carp directly. That, of course, is a lot of work, though.
> 
> I talked about this before in the context of reification of data types. This PR would fix the reification, but in a different way than I imagined when dabbling with it. I imagined a new primitive `hoist` that would hoist any call inside it into the global scope. This, too, is something we could implement in Carp if we had dynamic access to environment and context instead, of course.

Makes sense to me.

scolsen:
Closing for now--another issue related to USE modules resolved in this PR is resolved by @hellerve's PR #1150

PR #1130, opened 2021-01-16
feat: add string? and number?

hellerve:
This PR adds the `string?` and `number?` predicate commands.

Cheers


TimDeve:
Nice, I was literally looking for that the other day.

scolsen:
I'm ok w/ having this as a command but I think you could technically implement this in Carp with `type`.

hellerve:
`type` will only report static types. We’re talking about dynamic numbers here, whose type will always be `Dynamic`, unless they are literals. All dynamic symbols type dynamically. Moreover, `type` is currently a little weird, as I discovered when playing around with it just now:

```clojure
> (let [x 2] (type x))
I can’t find the symbol `x` at REPL:1:18.

Traceback:
(type x) at REPL:1:12.

> (defndynamic e [x] (type x))
> (e 1)
I can’t find the symbol `x` at REPL:1:26.

Traceback:
(type x) at REPL:1:20.
(e 1) at REPL:1:1.
```

scolsen:
> `type` will only report static types. We’re talking about dynamic numbers here, whose type will always be `Dynamic`, unless they are literals. All dynamic symbols type dynamically. Moreover, `type` is currently a little weird, as I discovered when playing around with it just now:
> 
> ```clojure
> > (let [x 2] (type x))
> I can’t find the symbol `x` at REPL:1:18.
> 
> Traceback:
> (type x) at REPL:1:12.
> 
> > (defndynamic e [x] (type x))
> > (e 1)
> I can’t find the symbol `x` at REPL:1:26.
> 
> Traceback:
> (type x) at REPL:1:20.
> (e 1) at REPL:1:1.
> ```

Ah, right! Thanks! I'll look into the `type` problems (looks like it doesn't look in internal envs)

scolsen (approved):

hellerve:
Hopefully at some point we will figure out how to avoid having to have different implementations for lookup for different primitives!

eriksvedang:
Maybe we could have a way of getting the dynamic (Obj) type also?

eriksvedang:
I'm fine with merging this as-is of course.

hellerve:
You mean like a symbol for each AST node?

eriksvedang:
> You mean like a symbol for each AST node?

Yes, something like that.

eriksvedang:
I'll merge this to not block any progress, but if you agree that the more general `dynamic-type` idea is good maybe we can switch over to that in the future?

hellerve:
That sounds reasonable to me!

PR #1131, opened 2021-01-16
fix: expand macros in closure bodies

hellerve:
This PR fixes a problem where previously, macros where not expanded inside dynamic closures.

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -300,11 +300,15 @@ eval ctx xobj@(XObj o info ty) preference resolver =
                           Just e = contextInternalEnv newCtx
                       pure $ Right (newCtx {contextInternalEnv = Just (envInsertAt e (SymPath [] n) binder)})
                     Left err -> pure $ Left err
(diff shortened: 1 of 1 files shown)


scolsen (approved):

scolsen:
Great!

eriksvedang:
Nice fix. I'm a little suspicious of it though, since we don't call `macroExpand` in any other place except in macros – so maybe this fix should actually be in the function `macroExpand`, where we *do* match on a bunch of different things? I might just be misguided, let me know in that case!

hellerve:
We actually do call `macroExpand for all `def` forms in `dynamicOrMacro` and for `defmodule` directly in `step`, i.e. on definition time. Since `fn` goes through neither definition form, it doesn’t get the same treatment. Does that make sense?

eriksvedang:
OK, thanks for the clarification. I'll look at it once more when not tired but seems legit!

PR #1132, opened 2021-01-16
refactor: make dynamic equality more permissive

hellerve:
This PR makes dynamic `=` more permissive, allowing equality on everything. This might seem "unsafe" to some of you, but is useful in dynamic contexts and macros: if I traverse a form, for instance, I might need to compare a list to a number and such things. They are not equal, and they never should be, but I need `=` to not throw an error in such cases.

Cheers

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -521,36 +521,12 @@ commandEq :: BinaryCommandCallback
 commandEq ctx a b =
   pure (ctx, Right (boolToXObj (cmp (a, b))))
   where
-    cmp (XObj (Num aTy aNum) _ _, XObj (Num bTy bNum) _ _)
-      | aTy == bTy =
-        aNum == bNum
-    cmp (XObj (Str sa) _ _, XObj (Str sb) _ _) = sa == sb
-    cmp (XObj (Chr ca) _ _, XObj (Chr cb) _ _) = ca == cb
     cmp (XObj (Sym sa _) _ _, XObj (Sym sb _) _ _) = sa == sb
-    cmp (XObj (Bol xa) _ _, XObj (Bol xb) _ _) = xa == xb
-    cmp (XObj Def _ _, XObj Def _ _) = True
-    cmp (XObj Do _ _, XObj Do _ _) = True
-    cmp (XObj Let _ _, XObj Let _ _) = True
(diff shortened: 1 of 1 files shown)


src/Commands.hs
```diff
@@ -519,47 +519,38 @@ commandMacroLog ctx msgs = do
...
-    cmp (XObj Macro _ _, XObj Macro _ _) = Right True
-    cmp (XObj Dynamic _ _, XObj Dynamic _ _) = Right True
-    cmp (XObj DefDynamic _ _, XObj DefDynamic _ _) = Right True
-    cmp (XObj The _ _, XObj The _ _) = Right True
-    cmp (XObj Ref _ _, XObj Ref _ _) = Right True
-    cmp (XObj Deref _ _, XObj Deref _ _) = Right True
-    cmp (XObj (Lst []) _ _, XObj (Lst []) _ _) = Right True
+        aNum == bNum
+    cmp (XObj (Str sa) _ _, XObj (Str sb) _ _) = sa == sb
+    cmp (XObj (Chr ca) _ _, XObj (Chr cb) _ _) = ca == cb
```
scolsen, line 528: Are all these explicit matches on XObjs needed? 

Couldn't this be simplified to:

```
cmp x x = obj x = obj x
```

Retaining the clauses for lists and arrays since they require length comparisons etc.

scolsen:
Makes a lot of sense to me. I don't think dynamic land needs to be safe at all.

jacereda:
So, this is basically making https://github.com/carp-lang/Carp/blob/9147df435ffd530206974cd05ca735711f6e6591/core/List.carp#L265 a simple `(= x '+)`, right? LGTM.

hellerve:
> So, this is basically making
> 
> https://github.com/carp-lang/Carp/blob/9147df435ffd530206974cd05ca735711f6e6591/core/List.carp#L265
> a simple `(= x '+)`, right? LGTM.

That’s correct!

src/Commands.hs
hellerve: Huh... Actually that seems to work fine. I also had to leave the case for symbols, since we don’t care about a symbol’s mode when checking for equality. That simplifies the function tremendously!

PR #1136, opened 2021-01-18
feat: add deprecation meta

hellerve:
This PR adds the `meta` form `deprecated`. It will be rendered in HTML (see the screenshot below) and in the REPL, like so:

```
> :i =>
=> : Macro
  Defined at line 27, column 3 in '/Users/veitheller/Documents/Code/Github/carp/carp/core/ControlMacros.carp'
  Deprecated: true
```

Bildschirmfoto 2021-01-18 um 13 46 44

Cheers

### DIFF
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -22,9 +22,11 @@
       (cons-last (thread-last-internal (all-but-last xs)) (last xs))
       (list (last xs) (thread-last-internal (all-but-last xs))))))
 
+(deprecated =>)
 (defmacro => [:rest forms]
   (thread-first-internal forms))
 
+(deprecated ==>)
 (defmacro ==> [:rest forms]
   (thread-last-internal forms))
 
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -22,11 +22,11 @@
       (cons-last (thread-last-internal (all-but-last xs)) (last xs))
       (list (last xs) (thread-last-internal (all-but-last xs))))))
 
-(deprecated =>)
+(deprecated => "deprecated in favor of `->`.")
 (defmacro => [:rest forms]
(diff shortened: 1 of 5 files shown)


eriksvedang (approved):

jacereda:
I would prefer "Deprecated" instead of "Deprecated: true", I guess that's also a problem now for hidden and friends.

jacereda (approved):

TimDeve:
I would make an explanation of the deprecation mandatory, personally.

hellerve:
> I would prefer "Deprecated" instead of "Deprecated: true", I guess that's also a problem now for hidden and friends.

I agree! I went the extra mile and fixed this for all the boolean values we work with.

hellerve:
> I would make an explanation of the deprecation mandatory, personally.

What would go in the explanation? What to use instead? Why it was deprecated?

TimDeve:
Yeah, I feel that a deprecated note without any guidance is not very useful to the end user.

eriksvedang:
Maybe any non-nil value means it's deprecated. Would enable this:

`Deprecated: Replaced by ->`

hellerve:
`deprecated` now receives an optional string argument that will be rendered if it is provided.

 
Bildschirmfoto 2021-01-19 um 09 45 58

This code was used to generated the above:

```clojure
(deprecated example)
(defn example [] 1)

(deprecated another-example "was deprecated in favor of nothing. Have fun!")
(defn another-example [] 1)
```

I think this is the best compromise.

eriksvedang:
Looks great, thanks for pleasing us all!

eriksvedang:
Huh, I thought I merged this (sorry for dragging my feet). Has conflicts now, will merge when fixed.

hellerve:
Fixed.

PR #1139, opened 2021-01-19
feat: make empty? an interface

hellerve:
This PR makes `empty?` an interface.

Cheers


scolsen (approved):
Nice!

eriksvedang:
That makes a lot of sense!

PR #1140, opened 2021-01-20
feat: a proper dynamic numeric tower

hellerve:
The following things were changed and/or added:
- `Dynamic.neg` was added
- `Dynamic.mod` was changed to work on float values
- `Dynamic.cxr` was changed to work with `0` instructions
- `Dynamic.=` was changed to ignore the type of the number
- `Dynamic.round` was added
- dynamic arithmetic was changed to respect the numeric type tower
- the instances of `Eq` and `Ord` for `Number` are no longer derived, so that they work across numeric types
- the instance of `Num` for `Number` was changed to work across numeric types
- `promoteNumber` was added as a type function to implement the numeric tower.

The numeric tower is as follows:

```
Byte -> Int -> Long -> Float -> Double
```

Cheers

### DIFF
--- a/core/Dynamic.carp
+++ b/core/Dynamic.carp
@@ -9,8 +9,16 @@
   (defndynamic dec [x]
     (- x 1))
 
+  (defndynamic neg [x]
+    (* -1 x))
+
   (defndynamic mod [x y]
-    (- x (* y (/ x y))))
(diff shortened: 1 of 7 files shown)


hellerve:
I’d be happy to change anything about this, adding or removing whatever.

eriksvedang (approved):
I like this a lot!

Issue #1115, opened 2021-01-06
RFC: Format Strings


hellerve:
As discussed in the Gitter chat, I recently implemented a very simple yet powerful version of format strings. This RFC is an effort to agree on a standard and push it into the standard library.

For fun and reference, here is the code in full:

```clojure
(hidden f-internal)
(defndynamic f-internal [s]
  (let [idx (String.index-of s \{)
        len (String.length s)]
    (if (= idx -1)
      [(list 'copy s)]
      (let [ss (String.suffix s (inc idx))
            endx (String.index-of ss \})]
        (if (= endx -1)
          (macro-error "error in format string: unclosed open bracket")
          (append [(list 'copy (String.prefix s idx)) (list 'str (parse (String.prefix ss endx)))]
            (f-internal (String.suffix ss (inc endx)))))))))

(doc f "formats a string using `str`.")
(defmacro f [s]
  (list 'String.concat (list 'ref (f-internal s))))
```

It expands into a single call to `String.concat`, with the string literal becoming copied chunks and the expressions evaluating in order and being stringified. Please note that this does allocate. I am not married to the name.

Here is how one would use it:

```clojure
(def x 1)
(def y "hi")

(IO.println &(f "this is x: {x} and this is the first letter of hi: {(head y)}"))
; => this is x: 1 and this is the first letter of hi: h
```

The ergonomics of this could be further improved by adding a reader macro, for instance using `f` in front of the string, like so:

```clojure
(def x 1)
(def y "hi")

(IO.println &f"this is x: {x} and this is the first letter of hi: {(head y)}")
; => this is x: 1 and this is the first letter of hi: h
```
Again, I’m not married to the glyph `f` here, but it seems like an obvious choice, and has prior art in Python f-strings, to which they are related.

I’d like to hear any and all comments, criticisms, or thoughts you might have!

Cheers

scolsen:
This is great! I think `f` is fine as a reader macro. If we define it as a normal macro as well, I don't love the name `f` simply because it's typically used for local bindings for functions when writing in a functional style (f, g, h) etc. which might make it harder to decipher cases in which a locally bound lambda `f` is being called vs. the `f` macro. Since it uses `str`, would `fstr` be acceptable?

But again, as a reader macro, I think `f` is fine since it's unambiguous when appended to a string literal.

> Please note that this does allocate.

We should probably mention this in the doc for the macro.

Awesome work!

hellerve:
Thanks for the kind words!

[...]

PR #1142, opened 2021-01-21
feat: add fstr

hellerve:
This PR deals with #1115, at least partially. As proposed by  @scolsen, I changed the name to `fstr`.

Cheers

### DIFF
--- a/core/Dynamic.carp
+++ b/core/Dynamic.carp
@@ -26,7 +26,10 @@
     (defndynamic suffix [s from]
       (String.slice s from (String.length s)))
 
-    (defndynamic tail [s ]
+    (defndynamic head [s]
+      (String.prefix s 1))
+
+    (defndynamic tail [s]
       (String.suffix s 1))
   )
 )
--- a/core/Format.carp
+++ b/core/Format.carp
@@ -42,18 +42,57 @@
 (defmacro fmt [s :rest args]
   (list 'copy (fmt-internal s args)))
 
+(hidden f-parse-expr-string)
+(defndynamic f-parse-expr-string [s r]
+  (if (= (String.length s) 0)
+    -1
+    (let [h (String.head s)]
+      (if (= h "\"")
+        (inc r)
+        (let [i (if (= h "\\") 2 1)]
+          (f-parse-expr-string
+            (String.suffix s i)
+            (+ r i)))))))
+
+(hidden f-parse-expr)
+(defndynamic f-parse-expr [s idx r]
+  (cond
+    (= idx 0) r
+    (= (String.length s) 0) -1
+
+    (let [h (String.head s)
+          t (String.tail s)]
+      (if (= h "\"")
+        (let [l (f-parse-expr-string t 0)]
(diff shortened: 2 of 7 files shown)


TimDeve:
Looking good. 

I like that we're holding off on the reader macro for now, maybe we should wait till we're closer to 1.0 so we don't end up with symbols overload ^^

hellerve:
Sounds good. We’ve been adding a lot of features lately, and I think it’s good to not add them in a way that interferes with other goals we might have.

I don’t think we have comprehensive docs on `fmt` or `fstr`? I could add them as part of this PR if that’s desired? If so, should they get their own document?

TimDeve:
I think it would be useful, especially outlining the fact that curly braces shouldn't be used inside other curly braces. How hard would it be to make the macro fail with a readable error if people open new curlies?

hellerve:
The problem is that you need to build a parser for that (because `{` could also show up in a string inside the fstring, for instance), and it gets way more complicated than what I built very quickly.

TimDeve:
Just think we should document the limitations just so people are not confused by the error messages:
```
鲤 (fstr "Hello {{@"a" 2}}")
fstr expected 1 arguments, but received 3.
```

hellerve:
> Just think we should document the limitations just so people are not confused by the error messages:
> 
> ```
> 鲤 (fstr "Hello {{@"a" 2}}")
> fstr expected 1 arguments, but received 3.
> ```

This is actually a Carp parser thing, since you’re closing the string with the second quote, so the thing gets parsed as `(fstr "Hello {{@" a " 2}}")`. As in any string you’ll have  to escape the quotes.

hellerve:
But, of course, the code is still problematic. I’ll add documentation about the limitations of `fstr`.

TimDeve:
Ah yes you're right. A better example:
```
鲤 (fstr "Hello {{1 1}}")
"command:parse" (line 1, column 5):
unexpected end of input
expecting "&", "~", "@", "'", "`", "%@", "%", "(", "$[", "[", "{", "-", digit, "0b", "0x", "#\"", "r\"", "\"", "\\", letter or "}" at dummy-file:0:0.
```

hellerve:
I added a parser for format expression that eliminates the problems pointed out by @TimDeve. To make it work, I also had to add the  following:

* `Dynamic./=`
* `Dynamic.String.head`
* And I had to teach the parser and emitter how to properly deal with escape codes in strings.

TimDeve:
I like the doc, should there be a link to it in the language guide?

hellerve:
I guess; maybe also for quasiquotation and `derive`?

TimDeve:
Should we have a "related page" section like on the [compiler manual page](https://github.com/carp-lang/Carp/blob/master/docs/Manual.md)?

hellerve:
Sounds good! I would avoid making the scope of this PR creep even more and instead add another PR that creates that section. How does that sounds?

eriksvedang:
Lots of good stuff in this PR now :)

PR #1146, opened 2021-01-23
fix: deal with literals in match

### LINKED ISSUE #1113: Literals inside `match` crash the compiler
When using literals in `match`, the compiler crashes. Consider this program:

```clojure
(defn test [m]
  (match m
     (Maybe.Just 0) (println* "hi")
     _ (println* "bye")))
```

This generates the following error:

```
carp: Can't create binder for non-symbol in 'case' variable match:XObj {xobjObj = Num Int 0, xobjInfo = Just (Info {infoLine = 3, infoColumn = 18, infoFile = "REPL", infoDelete = Set {unSet = fromList []}, infoIdentifier = 7}), xobjTy = Nothing}
  error, called at src/InitialTypes.hs:488:42 in CarpHask-0.4.2.0-In3bsPt14LZHMVFTWVtUXQ:InitialTypes
  ```
  
 Ideally, the expander should catch this error before it gets passed to `initialTypes`.
 
 Cheers

hellerve:
This PR fixes #1113 by adding an analysis into `expand` that checks whether anything other than lists and symbols are used in the left hand side of a `match`. The code is quite inelegant; if anyone has any ideas on how to make it better, I’d love to hear it!

I also added a test case for the new error message.

Cheers

### DIFF
--- a/src/Expand.hs
+++ b/src/Expand.hs
@@ -132,7 +132,7 @@ expand eval ctx xobj =
           | even (length rest) ->
             do
               (ctx', expandedExpr) <- expand eval ctx expr
-              (newCtx, expandedPairs) <- foldlM successiveExpandLR (ctx', Right []) (pairwise rest)
+              (newCtx, expandedPairs) <- foldlM successiveExpandLRMatch (ctx', Right []) (pairwise rest)
               pure
                 ( newCtx,
                   do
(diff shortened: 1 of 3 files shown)


eriksvedang (approved):
Very important fix!

PR #1148, opened 2021-01-23
feat: add dynamic-type command

hellerve:
This PR adds the `dynamic-type` command as discussed in #1130.

Cheers


eriksvedang:
Very nice! Any particular reason for not choosing quoted symbols over strings as the "token"? I can see the merit of both, just interested in your opinion @hellerve (and everybody else's)

hellerve:
That’s a good question, actually! I naturally gravitated towards strings, but maybe symbols make more sense, actually. The change to the command should be trivial, and I’m up for either.

eriksvedang:
@hellerve I think symbols will look slightly better, though of course the difference is really small. Unless someone complains loudly I'd prefer if we switch to that!

PR #1150, opened 2021-01-25
refactor: use Set for used modules

### LINKED ISSUE #1128: Refactor: Make `envUseModules` a Set instead of a list
`envUseModules` is a list of sympaths, but it doesn't make sense to have any duplicates--once one declares a module as `used` a second use does nothing.

It'd be convenient if this invariant was baked into the data structure we used for tracking used modules, as it would simplify our management of them, e.g. 25344e4 resolves a teeny bug that would not have even been possible if we used a data structure like `Set` instead of list

hellerve:
This PR fixes #1128 by  refactoring `Env` to use a `Set` for used modules.

Cheers


scolsen (approved):
Great!

PR #1151, opened 2021-01-25
fix: make set! work with dynamic args (thanks @hellerve)

### LINKED ISSUE #1144: Dynamic set! cannot reset arguments
Dynamic `set` is unable to reset a variable received as an argument to, for instance, a dynamic function.

Example:

```clojure
(defndynamic x [i]
  (while (/= i 10)
    (set! i (inc i))))
```

Calling `(x 0)` will loop forever. Putting `i` in a `let` instead of using an argument resolves the issue.

Cheers

scolsen:
Like `let` before it, we used to bind function arguments to their values
only, which wasn't accounted for in `set!` such that one could not
`set!` `i` in `defndynamic [i] ...`. To fix this for let bindings we
introduced a `LetDef` form for consistency with Def forms. This commit
renames `LetDef` to `LocalDef` and uses it as a value for function
arguments in addition to let bindings, ensuring `set!` works on function
arguments too. Big thanks to @hellerve for the suggestion!

Fixes #1144

### DIFF
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -168,7 +168,7 @@ toC toCMode (Binder meta root) = emitterSrc (execState (visit startingIndent roo
             (Interface _ _) -> dontVisit
             (Dict _) -> dontVisit
             (Fn _ _) -> dontVisit
-            LetDef -> dontVisit
+            LocalDef -> dontVisit
             (Match _) -> dontVisit
             With -> dontVisit
             MetaStub -> dontVisit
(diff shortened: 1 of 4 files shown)


hellerve (commented):
Looks great! 🎉 

Can we add a test for this? Maybe in `test/regression.carp`?

scolsen:
> Looks great! 🎉
> 
> Can we add a test for this? Maybe in `test/regression.carp`?

Good call! Added.

hellerve (approved):
Yay!

PR #1153, opened 2021-01-26
fix: release resources on break

### LINKED ISSUE #1052: `break` keyword not handled by memory management system
We use `break` for important optimizations in Core (where it's correct, by inspection). 

Ideally we would be able to rule out misuses like the following:

```clojure
(defn f []
  (for [n 0 5]
    (let-do [leak @"LEAK"]
      (break))))
```

hellerve:
This PR fixes #1052 by fixing up `break` statements with the context of which resources need to be released.

This is achieved by descending from any `let` and `while` statement into an enclosed `break` and adding any variables that need to be freed in that scope to its deleters.

Cheers

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE LambdaCase #-}
 
 module Concretize where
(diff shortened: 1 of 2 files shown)


src/Concretize.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
```
scolsen, line 1: Seems like no BangPatterns are used?

scolsen (approved):

src/Concretize.hs
hellerve: Good catch! I used this for debugging, but then didn’t need it in the final product.

src/Concretize.hs
```diff
@@ -1262,6 +1268,15 @@ manageMemory typeEnv globalEnv root =
                   Right (XObj (Lst (okF : okArgs)) i t)
         [] -> pure (Right xobj)
     visitList _ = error "Must visit list."
+    searchBreak :: Set.Set Deleter -> XObj -> XObj
```
eriksvedang, line 1271: I'm confused by this function's name, is there perhaps an alternative that fits it better? Or maybe I just need to be enlightened.

hellerve: `searchForInnerBreak` might be better?

hellerve: Or `fixupInnerBreak`?

eriksvedang: searchForInnerBreak is more clear in my mind

hellerve: Renamed!

PR #1156, opened 2021-01-27
fix: check calls of address and ref

### LINKED ISSUE #806: address shouldn’t allow literals
Currently `address` will accept literals, such as `(address 1)`. This will emit `&1`, which does not work in C.

I think what needs to happen is that `address` needs to be a primitive that needs to check whether the argument passed is a symbol.

I volunteer to work on this.

Cheers

hellerve:
This PR fixes #806 by checking the calls to `address` and `ref` statically as to whether they make sense.

Cheers


eriksvedang (approved):
Can you please remind me of why these couldn't be commands?
I trust that this is correct, but I just fail to understand why at the moment :)

hellerve:
I’m not sure why they’re not implemented that way and as special forms instead, but the special command that I’m removing just tried to verify what we’re now verifying in `expand` unsuccessfully. It wasn’t actually a special command in the regular sense.

I think making them primitives could work, but then we’d need special cases in the emitter.

eriksvedang:
@hellerve I see, yeah this is fine.

PR #1157, opened 2021-01-27
refactor: don’t print compilation info on implicit compile

### LINKED ISSUE #779: Hide compiler logging when calling functions from the repl
When building and running a project from the repl there's a little message that looks something like this:
```
Compiled to '....../Carp/out/Untitled' (executable)
```

When calling a function from the repl, it would be nice if this message was not displayed since it's unnecessarily noisy and makes the evaluation result harder to see.

hellerve:
This PR fixes #779 by hiding the compile information on an implicit compile (for instance, when typing `(Int.+ 1 2)` in the REPL) and letting it return `nil` instead of `0`, avoiding a print.

To this end, it also extends `build` by adding an optional boolean argument that specifies whether build information should be printed.

Cheers


eriksvedang:
Awesome, this will feel a lot nicer when working in the REPL!

PR #1158, opened 2021-01-28
refactor: Move evaluation errors into a separate module

scolsen:
Following the same scheme as PrimitiveError and TypeError, this adds a
separate module for evaluation errors, removing a bunch of inline
strings in Eval.hs

### DIFF
--- a/src/EvalError.hs
+++ b/src/EvalError.hs
@@ -1,7 +1,12 @@
--- | Module EvalError defines errors that may occur during evaluation
+-- | Module EvalError defines errors that may occur during evaluation.
 module EvalError
   ( EvaluationError (..),
     throwErr,
+    defnInvalidArgs,
+    loadInvalidArgs,
+    loadOnceInvalidArgs,
+    withInvalidArgs,
+    setInvalidArgs,
   )
 where
 
@@ -12,7 +17,11 @@ import TypeError
 import Util
 import Types
 
+--------------------------------------------------------------------------------
+-- Data (Evaluation Errors)
+
 -- | Errors that occur when evaluating forms.
+-- Mostly called from the evaluator.
 data EvaluationError
   = SymbolNotFound SymPath
   | PrivateBinding SymPath
@@ -22,7 +31,6 @@ data EvaluationError
   | -- Defn
     DefnContainsQualifiedArgs XObj
   | DefnIdentifierIsQualified XObj
-  | DefnContainsInvalidArgs XObj
   | DefnMalformed XObj
   | -- Def
     DefIdentifierIsQualified XObj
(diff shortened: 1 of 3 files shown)


src/EvalError.hs
```diff
@@ -0,0 +1,146 @@
...
+  | -- Address
+    AddressContainsNonSymbol XObj
+  | -- While
+    WhileContainsNonBool XObj
+  | -- Defmodule
+    ModuleRedefinition String
+  | DefmoduleContainsNonSymbol XObj
+  | DefmoduleNoArgs
+  | -- Load
+    LoadInvalidArgs [XObj]
```
hellerve, line 53: While it’s certainly more type safe this way, wouldn’ it increase reusability if we had a single `InvalidArgs` error?

hellerve (approved):

src/EvalError.hs
scolsen: Good point! I think that'd be more consistent with primitive error handling too, right?

I do sort of like having errors specialized to each type of form, but I don't have a strong preference either way, so I'm happy to change it. Less code is better code, so let's try it the way you suggested!

jacereda (approved):

eriksvedang (approved):
LGTM, can I merge?

hellerve:
I think the change discussed in the comment hasn’t been implemented yet?

src/EvalError.hs
eriksvedang: @hellerve Can you please explain the issue a little bit more, I'm not sure I follow.

eriksvedang:
@hellerve OK, then I'll wait!

Seems to be some conflicts now, too.

src/EvalError.hs
hellerve: At the moment, there is a `LoadInvalidArgs`, a `LoadOnceInvalidArgs`, a `SetInvalidArgs`, a `WithContainsInvalidArgs`, and a `DefnContainsInvalidArgs`. Since the error message is mostly the same, I would argue we should make a `InvalidArgs String XObj` instead, with the first argument being the name of the form that got invalid arguments (could also be an `XObj` with a `Sym`, of course).

Does that make sense?

eriksvedang:
@scolsen Seems like there's some little conflict here now. Let me know if you want to merge this before or after the big refactor PR!

scolsen:
> @scolsen Seems like there's some little conflict here now. Let me know if you want to merge this before or after the big refactor PR!

I think I'd rather merge the context PR #1170 first, since I think resolving conflicts in this PR will be slightly easier.

eriksvedang:
@scolsen Gotcha!

eriksvedang:
@scolsen Ping!

scolsen:
> @scolsen Ping!

Sorry for the wait! Just merged in master so this should be good to go.

PR #1160, opened 2021-01-28
fix: respect quotes in macro expand

hellerve:
This PR fixes an issue where quoted macros would be expanded still because the expander didn’t actually stop when it encountered a quote.

Cheers


scolsen (approved):
Good catch!

PR #1161, opened 2021-01-28
feat: make reseeding of Random at startup configurable

hellerve:
**Note**: *this PR depends on #1160.*

This PR makes reseeding the random number generator in `Random` at startup configurable.

It adds the following functions:
- `Random.startup-reseed` will enable reseeding at startup. **This is the default**.
- `Random.no-startup-reseed` will disable reseeding at startup.
- `Random.startup-reseed?` can be called to check whether `Random` was reseeded at startup.
- `Dynamic.Random.startup-reseed?` can be called to check whether `Random` will be reseeded at startup.

Test cases are included.

An open question remains whether we need two functions to enable and disable reseeding to make the intent clear, or whether one function with a boolean argument—e.g. `(Random.startup-reseed true)`—would be enough.

Cheers

### DIFF
--- a/core/Random.carp
+++ b/core/Random.carp
@@ -27,11 +27,29 @@
       (set! s (mod (+ (* s a) c) m))
       (/ s m)))
 
-  (private _)
   (hidden _)
   (doc _ "forces reseeding based on the current time at program start")
-  (def _ (do (seed) 0))
+  (def _ (do (seed) true))
+
+  (doc startup-reseed "enables reseeding the random number generator at startup.")
+  (defmacro startup-reseed []
+    (eval '(defmodule Random (def _ (do (seed) true)))))
+
+  (doc no-startup-reseed "disables reseeding the random number generator at startup.")
+  (defmacro no-startup-reseed []
+    (eval '(defmodule Random (def _ (do false)))))
+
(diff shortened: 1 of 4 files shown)


TimDeve:
Thanks for taking the time to look at this, I like `(Random.startup-reseed true/false)` personally. I also feel something like `Random.seed-at-startup` would read better but that's nitpicking.

hellerve:
I’m open to renaming it, but then I’d call it `reseed-at-startup`, since to me `seed-at-startup` reads like a setter for a fixed seed. Does that make sense?

TimDeve:
I see your point but I find `reseed` ambiguous what about `gen-seed-at-startup`?

hellerve:
`gen-seed-at-startup` works for me. How is `reseed` ambiguous to you (genuinely interested)?

hellerve:
Okay, I changed the API to reflect this discussion!

TimDeve:
> `gen-seed-at-startup` works for me. How is `reseed` ambiguous to you (genuinely interested)?

I feel that if you were reseeding you would need to provide the new seed but I don't really know what the standard name for these type of function is.

eriksvedang:
Cool solution!

PR #1163, opened 2021-01-31
feat: add Introspect.arguments

hellerve:
This PR adds `Introspect.arguments`, which largely does the same thing as `Introspect.arity`, but without calling `length` but instead returning the argument array as-is.

Because it’s thus a  generalization of `Introspect.arity` that function was refactored to use `Introspect.arguments`.

Cheers


scolsen (approved):
👏

PR #1165, opened 2021-01-31
fix: remove unused vars in carp_stdint.h

hellerve:
This PR removes unused variables from `carp_stdint.h` that prevented `-Wall` form working. As a side note, this line also makes `-Wall` fail:

https://github.com/carp-lang/Carp/blob/59d43abf74cbdba77914312e0d714d92267e1dfb/src/Emit.hs#L420

Because it turns out that this variable is only used when a member of the sumtype is used in the right hand side of match. Since this was not a quick fix, and not necessarily a terrible bug (it’s just an unused variable) I didn’t bother to fix it for now.

Cheers

### DIFF
--- a/core/carp_stdint.h
+++ b/core/carp_stdint.h
@@ -63,9 +63,7 @@ Uint8 Uint8_copy(Uint8* x) {
 }
 
 Array Uint8_from_MINUS_bytes(Array* a) {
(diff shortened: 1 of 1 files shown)


eriksvedang:
Nice!

PR #1167, opened 2021-02-01
feat: beautify type variables on unification failure

### LINKED ISSUE #558: Make type variable names beautful in error messages

hellerve:
This PR fixes #558 by using the scheme provided by @jacereda in #565, updated for the current state of Carp (thus I set him as coauthor on the commit).

Please note that this will not beautify lifetime variables, since they sadly don’t appear in the mappings provided by the unification failure, which the prettifier depends on.

Cheers


src/TypeError.hs
```diff
@@ -1,65 +1,23 @@
-module TypeError where
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
jacereda, line 1: What's the problem here?

scolsen: By default GHC will complain if you declare a typeclass instance for a type in a different module than the one in which the type itself is declared.

Personally, I think it's a good safeguard to have, but at the same time it's not a *huge* deal to break this rule--it does open up the possibility of accidentally declaring overlapping instances etc. 

Most of the time it's turned off to permit implementation swapping a la SML functors.

src/TypeErrorDef.hs
```diff
@@ -0,0 +1,55 @@
+module TypeErrorDef where
```
scolsen, line 1: I mentioned this in another comment, but I don't actually see much benefit to separating this out and ignoring orphan instances unless we truly make error handling generic. If our long term plan is to make the type error module's show instance a generic holder for other types, that works, but it feels needlessly complicated to me.

That said this has the major benefit of being significantly easier to read.

src/TypeError.hs
hellerve: As I understand it this option is only active for this module, is this correct?

src/TypeErrorDef.hs
hellerve: We need to pull it out because `AssignTypes` and `TypeError` both depend on it, and `TypeError` depends on `AssignTypes`, meaning we would have a circular dependency otherwise.

src/TypeError.hs
scolsen: I believe so

src/TypeErrorDef.hs
scolsen: Couldn't we just move `typeVariablesInOrderOfAppearance` into `Types.hs` instead? It seems like a more appropriate home for that function anyway since it only refers to itself and is a generic operation on `Ty` constructors, having nothing to do with the actual assignment of types. And then this would permit us to move `beautifyTys` into typeError or Types.hs as well.

hellerve: Sadly, we can’t move this into `Types` (which I think would be nicest) because it depends on `TypeError`. We also can’t move `beautifyTypeVariables` out of `AssignTypes`, since it depends on `assignTypes`. So we have  to move out `beautifyTy` and `typeVariablesInOrderOfAppearance`. This is a little confusing, but  maybe the best solution available to us.

It also solves the orphan instances problem.

Rectified in `64630d75`!

src/TypeError.hs
hellerve: Rectified in `64630d75`!

Issue #828, opened 2020-05-23
Dynamic Map (dictionary)


eriksvedang:
I think adding dynamic maps (using the existing `{ key value, key value }` syntax would be a nice and not overly hard thing to do. Obviously it should replicate as much as possible of the existing functionality (and names) of the static `Map` module.

One interesting design choice is whether to have something like the keywords of Clojure (e.g. `:stuff`) or if we're fine with just symbols, which would look something like this:

```clojure
{ 'stuff 10, 'more-stuff 20 }
```

What do you all think?

scolsen:
Neat! At first I kind of missed `:key` syntax in carp, but at this point I kind of like it's absence, actually. I think it makes things simpler. 

If we plan on introducing keyword arguments, I think using `:key` or something similar is probably necessary, but I think for this case just a symbol should be fine!

TimDeve:
Would you be able to mix types?

scolsen:
> Would you be able to mix types?

It should allow that, I think--dynamic code isn't type checked. Should likely work the same as lists and arrays in dynamic code:

```clojure
鲤 '[1 a b]
=> [1 a b]
鲤 [1 a b]
Can't find symbol 'a' at REPL:1:4.
鲤 
```

TimDeve:
Nice, might rewrite Hoquet as a macro if that lands.

eriksvedang:
Yeah I agree, it should be just as dynamic as dynamic arrays, functions, etc.

hellerve:
As we have them now, what is left to do here?

PR #1168, opened 2021-02-01
feat: add dynamic Map type

hellerve:
This PR implements a dynamic hashmap datatype in Carp, based on assoc lists and hashing. It uses the same algorithms as the static `Map` type under the hood.

This is a draft, and it’s pretty much in prototyping state. Getting, setting, and updating works, but there are a few unimplemented features and open questions.

To do:
- [x] Add `remove`, `kv-*` operations, and other functions to bring this to feature parity with the static map. This should be somewhat easy.
- [x] Document the module and all the helper functions that were introduced.
- [x] Write tests.

Open questions:
- Currently, just returning a map will result in having a massive assoc list printed to the REPL. Wrapping that in `Dynamic.Map.str` works, but it’s still somewhat unexpected. How should we make this work automatically?
- ~~Similarly, how do we make literals work? Presumably we can build `from-array`, and it shouldn’t be too hard and then “just” work when dynamic bindings are preferre, but that needs to be tested and verified.~~ Literals work.

Cheers

### DIFF
--- a/core/Dynamic.carp
+++ b/core/Dynamic.carp
@@ -32,6 +32,12 @@
   (defndynamic imod [x y]
     (- x (* y (/ x y))))
 
+  (doc even? "checks whether the number `n` is even.")
+  (defndynamic even? [n] (= 0 (mod n 2)))
+
+  (doc odd? "checks whether the number `n` is odd.")
+  (defndynamic odd? [n] (= 1 (mod n 2)))
+
   (defmodule Project
     (doc no-echo "Turn off debug printing in the compiler.")
     (defndynamic no-echo []
--- a/core/Dynamic.carp
+++ b/core/Dynamic.carp
@@ -20,13 +20,8 @@
         (set! m (- m b)))
       (if (< a 0) (neg m) m)))
 
-  (defmodule List
-    (doc pairs "makes a list of pairs out of a list `l`. If the number of
-elements is uneven, the trailing element will be discarded.")
-    (defndynamic pairs [l]
-      (if (< (length l) 2)
-        '()
-        (cons `(%(car l) %(cadr l)) (List.pairs (cddr l))))))
+  (defndynamic imod [x y]
+    (- x (* y (/ x y))))
 
   (defmodule Project
     (doc no-echo "Turn off debug printing in the compiler.")
--- a/core/Dynamic.carp
+++ b/core/Dynamic.carp
@@ -3,6 +3,13 @@
 (defmodule Dynamic
   ;; Functions for doing things at the REPL and during compile time.
 
+  (doc nil "is the value `nil`, i.e. the empty list.")
+  (defdynamic nil '())
+
+  (doc nil? "checks whether a value is nil, i.e. the empty list.")
+  (defndynamic nil? [value]
+    (= value nil))
+
   (defndynamic inc [x]
     (+ x 1))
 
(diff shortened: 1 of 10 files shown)


hellerve:
I still have the open question regarding stringification, but since the rest works, everything is documented, and there are tests, I guess this is ready for review!

hellerve:
I guess that `test/dynamic.carp` could profit from the new dynamic asserts in the test framework, but I’m moving this to a later PR, since there is already so much in here.

hellerve:
Oh, this also changes the API of `Map.from-array` from `(Fn [(Ref (Array (Pair a b)))] ())` to `(Fn [(Array (Pair a b))] ())`. This is to facilitate using the same API for dynamic maps.

eriksvedang:
> Oh, this also changes the API of `Map.from-array` from `(Fn [(Ref (Array (Pair a b)))] ())` to `(Fn [(Array (Pair a b))] ())`. This is to facilitate using the same API for dynamic maps.

Can you expand on this comment, I don't understand how the signature is related to dynamic code. Does the  implementation of `from-array` need to consume the pairs now, or will it still copy the data over and then delete the original array of pairs?

hellerve:
It is necessary because the reader rewrites map literals into `Map.from-array` forms. If there are ref forms in there, it will always end up being a static form, without the refs the heuristic is able to differentiate whether a map can be made dynamic or has to be made static.

The implementation of `from-array` doesn’t change, but the API needed to be the same for static and dynamic, and refs don’t exist dynamically.

eriksvedang:
> It is necessary because the reader rewrites map literals into `Map.from-array` forms. If there are ref forms in there, it will always end up being a static form, without the refs the heuristic is able to differentiate whether a map can be made dynamic or has to be made static.
> 
> The implementation of `from-array` doesn’t change, but the API needed to be the same for static and dynamic, and refs don’t exist dynamically.

Thanks a lot for the clarification, now I get it!

eriksvedang:
@hellerve Epic stuff! I wonder if we could add some small thing to the compiler to make it print nicely... I'll think a little about it. We could just have something in the outermost XObj that says that it's a DynamicMap, kind of – right?

eriksvedang:
Closes #828

eriksvedang:
> The implementation of from-array doesn’t change

It changes in the way that it adds an (implicit) delete at the end of the function though, right? I think that's a fine trade-off though.

hellerve:
> > The implementation of from-array doesn’t change
> 
> It changes in the way that it adds an (implicit) delete at the end of the function though, right? I think that's a fine trade-off though.

That is correct; I just meant that there was nothing in the function that actually needed to take ownership.

hellerve:
> @hellerve Epic stuff! I wonder if we could add some small thing to the compiler to make it print nicely... I'll think a little about it. We could just have something in the outermost XObj that says that it's a DynamicMap, kind of – right?

We could do that; we could make a tagging system for higher dynamic types, so that they can carry around information on where to look for their implementation of `str` and so on; possible similar to [SRFI 9](https://srfi.schemers.org/srfi-9/srfi-9.html), which I have some experience with.

This would blow up the scope of this PR, though, but I could start work on that while this PR is reviewed, and use `Map` as an experimental type to sanity check the implementation.

eriksvedang:
> > @hellerve Epic stuff! I wonder if we could add some small thing to the compiler to make it print nicely... I'll think a little about it. We could just have something in the outermost XObj that says that it's a DynamicMap, kind of – right?
> 
> 
> 
> We could do that; we could make a tagging system for higher dynamic types, so that they can carry around information on where to look for their implementation of `str` and so on; possible similar to [SRFI 9](https://srfi.schemers.org/srfi-9/srfi-9.html), which I have some experience with.

> This would blow up the scope of this PR, though, but I could start work on that while this PR is reviewed, and use `Map` as an experimental type to sanity check the implementation.

Sounds like a plan!

PR #1170, opened 2021-02-08
refactor: Context and Qualify

scolsen:
This PR is on the bigger side and does a few things:

1. Augments `Context.hs` with several utility functions for managing environments and other properties of a context--this enables us to use a more abstract/opaque API and move all context updating into a single place, making it harder to get things wrong.
2. Refactors the `Qualify` module to include types for representing *fully qualified xobjs and sympaths*. While I was working through the code base I realized that we were qualifying symbols across module boundaries (for example, the xobjs passed to `define` in `Primitives.hs` are fully qualified by `annotateWithinContext` in `Eval.hs`, which is also called in a few other contexts, meanwhile, several other primitives qualified symbols themselves, using contextPaths.) The old situation made code hard to grok and made it easy to mistakenly re-qualify a symbol that was already qualified or forget to qualify a symbol that still needed qualification. Using a type to represent the difference is a more defensive strategy which should make communication across modules more apparent and make it harder to do the wrong thing.
3. A couple of other miscellaneous refactors are included.

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -85,7 +85,7 @@
                                             (Dynamic.String.concat [x newline])))
                                 strings
                                 ())]
-    (eval (list 'meta-set! name "doc" (Dynamic.String.concat (list-to-array-internal separated []))))))
+    (list 'meta-set! name "doc" (Dynamic.String.concat (list-to-array-internal separated [])))))
 
 (doc print-doc "Print the documentation for a binding.")
 (defmacro print-doc [name]
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -124,7 +124,7 @@ primitiveColumn x@(XObj _ i _) ctx args =
 primitiveImplements :: BinaryPrimitiveCallback
 primitiveImplements call ctx x@(XObj (Sym interface@(SymPath _ _) _) _ _) (XObj (Sym (SymPath prefixes name) _) _ _) =
   do
-    (maybeInterface, maybeImpl) <- pure (lookupBinder interface tyEnv, lookupBinder (SymPath modules name) global)
+    (maybeInterface, maybeImpl) <- pure (lookupBinderInTypeEnv ctx interface, lookupBinderInGlobalEnv ctx (SymPath modules name))
     case (maybeInterface, maybeImpl) of
       (_, Nothing) ->
         if null modules
@@ -135,8 +135,6 @@ primitiveImplements call ctx x@(XObj (Sym interface@(SymPath _ _) _) _ _) (XObj 
       (Just interfaceBinder, Just implBinder) ->
         addToInterface interfaceBinder implBinder
   where
-    global = contextGlobalEnv ctx
-    tyEnv = getTypeEnv . contextTypeEnv $ ctx
     SymPath modules _ = consPath (contextPath ctx `union` prefixes) (SymPath [] name)
     warn :: IO ()
     warn = emitWarning (show (NonExistentInterfaceWarning x))
@@ -173,7 +171,7 @@ primitiveImplements _ ctx x _ =
   pure $ toEvalError ctx x (ArgumentTypeError "implements" "a symbol" "first" x)
 
 define :: Bool -> Context -> XObj -> IO Context
-define hidden ctx@(Context globalEnv _ typeEnv _ proj _ _ _) annXObj =
+define hidden ctx annXObj =
   pure (hideIt freshBinder)
     >>= \newBinder ->
       if isTypeDef annXObj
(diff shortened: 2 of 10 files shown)


src/Primitives.hs
```diff
@@ -430,48 +409,36 @@ primitiveMembers _ ctx target = do
...
-              newEnv = envInsertAt global fullPath updated
-           in (ctx {contextGlobalEnv = newEnv}, dynamicNil)
-        else evalError ctx ("`meta-set!` failed, I can't find the symbol `" ++ pretty target ++ "`") (xobjInfo target)
+      let updated = Meta.updateBinderMeta (Meta.stub fullPath) key value
+       in (insertInGlobalEnv ctx qpath updated, dynamicNil)
+--if null prefixes
+--  then
+--    let updated = Meta.updateBinderMeta (Meta.stub fullPath) key value
+--     in (insertInGlobalEnv ctx fullPath updated, dynamicNil)
+--  else evalError ctx ("`meta-set!` failed, I can't find the symbol `" ++ pretty target ++ "`") (xobjInfo target)
```
hellerve, line 441: Can we get rid of this?

hellerve (commented):
I can’t say I fully understood all the code and its consequences in this PR yet, but it looks very very good so far! Thanks for all of this 🎉

core/Macros.carp
```diff
@@ -85,7 +85,7 @@
                                             (Dynamic.String.concat [x newline])))
                                 strings
                                 ())]
-    (eval (list 'meta-set! name "doc" (Dynamic.String.concat (list-to-array-internal separated []))))))
```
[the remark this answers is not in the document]
scolsen: When I build and run this version of the compiler it does correctly add the docs when files are loaded in the REPL. Let me verify that it actually generates docs when generation is called.

src/Primitives.hs
scolsen: Actually, I likely need to restore it. This removes the error for calling `meta-set` against a qualified path so that `(doc Foo.bar "")` would work. Of course, if you actually try to evaluate this, you'll get the "can't insert into non-existing module" error when `Foo` does not yet exist. Likely we should just make that error an actual error since it's more general.

core/Macros.carp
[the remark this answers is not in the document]
hellerve: Okay, that’s great! Does the same apply to the other annotations (`sig`, `hidden`, etc)?

src/Primitives.hs
hellerve: Oh yeah, that makes sense of course!

src/Qualify.hs
```diff
@@ -11,147 +27,293 @@ import qualified Set
...
+          >>= \qualifiedCases -> pure $ Qualified $ XObj (Lst (matchExpr : qualifiedExpr : concat qualifiedCases)) i t
+  where
+    Just ii = i
+    lvl = envFunctionNestingLevel env
+    innerEnv :: Env
+    innerEnv = Env Map.empty (Just env) (Just ("case-env-" ++ show (infoIdentifier ii))) Set.empty InternalEnv lvl
+    qualifyCases :: (XObj, XObj) -> Either QualificationError [Qualified]
+    qualifyCases (l@(XObj (Lst (_ : xs)) _ _), r) =
+      let innerEnv' = foldl' foldVars innerEnv xs
+       in setFullyQualifiedSymbols typeEnv globalEnv env l
```
eriksvedang, line 257: Do notation must look better here, no?

src/Qualify.hs
```diff
@@ -11,147 +27,293 @@ import qualified Set
...
+      ++ "in the given context."
+
+--------------------------------------------------------------------------------
+-- Data
+
+-- | Denotes an XObj containing only symbols that *have been fully qualified*.
+--
+-- A fully qualified xobj **must not** be qualified further (e.g. using context
+-- paths).
+newtype Qualified = Qualified {runQualified :: XObj}
```
eriksvedang, line 55: To me, `run` signals that there's a monadic action going on.

eriksvedang (commented):
Didn't have time to fully read and understand it all (but most of it) and it looks very promising. The enhanced type safety will hopefully do wonders!

src/Qualify.hs
eriksvedang: (see line 258...)

src/Qualify.hs
scolsen: Good point! changed.

core/Macros.carp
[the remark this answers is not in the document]
scolsen: Yes, that's correct.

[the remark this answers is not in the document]
scolsen: I'll make the same change for the other meta-set forms in a follow-up PR.

src/Primitives.hs
scolsen: restored for now f61526d--I'll add proper error handling to `envInsertAt` in a folllow-up.

src/Qualify.hs
scolsen: I changed pretty much all these instances of `>>=` to do notation based on our discussions!

src/Qualify.hs
scolsen: 3cdc04e

scolsen:
@eriksvedang assuming the tests still pass, I think this should be ready to merge if everything looks good

eriksvedang:
@scolsen Thanks for your hard work – outstanding!

PR #1172, opened 2021-02-18
fix: don't qualify paths of registered types

scolsen:
The previous commit added qualification to paths of registered types.
However, while types defined in Carp support qualification (definition
within a module) thanks to special logic, this logic is not yet
implemented for registered types, leading to errors if their paths are
qualified in the type environment. This commit fixes the behavior and
adds a TODO to support this for registered types.

### DIFF
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -237,7 +237,8 @@ primitiveRegisterTypeWithoutFields :: Context -> String -> Maybe String -> IO (C
 primitiveRegisterTypeWithoutFields ctx t override = do
   let path = SymPath [] t
       typeDefinition = XObj (Lst [XObj (ExternalType override) Nothing Nothing, XObj (Sym path Symbol) Nothing Nothing]) Nothing (Just TypeTy)
-  pure (insertInTypeEnv ctx (qualifyPath ctx path) (toBinder typeDefinition), dynamicNil)
+  -- TODO: Support registering types in modules
+  pure (insertInTypeEnv ctx (markQualified path) (toBinder typeDefinition), dynamicNil)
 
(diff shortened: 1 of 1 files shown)


eriksvedang:
I'm not sure I get how the behavior changes with this PR?

scolsen:
> I'm not sure I get how the behavior changes with this PR?

Basically, before this PR if you did:

```
(defmodule Foo
  (defmodule Bar
    (register-type Baz)
  )
)
```

You'd encounter the "can't insert into non-existing env" error since it would try to put `Foo.Bar.Baz` in the type environment. Note that this *is* supported for `deftype`:

```
(defmodule Foo
  (defmodule Bar
    (deftype Baz)
  )
)
```
works fine, but that's only because there's special logic in the `deftype` primitive that makes it work -- that logic is currently missing for `register-type`, so it doesn't know how to insert something into the type environment.

Sorry, I should have explained this in the OP!

eriksvedang:
OK, so it's not crashing now (which is great!) but it does the wrong thing – is that a correct assesment? :)

eriksvedang:
The registered type just gets inserted at the toplevel, right?

scolsen:
> The registered type just gets inserted at the toplevel, right?

Yes, that's correct, which was the behavior for `deftype` as well until recently.

If it makes more sense, I can go ahead and implement the same module insertion behavior for register-type before we merge this--probably better than just adding a TODO 😅

eriksvedang:
@scolsen Yeah let's go with that!

eriksvedang:
I went ahead and merged it, seems better than crashing. But the todo should be addressed still, of course.

PR #1173, opened 2021-02-18
fix: don't `eval` meta-set! macros

scolsen:
Previously, we called eval on meta-set! macros like doc and sig. The
effect of this is to make such calls in the repl have immediate effect,
rather than print their macro expansions. Unfortunately, this behavior
has a dark and unintended consequence when evaluating files. Because
these macro calls force evaluation of their bodies, the meta-set! call
is executed immediately on macro expansion, *often in the wrong
context*, resulting in "stub" meta forms for undefined bindings to be
added to the wrong module scope. Consider the following (assuming the
old definitions of these functions using eval hold):

```
(defmodule Foo
  (defmodule Bar
    (doc baz "foo")
  )
)
```

One would expect this code to result in one binding/definition being
created, `Foo.Bar.baz`. However, this code will actually produce two
bindings `Foo.baz` and the expected `Foo.Bar.baz`. This is due to the
way we evaluate macros. Macro expansion will happen in two passes, once
for all forms in `defmodule Foo`, and a second time for all forms in
`defmodule Bar`. When `doc` contains an `eval` call, the meta-set is
forced during macro expansion cycle 1, in `defmodule Foo`, resulting
in an additional, unanticipated binding being added to Foo.
Contrarily, if we remove the call to `eval`, the first expansion
during evaluation of `defmodule Foo` simply expands the call into
meta-set, which is then evaluated normally during evaluation of
`defmodule Bar`.

The primary trade-off entailed by this change is that calls like `doc
foo` will no longer be immediately evaluated in the repl. The utility of
these macros seems more aligned to their use in files, so I don't think
it's a change many will gripe about. Users can also wrap such calls in
eval themselves to evaluate the macro expansion in the repl when needed
(or even in files, too).

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -95,7 +95,7 @@
 
 (doc sig "Annotate a binding with the desired signature.")
 (defmacro sig [name signature]
-  (eval (list 'meta-set! name "sig" signature)))
+  (list 'meta-set! name "sig" signature))
 
 (doc print-sig "Print the annotated signature for a binding.")
 (defmacro print-sig [name]
@@ -103,11 +103,11 @@
 
 (doc hide "Mark a binding as hidden, this will make it not print with the 'info' command.")
 (defmacro hidden [name]
-  (eval (list 'meta-set! name "hidden" true)))
+  (list 'meta-set! name "hidden" true))
 
 (doc private "Mark a binding as private, this will make it inaccessible from other modules.")
 (defmacro private [name]
-  (eval (list 'meta-set! name "private" true)))
+  (list 'meta-set! name "private" true))
 
 (doc todo "sets the todo property for a binding.")
 (defmacro todo [name value]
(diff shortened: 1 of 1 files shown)


eriksvedang:
Oh wow, neat!!! Something broke in the tests though.

scolsen:
Interesting! Definitely wasn't expecting tests to break, I'll look into it.

scolsen:
> Oh wow, neat!!! Something broke in the tests though.

hah, turns out this bug was hiding an error in the test!

The error occurs because `gojira : (Ref a b) -> a` tries to implement `monster : a -> String`

It worked before thanks to the bug--gojira has a signature, which meant before an extra binding `gojira : a` was added in addition to `gojira : &String -> String` (thanks to the behavior described in the OP). While the actual definition of `gojira` doesn't satisfy `monster` `gojira : a` can unify with anything (since it's just a type var) so it was accepting that function as an implementation!

This highlights another issue we haven't addressed yet, namely that `sig` currently has no actual effect on the typing of the binding it's called on -- it should, which would also give `gojira` the correct typing. (right now it's typed more generically thanks to its implementation, which uses `@` and infers a generic ref because of `@`'s typing).

eriksvedang:
> This highlights another issue we haven't addressed yet, namely that sig currently has no actual effect on the typing of the binding it's called on

Ah, right that's weird. It is checked once you try do define whatever you gave the signature, but until then it's not. Seems like a great little thing to correct. But maybe we can just work around it in the test to make the PR merge:able now? (you decide)

eriksvedang:
Should we just adjust this test for now?

eriksvedang:
Is this PR still relevant @scolsen ?

scolsen:
> Is this PR still relevant @scolsen ?

It's definitely still relevant, but the right approach to fixing this might be a bit different after all the recent changes. I'll revisit this and re-open when there's a good fix.

PR #1174, opened 2021-02-25
feat: Adds `defn-` and `def-` macros

TimDeve:
Adding these macros as a shortand for declaring a def or defn and making them `hidden` and `private`, useful to keep things internal to a module. Similar to Clojure's [`defn-`](https://clojuredocs.org/clojure.core/defn-).

### DIFF
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -101,7 +101,7 @@
 (defmacro print-sig [name]
   (eval (list 'macro-log (list 'meta name "sig"))))
 
-(doc hide "Mark a binding as hidden, this will make it not print with the 'info' command.")
+(doc hidden "Mark a binding as hidden, this will make it not print with the 'info' command.")
 (defmacro hidden [name]
   (eval (list 'meta-set! name "hidden" true)))
 
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -18,6 +18,9 @@
         (list-to-array-internal (cdr xs) (append acc (array (car xs))))))
 
 (defmodule Dynamic
+  (defndynamic quoted [x]
+    (list 'quote x))
+
   (defndynamic /= [a b] (not (= a b)))
 
   (defndynamic caar [pair] (car (car pair)))
@@ -71,10 +74,7 @@
     (= (dynamic-type s) 'array))
 
   (defndynamic number? [s]
-    (List.in? (dynamic-type s) '(int long double float byte)))
-
-)
-
+    (List.in? (dynamic-type s) '(int long double float byte))))
 
 (meta-set! doc "doc" "Set documentation for a binding.")
 (defmacro doc [name :rest strings]
@@ -132,13 +132,6 @@
 (defmacro deprecated [name :rest explanation]
   (let [v (if (= (length explanation) 0) true (car explanation))]
     (eval (list 'meta-set! name "deprecated" v))))
-
-(defmodule Dynamic
-
-  (defndynamic quoted [x]
-    (list 'quote x))
-
-)
 
 (defndynamic cond-internal [xs]
   (if (= (length xs) 0)
--- a/core/Macros.carp
+++ b/core/Macros.carp
@@ -133,6 +133,20 @@
   (let [v (if (= (length explanation) 0) true (car explanation))]
     (eval (list 'meta-set! name "deprecated" v))))
 
+(doc defn- "Declares a function while marking it as private and hidden.")
+(defmacro defn- [name args form]
+  (do
+   (eval (list 'private name))
+   (eval (list 'hidden name))
+   (list 'defn name args form)))
+
+(doc def- "Declares a variable while marking it as private and hidden.")
+(defmacro def- [name value]
+  (do
+   (eval (list 'private name))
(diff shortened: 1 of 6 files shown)


TimDeve:
Thinking of adding some example of using `hidden` & `private` in the module section of the [Language Guide](https://github.com/carp-lang/Carp/blob/master/docs/LanguageGuide.md) and introduce these two macros at the same time. Does that sound approriate?

scolsen:
> Thinking of adding some example of using `hidden` & `private` in the module section of the [Language Guide](https://github.com/carp-lang/Carp/blob/master/docs/LanguageGuide.md) and introduce these two macros at the same time. Does that sound approriate?

Sounds great! We should add some tests for these too.

TimDeve:
Yeah I was thinking about the tests yesterday. Would you expect some "error message" tests?

hellerve:
> Yeah I was thinking about the tests yesterday. Would you expect some "error message" tests?

I’d say so! We can use the `test-for-errors` setup.

core/Macros.carp
```diff
@@ -133,12 +133,19 @@
   (let [v (if (= (length explanation) 0) true (car explanation))]
     (eval (list 'meta-set! name "deprecated" v))))
 
-(defmodule Dynamic
-
-  (defndynamic quoted [x]
```
hellerve, line 138: Where did this go?

TimDeve:
> I’d say so! We can use the test-for-errors setup.

Added the tests a few minutes before your message. Doing docs next.

TimDeve:
Took me weeks to find the time to make this simple PR but I think it's finally ready for review 😄

docs/LanguageGuide.md
```diff
@@ -227,6 +156,88 @@ These can only be used at the REPL and during macro evaluation. Here's a subset
 
 To see all functions available in the `Dynamic` module, enter `(info Dynamic)` at the REPL.
 
+
+### Structs
+Any structure type defined in Carp has an init method that can be used to create a new instance. It must be called with all the arguments in the order they are defined.
```
hellerve, line 161: Could we wrap `init` in backticks here?

Also, maybe "in the order in which they are defined" or "in definition order" might be better?

docs/LanguageGuide.md
```diff
@@ -227,6 +156,88 @@ These can only be used at the REPL and during macro evaluation. Here's a subset
...
+```
+
+**Data:**
+```clojure
+(deftype (Either a b)
+  (Left [a])
+  (Right [b]))
+```
+
+A Variant can be created with the same syntax as call expression:
```
hellerve, line 200: `A variant can be created through a function call`?

docs/LanguageGuide.md
```diff
@@ -227,6 +156,88 @@ These can only be used at the REPL and during macro evaluation. Here's a subset
...
+    (Either.Right b) b))
+
+(with MyEnum
+  ;; You can give a generic "otherwise" statement as well
+  (match myenum
+    (Kind1) (logic1)
+    _ (logic-other)))
+```
+
+Note that match works with *values* (not references) takes ownership over the value being matched on. If you instead want to match on a reference, you can use `match-ref`:
```
hellerve, line 231: I think backticks around `match` would be good here, and that there is an `and` missing after the parentheses.

hellerve (commented):
I realized about halfway through that the docs I was annotating weren’t new but rearranged. My comments still hold, I think 😆

hellerve:
I think it would be good to use these forms in stdlib in a follow-up PR, since this could remove a lot of code and test the forms at the same time.

TimDeve:
The follow up PR sounds like a good idea. About the comments you've added I was thinking of making a refractor PR for the language guide, the style is not very consistent and some of it seems straight up wrong with the changes in the language since it was written, so I could put your suggested changes it that.

hellerve:
> The follow up PR sounds like a good idea. About the comments you've added I was thinking of making a refractor PR for the language guide, the style is not very consistent and some of it seems straight up wrong with the changes in the language since it was written, so I could put your suggested changes it that.

Sounds like a good plan! I’ll be happy to help and/or review :)

TimDeve:
Once/if this is merged I'll open a draft PR with your suggested changes and we can comment on it on what we add to that.

eriksvedang:
Looks good! The changes to the language guide is a bit hard to follow though, has anything been deleted or have things just moved around?

TimDeve:
Sorry the merged diff looks really messy, it's easier to follow if you go commit by commit. I moved the module and interface section after the struct section first, and then I added a section about `private` and `hidden` in the module section.

eriksvedang:
@TimDeve Alright, thanks I'll do that!

eriksvedang (approved):
@hellerve had some comments that you can choose to address, apart from that I think we're good to go!

eriksvedang:
@TimDeve Maybe you already discussed that and this should just go in as-is now?

TimDeve:
Yeah the comment where on existing docs so we're going to address them in a different PR.

eriksvedang:
Great work!

PR #1177, opened 2021-03-01
feat: add dynamic types

hellerve:
This PR adds dynamic types to Carp. Yes, it’s a big feature.

The most important part here is `deftype-dynamic`, which takes a name and a list of slots, or members. Calling it will generate an `init` function, getters and setters, and `str`, similar to what would happen if it were a static type. It also generates a predicate to check whether a value is actually such a tagged type under `<type>.<type>?`.

Under the hood, we are using a tagged array, similar to approaches like [SRFI-9](https://srfi.schemers.org/srfi-9/srfi-9.html).

This PR also makes three other changes:
- it changes dynamic map to be a `deftype-dynamic`.
- it changes `str` to call `<Type>.str` if we’re handling a tagged type and `prim-str` (which is the compiler one) if we’re given anything else.
- it changes the top-level command evaluator to call `str` on any array that is being returned. This means that we get a nice string if we return a tagged type at the top-level REPL.

Code documentation is included, but I will add a section to the docs; I wanted to hold off on that until we’ve decided that this is a good idea.

Cheers

### DIFF
--- a/core/Map.carp
+++ b/core/Map.carp
@@ -632,7 +632,7 @@
     (doc init "creates a dynamic pair, i.e. a list with two elements.")
     (defndynamic init [k v] (list k v)))
 
-  (deftype-dynamic Map (l))
+  (deftype-dynamic Map [buckets])
 
   (defmodule Map
     (hidden dflt-len)
@@ -662,39 +662,39 @@
     (doc resize "resizes a dynamic map to size `s`.")
     (defndynamic resize [m s]
       (let-do [n (n-times s list)]
-        (Map.init (Map.resize- n (Map.l m)))))
+        (Map.init (Map.resize- n (Map.buckets m)))))
 
     (doc grow "grows a dynamic map.")
-    (defndynamic grow [m] (Map.resize m (* (length (Map.l m)) 2)))
+    (defndynamic grow [m] (Map.resize m (* (length (Map.buckets m)) 2)))
     (doc shrink "shrinks dynamic map.")
     (defndynamic shrink [m]
-      (let [new-size (/ (length (Map.l m)) 2)]
+      (let [new-size (/ (length (Map.buckets m)) 2)]
         (if (< new-size dflt-len)
           m
           (Map.resize m new-size))))
 
     (doc contains? "checks whether the dynamic map `m` contains the key `k`.")
     (defndynamic contains? [m k]
-      (let [l (Map.l m)
+      (let [l (Map.buckets m)
             idx (Dynamic.imod (hash k) (length l))]
         (List.in? k (map car (List.nth l idx)))))
 
     (doc put "adds a value `v` under the key `k` into the map. If `k` already
 exists in the map, it is updated.")
     (defndynamic put [m k v]
-      (if (> (/ (* (Map.len m) 100) (length (Map.l m))) Map.min-load)
+      (if (> (/ (* (Map.len m) 100) (length (Map.buckets m))) Map.min-load)
         (Map.put (Map.grow m) k v)
         (if (Map.contains? m k)
           (Map.update m k (fn [_] v))
-          (let [idx (Dynamic.imod (hash k) (length (Map.l m)))]
-            (Map.init (List.update-nth (Map.l m) idx (fn [l] (cons (list k v) l))))))))
+          (let [idx (Dynamic.imod (hash k) (length (Map.buckets m)))]
+            (Map.init (List.update-nth (Map.buckets m) idx (fn [l] (cons (list k v) l))))))))
 
     (doc get-with-default "gets the value under the key `k`. If `k` doesn’t
 exist in the map, the default value is returned.")
     (defndynamic get-with-default [m k default-value]
       (if (Map.contains? m k)
-        (let [idx (Dynamic.imod (hash k) (length (Map.l m)))
-              l (List.nth (Map.l m) idx)]
+        (let [idx (Dynamic.imod (hash k) (length (Map.buckets m)))
+              l (List.nth (Map.buckets m) idx)]
           (cadr (List.find l (fn [pair] (= (car pair) k)))))
           default-value))
 
@@ -707,8 +707,8 @@ map, `nil` is returned.")
 `k` doesn’t exist in the map, it is returned unchanged.")
     (defndynamic update [m k f]
       (if (Map.contains? m k)
-        (let [idx (Dynamic.imod (hash k) (length (Map.l m)))]
-          (Map.init (List.update-nth (Map.l m) idx
+        (let [idx (Dynamic.imod (hash k) (length (Map.buckets m)))]
+          (Map.init (List.update-nth (Map.buckets m) idx
             (curry
                 map
                 (fn [pair] (if (= (car pair) k) (list k (f (cadr pair))) pair))))))
@@ -718,8 +718,8 @@ map, `nil` is returned.")
 function `f`. If `k` doesn’t exist in the map, it is set to `(f v)`.")
     (defndynamic update-with-default [m k f v]
       (if (Map.contains? m k)
-        (let [idx (Dynamic.imod (hash k) (length (Map.l m)))]
-          (Map.init (List.update-nth (Map.l m) idx
+        (let [idx (Dynamic.imod (hash k) (length (Map.buckets m)))]
+          (Map.init (List.update-nth (Map.buckets m) idx
             (curry
                 map
                 (fn [pair] (if (= (car pair) k) (list k (f (cadr pair))) pair))))))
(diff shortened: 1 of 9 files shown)


src/Eval.hs
```diff
@@ -625,6 +625,8 @@ executeCommand ctx@(Context env _ _ _ _ _ _ _) xobj =
         executeCommand newCtx (withBuildAndRun (XObj (Lst []) (Just dummyInfo) Nothing))
       Left (HasStaticCall _ _) ->
         callFromRepl newCtx xobj
+      Right a@(XObj (Arr _) _ _) ->
+        executeCommand newCtx (XObj (Lst [XObj (Sym (SymPath [] "str") Symbol) (Just dummyInfo) Nothing, XObj (Lst [XObj (Sym (SymPath [] "quote") Symbol) (Just dummyInfo) Nothing, a]) (Just dummyInfo) Nothing]) (Just dummyInfo) Nothing)
```
eriksvedang, line 629: What does this impressive line of code do? :D

core/Map.carp
```diff
@@ -632,6 +632,8 @@
     (doc init "creates a dynamic pair, i.e. a list with two elements.")
     (defndynamic init [k v] (list k v)))
 
+  (deftype-dynamic Map (l))
```
eriksvedang, line 635: This call to `(l)` is a bit cryptic to me...

eriksvedang (commented):
Can I please ask for some clarifications in the code <3

core/Map.carp
hellerve: This is not a call, it just says that `Map` has one member, `l`, which  is the underlying list structure!

src/Eval.hs
hellerve: Ah yes, that requires cleaning up. Basically what it says is that if we return an array from the top-level evaluator, we call it again with `(str '<array)`. This should be safe, since arrays are "just" data, I think?! What it enables is that `str` then decides whether this is a dynamic type it knows how to stringify or not!

core/Map.carp
hellerve: Maybe we should change the syntax to an array?

eriksvedang: Ah, I see. Yeah that would be better I think. Also give the member a more descriptive name?

src/Eval.hs
eriksvedang: I see, thanks for the explanation! The array can contain function calls, of course – but I gather that that should be fine?

hellerve: If the array is returned, those shouldn’t be  evaluated anymore anyway. Quoting the array should lead to the same effect I think.

core/Map.carp
hellerve: I changed it so that it doesn’t matter, but I suggest we make array the convention.

I also changed the member name to `buckets`.

hellerve:
Do we want this feature? If so, I’d be happy to start writing some documentation!

eriksvedang:
> Do we want this feature? If so, I’d be happy to start writing some documentation!

I think we do! Can you explain the tagged array approach just a litte though, I'm not familiar with it.

hellerve:
Basically, this code generates an array that holds all the members and a tag that identifies it.

```clojure
; we define a type
(deftype-dynamic Point [x y])

; we initialize it
(Point.init 1 2) ; => [carp-dynamic-type-tag-Point, 1, 2]
```

The predicate function then, for instance, checks whether what is passed in is a non-empty array, and whether the first element is the tag it expects (it should probably also check the size, now that I type this out). The getter function indexes into the array, incrementing the slot index by one, to ignore the type tag. Such work all the functions here.

As such, this approach is very similar to using a tagged union, but with arrays. It’s also not dissimilar to what a lot of compiled-but-dynamic languages like Common Lisp do (usually using bytes and bits instead of arrays, but the idea is the same).

Does that make it clear?

hellerve:
The size check is implemented!

hellerve:
I also added a feature now that enable `dynamic-type` to identify dynamic deftypes!

hellerve:
The windows build failure seems to be a network hiccup unrelated to this PR.

eriksvedang:
@hellerve Thanks for the explanation, that made it a lot clearer! Could there be an advantage using a kind of special XObj to carry around such dynamic values, instead of piggybacking on array? This is just a question, I'm not sure it's worth it or that I actually think it's better – I just feel like this is a good point to have the discussion. Having to not make changes to the compiler innards is definitely nice with the current approach, it just feels a little... "different" to how all other dynamic values are treated. Maybe if the compiler had a slightly more extensible dynamic type system we could use it for everything and not just maps/structs..?

hellerve:
That seems useful, but I’d say the less we can push into the compiler and the more we can do in Carp, the better.

Yes, it creates a two-class system, with primitive types being a little different to user land types. I’d argue  that that makes sense and might even be a feature; but obviously I’m in favor of all this, or I wouldn’t have  implemented it.

If anyone else wants to voice their opinion, I’d love to hear it!

eriksvedang:
Thinking about it some more and I like the divide less and less unfortunately (it's super impressive though)

I think it's problematic to "randomly" (from the user's perspective) have some central dynamic types that are supported and some that the compiler doesn't know anything at all about. The compiler should be aware of things like structs and dictionaries existing in dynamic land, to be able to give proper error messages and do the correct thing when encountering those data structures. Right now it's more or less left in the dust.

It's also confusing for any language contributor, who will expect these things to all be on par. My suggestion is that we try to rethink this a bit to make it more unified and not create such a sharp divide between different seemingly similar dynamic types.

hellerve:
Huh, that’s a bummer. As a Lisp person, this is exactly the kind of power I expect or even demand to be there. I understand the other concerns, but I don’t know how much they are not better served by better tooling instead of constraining what is possible in the language.

For me, the most important thing is to keep the compiler as small as possible. If there is a choice between implementing something in Carp or in the compiler, I will definitely want it in Carp. This way, if the compiler ever needs to change drastically, which seems more and more likely (it’s becoming the bottleneck for most of the things I push on), less work needs to be done.

I understand that this is an important decision, and I’d like to talk about it more in-depth, because for me the underlying discussion—not this particular feature—is an important driver for me. If we have diverging opinions here, which is totally okay and reasonable—and I do want to emphasize that I understand your position—, I will have to reconsider my time investment in Carp. I understand that this sounds like a threat, but I totally don’t mean it to be! I just spend considerable amounts of time with and around Carp, and if my philosophy doesn’t align with where the project is going, then I can’t really justify that to myself!

hellerve:
I want to add that my position being so hardline should not influence your judgement: I think having a coherent vision is important for the project going forward. If ours don’t happen to align, that’s fine! I don’t want to force anything by saying what I said, but I did want to make my position clear and start the important, deeper discussion that happened to be sparked by an ultimately rather silly feature.

scolsen:
I'm personally OK with implementing things this way in dynamic carp. To me, one of the big benefits to having the dynamic language is the capability to do cool, uber-powerful stuff like this. imo, it's totally cool for the only types recognized in dynamicland to be `list` and `array` and everything else to be structurally and dynamically defined in the language itself. I've also personally found myself wanting this feature instead of using plain old lists for Pairs in some cases.

I do appreciate the thin line here though, I agree that it can lower the predictability of the language, but I think it's fine to have it as a *capability*. I do think we get into some tricky territory if we start doing things automatically or on-behalf of the user, but exposing it as an option the user can use if they want seems 100% ok with me.

One thing that does feel a little confusing though, is that while reading:

```
(defdynamic p (Pair.init 1 2))
```

Is clear, the other points of use become less so:

```
(Pair.x p) ; => 1
(Pair.set-x p 4) ; => (Pair 4 2)
```

While the type checker will make sure it's right, as a reader I find it kind of easy to forget/miss the fact that `p` is a dynamic pair and not a static pair when reading this code. I do wonder if it's worthwhile introducing some kind of naming convention for this reason. Languages like ATS use symbols to differentiate between boxed and unboxed types, I wonder if we can use a macro to do something similar here, kind of like we do for quasiquote? i.e. references to dynamic types would need to be prefixed with a `*` or something?  idk, kind of noisy, I suppose the user can just adopt whatever convention they want to help clarify point of use.

hellerve:
> While the type checker will make sure it's right, as a reader I find it kind of easy to forget/miss the fact that p is a dynamic pair and not a static pair when reading this code. I do wonder if it's worthwhile introducing some kind of naming convention for this reason.

That’s a very fair point. I’m not quite sure what the convention or even the enforced guarantee should be, but I think we should start documenting naming conventions and so on either way. I know, for instance, that I’ve used earmuffs to denote globals in the past (i.e. `*global-var*`), which is the standard in some Lisps (Clojure, Common Lisp), until I realized that the preference of Erik was `ALL-CAPS` and changed most (?) of my names for globals. This is especially important for newcomers I believe, since some uniformity here goes a long way towards grokking difficult bits of code.

scolsen:
Yeah, definitely! I think Carp is in an interesting position as it can either take influence from lisp or from c. I'm not too picky so long as we're consistent and not extreme.

hellerve:
To summarize another point that was made in discussing this: to integrate this better with the compiler, one could leverage the type annotation on `XObj`s, but this would likely require a huge rewrite, since none of the commands or primitives currently check the type (usually they only check the constructor contained in the `XObj`), and the emitter would likely be confused if, say, a numeric type was annotated with a different type.

PR #1178, opened 2021-03-02
feat: Add support for emitting literal C

scolsen:
This PR introduces two new features that allow users to add arbitrary C code to the compiler's output:

## `Unsafe.emit-c : (Fn [String] C)`

Emits literal C code inline. One can use this for certain interop scenarios in which some C code *must* take e.g. a string literal as an argument. For example, the c11 `static_assert` macro *only* accepts a C string literal in the second argument position. A call to `(static-assert 0 (Unsafe.emit-c "\"foo\""))` emits the proper code `static_assert(0, "foo")`.

### Why do we need it?
Two existing solutions, just using a Carp string literal and using a deftemplate (or `inline-c`, which uses deftemplate) don't work for different reasons:

1. Carp's strings are managed, so they tend to be assigned to a generated variable name, such as `_8` for this purpose. Unfortunately, that means calls to `(static-assert 0 "foo")` in Carp are emitted as `static_assert(0, _8)` which C treats as invalid since the macro only accepts a string literal according to the standard.
2. Deftemplates aren't flexible enough to emit a literal inline on demand. Furthermore, they always generate both a declaration and definition. They are also replaced with the symbols they are bound to. Thus `(deftemplate foo "" "\"foo\"")` and `(static-assert foo)` is emitted as `static_assert(0, foo)`; invalid for the same reason as 1.

## `Unsafe.preproc` : (Fn [C] Unit)

Adds some literal C code (a value of `C` type in Carp) to a set of "preprocessor" strings to be appended to the compiler's output after `include` directives but *before* any other emitted code. This function allows users to add arbitrary C code to the code emitted by the compiler and ensure whatever code they add is available before any Carp code is called. It's best illustrated with an example:

```clojure
(Unsafe.preproc (Unsafe.emit-c "#define FOO 0"))                                                                                                                                                                                                                                
(Unsafe.preproc (Unsafe.emit-c "void foo() { printf(\"%d\\n\", 1); }"))                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                
(register FOO Int)                                                                                                                                                                                                                                                              
(register foo (Fn [] ()))                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                
(defn main []                                                                                                                                                                                                                                                                   
  (do (foo)                                                                                                                                                                                                                                                                     
      (IO.println &(fmt "%d" FOO))))
```

When run, this code will first print `1` and then print `0`. The code that's emitted for this carp looks as follows:

```c
#include <assert.h>                                                                                                                                                                                                                                                             
#include <stddef.h>                                                                                                                                                                                                                                                             
#include <stdlib.h>                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
// .. a bunch of other includes                                                                                                                                                                                                                              
#include <signal.h>                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                
#define FOO 0                                                                                                                                                                                                                                                                   
void foo() { printf("%d\n", 1); }                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                
//Types:                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
// Depth 3                                                                                                                                                                                                                                                                      
typedef struct {                                                                                                                                                                                                                                                                
    union {                                                                                                                                                                                                                                                                     
    struct {                                                                                                                                                                                                                                                                    
        Long member0;                                                                                                                                                                                                                                                           
    } Just;                                                                                                                                                                                                                                                                     
    // Nothing                                                                                                                                                                                                                                                                  
    char __dummy;                                                                                                                                                                                                                                                               
    } u;                                                                                                                                                                                                                                                                        
    char _tag;                                                                                                                                                                                                                                                                  
} Maybe__Long;                                                                                                                                                                                                                                                                  
#define Maybe__Long_Just_tag 0                                                                                                                                                                                                                                                  
#define Maybe__Long_Nothing_tag 1
```

### Why do we need it?

I was trying to find a nice way of adding compiler pragmas from Carp. Currently there's no great way to do this primarily because of the deftemplate limitations described above. If this PR is merged, this gives us a straightforward way to do this. It also would enable us to replace many h helper files with `preproc` calls. e.g. once could just call `(Unsafe.preproc (Unsafe.emit-c "typedef char *String"))` in `core.carp`.

Since preproc only accepts a `C` value as an argument, the only way to use it is in with combination with `Unsafe.emit-c` which should make it clear to users that it's a risk call and give one pause before using it too heavily.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -71,7 +71,8 @@ defaultProject =
       projectForceReload = False,
       projectPkgConfigFlags = [],
       projectCModules = [],
-      projectLoadStack = []
+      projectLoadStack = [],
+      projectPreproc = []
     }
 
 -- | Starting point of the application.
--- a/core/core.h
+++ b/core/core.h
@@ -15,6 +15,7 @@ typedef char *Pattern;
 typedef int64_t Long;
 typedef uint32_t Char;
 typedef char CChar;
+typedef void *c_code;
 
 #if defined NDEBUG
 #define CHK_INDEX(i, n)
--- a/docs/CInterop.md
+++ b/docs/CInterop.md
@@ -12,6 +12,8 @@ This is an extension of what is covered in the [Language Guide](./LanguageGuide.
   - [`deftemplate`](#deftemplate)
     - [`Basic example`](#basic-example)
     - [`Generics`](#generics)
+  - [`emit-c`](#unsafe-emit-c)
+  - [`preproc`](#unsafe-preproc)
 - [Callbacks](#callbacks)
 
 
@@ -40,7 +42,7 @@ generated on the C side. Here are some examples:
 
 ; Generic signature
 (sig print-first-and-add (Fn [(Ref (Array a)) b b] b))
-(defn print-first-and-add [arr x y] 
+(defn print-first-and-add [arr x y]
  (do
    (println* (Array.unsafe-first arr))
    (+ x y)))
@@ -52,10 +54,10 @@ generated on the C side. Here are some examples:
 ; => print_MINUS_first_MINUS_and_MINUS_add__String_Long
 ```
 
-Looking at the examples should be clear enough but let's break it down:  
+Looking at the examples should be clear enough but let's break it down:
 Carp will replace illegal characters in C with a string representation of them
-(`- => _MINUS_`, `? => _QMARK_`, etc...)  
-If in modules it will prefix the identifier with the modules name.  
+(`- => _MINUS_`, `? => _QMARK_`, etc...)
+If in modules it will prefix the identifier with the modules name.
 When the arguments to a function are generic it will suffix the types to the
 identifiers, the identifiers are not able to be generated until it is used. If
 a function is potentially generic but you don't want it to be you can add a
@@ -274,19 +276,19 @@ We can instead define the previous example like so:
 (print @"Print this!")
 ```
 
-Let's break down what's going on here:  
+Let's break down what's going on here:
 The **first** argument to `deftemplate` is the name we'll use to refer to the
-function.  
+function.
 The **second** is a type signature and is identical to the one found
-in our previous `register` call.  
+in our previous `register` call.
 The **third** is our function declaration, it'll be injected at the top of the
-generated C file.  
+generated C file.
 The **last** argument represent the function definition.
 
-Two more things to look at:  
+Two more things to look at:
 `$NAME` is a variable that will be derived from the name you've given the
 function plus any module it's defined in, so no need to worry about name
-clashes with other `print` functions in other modules.  
+clashes with other `print` functions in other modules.
 `$DECL` will be replaced with the declaration passed as a third argument when
 the function is defined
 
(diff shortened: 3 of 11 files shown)


eriksvedang (commented):
Just had a quick look, but initial impression is good. Need to try a few things I have in mind.

One small question is if defines should go before includes in the emitted file?

hellerve:
I think if it’s used for `define`s that might be helpful. If it’s used for inlining functions as demonstrated above, then putting it below the `include`s might be more helpful so they can use the `include`d functions?

scolsen:
> Just had a quick look, but initial impression is good. Need to try a few things I have in mind.
> 
> One small question is if defines should go before includes in the emitted file?

> I think if it’s used for `define`s that might be helpful. If it’s used for inlining functions as demonstrated above, then putting it below the `include`s might be more helpful so they can use the `include`d functions?

At the moment it will accept totally arbitrary C code and emits in call-order. I think placing it after Carp includes is OK, if one needs to use some `preproc` def to conditionally emit an include, one could just do:

```
(preproc (emit-c "#define FOO 0))
(preproc (emit-c "#ifdef FOO 
                    #include <assert.h>
                  #endif"))
```

Which would emit:

```
#define FOO 0                                                                                                                                                                   
#ifdef FOO                                                                                                                                                                      
  #include<assert.h>                                                                                                                             
#endif
```

after all the unconditional inclusions given by `(system-include)` and `( relative-include)` calls

(the actual tabbing in the output is awkward since it emits tabs and newlines as is, like we do for doc strings, but one could use `String.concat` for nicer output)

hellerve:
Interested thought: can this be used to supersede `system-include` and `relative-include`?

scolsen:
> Interested thought: can this be used to supersede `system-include` and `relative-include`?

Yes, both of those functions should be definable as macros now (quick sketchy def):

```clojure
(defmacro system-include [incl]
  (list 'Unsafe.preproc (list 'Unsafe.emit-c (String.concat ["#include < " incl ">"])))
```

In fact, I've defined the following in another branch:

```clojure
(defmodule Unsafe                                                                                                                                                                                                                
  (defmodule C                                                                                                                                                                                                                   
    (defndynamic emit-c-line [append-strings args]                                                                                                                                                                               
      (let [strs (map (fn [x] (String.concat [x " "])) (map str args))                                                                                                                                                           
            arr (collect-into strs array)                                                                                                                                                                                        
            code (String.concat (append append-strings arr))]                                                                                                                                                                    
      (list 'Unsafe.preproc (list 'Unsafe.emit-c code))))                                                                                                                                                                        
                                                                                                                                                                                                                                 
    (doc pragma                                                                                                                                                                                                                  
       "Emits a #pragma compiler directive in Carp's c output.")                                                                                                                                                                 
    (defmacro pragma [:rest args]                                                                                                                                                                                                
      (eval (Unsafe.C.emit-c-line ["#pragma "] args)))                                                                                                                                                                           
                                                                                                                                                                                                                                 
    (doc define                                                                                                                                                                                                                  
      "Emits a #define compiler directive in Carp's c output.")                                                                                                                                                                  
    (defmacro define [name value]                                                                                                                                                                                                
      (Unsafe.C.emit-c-line ["#define "] [name value]))                                                                                                                                                                          
                                                                                                                                                                                                                                 
    (doc undef                                                                                                                                                                                                                   
      "Emits a #undef compiler directive in Carp's c output.")                                                                                                                                                                   
    (defmacro undef [name]                                                                                                                                                                                                       
      (Unsafe.C.emit-c-line ["#undef "] [name]))                                                                                                                                                                                 
                                                                                                                                                                                                                                 
    (defndynamic if- [if-pre name then else]                                                                                                                                                                                     
      (do (eval (Unsafe.C.emit-c-line [if-pre] [name]))                                                                                                                                                                          
          (eval (Unsafe.C.emit-c-line ["  "] [then]))                                                                                                                                                                            
          (if (not (empty? else))                                                                                                                                                                                                
              (do (eval (Unsafe.C.emit-c-line ["#else"] []))                                                                                                                                                                     
                  (eval (Unsafe.C.emit-c-line ["  "] else)))                                                                                                                                                                     
              ())                                                                                                                                                                                                                
          (eval (Unsafe.C.emit-c-line ["#endif"] []))))                                                                                                                                                                          
                                                                                                                                                                                                                                 
    (doc ifpre                                                                                                                                                                                                                   
      "Emits a #if compiler directive in Carp's c output.")                                                                                                                                                                      
    (defmacro ifpre [name then :rest else]                                                                                                                                                                                       
      (Unsafe.C.if- "#if " name then else))                                                                                                                                                                                      
                                                                                                                                                                                                                                 
    (doc ifdef                                                                                                                                                                                                                   
      "Emits a #ifdef compiler directive in Carp's c output.")                                                                                                                                                                   
    (defmacro ifdef [name then :rest else]                                                                                                                                                                                       
      (Unsafe.C.if- "#ifdef " name then else))                                                                                                                                                                                   
                                                                                                                                                                                                                                 
    (doc ifndef                                                                                                                                                                                                                  
      "Emits a #ifndef compiler directive in Carp's c output.")                                                                                                                                                                  
    (defmacro ifndef [name then :rest else]                                                                                                                                                                                      
      (Unsafe.C.if- "#ifndef " name then else))                                                                                                                                                                                  
                                                                                                                                                                                                                                 
    (doc warning                                                                                                                                                                                                                 
      "Emits a #warning compiler directive in Carp's c output.")                                                                                                                                                                 
    (defmacro warning [message]                                                                                                                                                                                                  
      (eval (Unsafe.C.emit-c-line ["#warning "] [message])))                                                                                                                                                                     
                                                                                                                                                                                                                                 
    (doc error                                                                                                                                                                                                                   
      "Emits a #error compiler directive in Carp's c output.")                                                                                                                                                                   
    (defmacro error [message]                                                                                                                                                                                                    
    (eval (Unsafe.C.emit-c-line ["#error "] [message])))
```

hellerve:
Nice! That’s pretty awesome!

scolsen:
Yeah, I hope it'll be useful! Perhaps more useful still, you can now do:

```clojure
(Unsafe.C.ifdef __STDC_NO_COMPLEX__                                                                                                                                             
  "#error Complex is not supported by this implementation of the C standard library."                                                                                           
  "#include <complex.h>")
```

Instead of having to write it in a C header file.

eriksvedang:
I think an important thing to consider is that while it's technically possible to reimplement **most** features like this, there's a lot of value and UX benefits in the current (rigid) way of doing things. 

The constraints and solidity we get from the compiler's way of doing things is often more important than saving some lines of Haskell code, imo.

scolsen:
> I think an important thing to consider is that while it's technically possible to reimplement **most** features like this, there's a lot of value and UX benefits in the current (rigid) way of doing things.
> 
> The constraints and solidity we get from the compiler's way of doing things is often more important than saving some lines of Haskell code, imo.

Yeah, I definitely see the value in that too. I think the compiler includes should stick around--it also gives us the flexibility to remove these features independently of one another.

eriksvedang:
Those C emitting functions look great though! I just want to point out that there is a point where we're better served by solidifying the language rules in the compiler binary.

scolsen:
Thanks! It's nice to have options. Even for defining things like functions, a separate header file is still the way to go if you have several definitions imo, but this is useful and nice to have if all you need is one or two provisional functions.

eriksvedang (approved):

PR #1180, opened 2021-03-07
feat: add 'when' functions for Maybe and Result

scolsen:
These functions enable users to execute some side-effecting function
(one that returns unit and take no arguments) based on the contents of a
Maybe or Result.

`when-success`: Executes a side-effect when given a `Result.Success`
`when-error`: Executes a side-effect when given a `Result.Error`
`when-just`: Executes a side-effect when given a `Maybe.Just`
`when-nothing`: Executes a side-effect when given a `Maybe.Nothing`

e.g.:

```clojure
(def suc (the (Result Int Int) (Result.Success 0)))
(def err (the (Result Int Int) (Result.Error 0)))

(when-success &(fn [] (IO.println "success!") suc))
=> "success!"
(when-success &(fn [] (IO.println "success!") err))
=>
```

### DIFF
--- a/core/Control.carp
+++ b/core/Control.carp
@@ -18,4 +18,71 @@
         (set! result (~f result)))
       result))
 
+  (doc when-success 
+    "Executes a side effect, `f`, when `result` is `Success`ful."
+    "```"
+    "(def suc (the (Result Int Int) (Result.Success 0)))"
+    "(def err (the (Result Int Int) (Result.Error 0)))"
+    ""
+    "(when-success &(fn [] (IO.println \"success!\")) suc)"
+    "=> success!"
+    "(when-success &(fn [] (IO.println \"success!\")) err)"
+    "=> "
+    "```")
+  (sig when-success (Fn [&(Fn [] ()) (Result a b)] ()))
+  (defn when-success [f result]
+    (match result
+      (Result.Success _) (~f)
+      _ ()))
+
+  (doc when-error 
+    "Executes a side effect, `f`, when `result` is `Error`oneus."
+    "```"
+    "(def suc (the (Result Int Int) (Result.Success 0)))"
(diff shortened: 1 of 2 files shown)


hellerve:
It sounds like `when-success` and  `when-error` are specializations of `Result.and-then` and `Result.or-else`. Is that correct?

scolsen:
> It sounds like `when-success` and `when-error` are specializations of `Result.and-then` and `Result.or-else`. Is that correct?

Yes, these functions are structurally speaking relatives of `map`, `and-then`, and `or-else`—the key difference is that the function passed does not take the value in `Success` as an argument and that the return value is always taken to `Unit` instead of remaining a `Result`. Basically, it either runs some (potentially totally unrelated to the Result value) side-effect or does nothing.

All of these functions have the shape:

```
(match thing
   (matcher _) (~f)
   _ ())
```

scolsen:
here's an example of how you might use these functions totally out of context:

```clojure
(let [color &(Grimoire.Color.new 24 126)]
        (do (Control.when-error 
              &(fn [] (ignore (addstr @"error!")))
              @color)
            (Control.when-success
              &(fn [] (Grimoire.Color.with-color
                &(Result.unsafe-from-success @color)
                (ignore (Grimoire.addstrln @"color!"))))
              @color)))
```

scolsen:
Actually, one could make `when` an interface using the type signature described above:

```clojure
(definterface when (Fn [(f a) (Fn [] ())] ()))
```

But this actually isn't desirable for the result case, since one would have to choose to implement either `when-success` or `when-error` as the interface (since their maximally generic typings would be the same)

hellerve:
That all makes sense to me!

scolsen:
Either or sounds good to me! I put it in Control initially since these functions can help with control-flow, but organizing them by their types makes equal sense to me.

PR #1182, opened 2021-03-08
feat: add macros for emitting C compiler directives

scolsen:
This suite of macros uses the `Unsafe.emit-c` and `Unsafe.preproc`
functions to provide macros for emitting common C compiler directives,
such as #ifdef, #define, #pragma and others.


eriksvedang:
🎉

PR #1184, opened 2021-03-09
fix: allow dynamic closures to mutate the global env

### LINKED ISSUE #1181: Dynamic `set!` when called inside a closure only mutate state inside the closure
When calling `set!` on a Dynamic global variable defined using `defdynamic` from inside a closure the variable seems mutated only within the scope of closure.
Example:
```clojure
(defdynamic counter 0)

(defndynamic inc-counter [x]
  (do (set! counter (inc counter))
      (macro-log (str "x: " x " counter: " counter))))

(defndynamic return-fn []
 (fn [x] (inc-counter x)))

(defmacro my-macro []
  (do
    (inc-counter 0)
    (map (return-fn) '(1 2 3))))

(my-macro)
```
```
carp -b main.carp
```

Expected output:
```
x: 0 counter: 1
x: 1 counter: 2
x: 2 counter: 3
x: 3 counter: 4
```
Actual output:
```
x: 0 counter: 1
x: 1 counter: 2
x: 2 counter: 2
x: 3 counter: 2
```

scolsen:
Previously, closures could not change the global environment. As such,
calls to commands like `set!` within a closure did not update global
variables as one might expect.

This commit fixes the issue by doing two things:

- Preserve any updates to the global environment that occur when
  evaluating a closure.
- Handle contexts correctly when evaluating commands.

The latter point requires a bit more explanation. At some point, we
stopped threading through the context produced after evaluating a
command's arguments, which had the unfortunate result that some commands
did not get access to the latest environment values.

Both of these fixes together fix issue 1181.

## Example 

```
(defdynamic foo 0)                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                
(def z 0)                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                
(defndynamic inc-counter [x]                                                                                                                                                                                                                                                    
  (do (set! foo (inc foo))                                                                                                                                                                                                                                                      
      (macro-log (str "x: " x " counter: " foo))))                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                
(defndynamic return-fn []                                                                                                                                                                                                                                                       
 (fn [x] (inc-counter x)))                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                
(defmacro my-macro []                                                                                                                                                                                                                                                           
  (do                                                                                                                                                                                                                                                                           
    (inc-counter 0)                                                                                                                                                                                                                                                             
    (map (return-fn) '(1 2 3))))                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                
(my-macro)
```
**Old Behavior**
```
carp -b example.carp
=> x: 0 counter: 1
   x: 1 counter: 2
   x: 2 counter: 2
   x: 3 counter: 2
```

**New Behavior**

```
carp -b example.carp
=> x: 0 counter: 1
   x: 1 counter: 2
   x: 2 counter: 3
   x: 3 counter: 4
```

fixes #1181

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -242,8 +242,8 @@ eval ctx xobj@(XObj o info ty) preference resolver =
                     let newGlobals = (contextGlobalEnv newCtx) <> (contextGlobalEnv c)
                         newTypes = TypeEnv $ (getTypeEnv (contextTypeEnv newCtx)) <> (getTypeEnv (contextTypeEnv c))
                         updater = replaceHistory' (contextHistory ctx) . replaceGlobalEnv' newGlobals . replaceTypeEnv' newTypes
-                    (_, res) <- apply (updater c) body params okArgs
-                    pure (newCtx, res)
+                    (ctx', res) <- apply (updater c) body params okArgs
+                    pure (replaceGlobalEnv newCtx (contextGlobalEnv ctx'), res)
                   Left err -> pure (newCtx, Left err)
         XObj (Lst [XObj Dynamic _ _, sym, XObj (Arr params) _ _, body]) i _ : args ->
           case checkArity (getName sym) params args of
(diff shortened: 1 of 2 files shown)


eriksvedang:
Great!

PR #1186, opened 2021-03-15
refactor: macro tests

hellerve:
Note: this PR depends on #1185 to avoid clashes.

This PR refactors the macro tests to use `assert-dynamic-equal` to remove a bunch of code.

It also fixes `cxr` and `IO.exit`, which were broken (they’re now used in the test, so we should notice if that comes up again).

Cheers

### DIFF
--- a/core/IO.carp
+++ b/core/IO.carp
@@ -21,7 +21,7 @@
   (doc read-file "returns the contents of a file passed as argument as a string.")
   (register read-file (Fn [&String] String))
   (doc exit "exit the current program with a return code.")
-  (register exit (Fn [Int] a))
+  (register exit (Fn [Int] ()) "exit")
   (register EOF Char)
   (doc EOF "the End-Of-File character as a literal.")
   (doc fopen "opens a file by name using a mode (one or multiple of [r]ead, [w]rite, and [a]ppend), returns a file pointer. Consider using the function open-file instead.")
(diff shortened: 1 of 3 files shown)


core/IO.carp
```diff
@@ -21,7 +21,7 @@
   (doc read-file "returns the contents of a file passed as argument as a string.")
   (register read-file (Fn [&String] String))
   (doc exit "exit the current program with a return code.")
-  (register exit (Fn [Int] a))
+  (register exit (Fn [Int] ()) "exit")
```
eriksvedang, line 24: I think this was supposed to be `a` for a reason (to emulate "bottom"?)

hellerve: Yup, but it will emit invalid C by assigning a `void` function to a variable.

eriksvedang: Aha...

eriksvedang: Do you have a reproducible for when that happens?

hellerve:
Same error as in #1187. _sigh_

core/IO.carp
hellerve: The simplest would probably be
```clojure
(let [x (IO.exit)] x)
; or
(and false (IO.exit))
```

hellerve: If it shouldn’t be used in these contexts, the easiest way to do that would probably to restrict its type to `()`.

eriksvedang: OK, I see. Yeah let's go with this then.

PR #1187, opened 2021-03-15
feat: add Char.to-byte and Char.from-byte

hellerve:
This PR adds `Char.to-byte` and `Char.from-byte`.

Test cases are included.

Cheers


hellerve:
Huh, the SDL dependency installation failed on Linux. Might someone be able to restart that check manually?

eriksvedang (approved):

PR #1190, opened 2021-03-29
feat: treat keywords as symbols

### LINKED ISSUE #1189: Some symbols use special AST types that complicate metaprogramming
Currently, a lot of the “special” symbols such as `def`, `defn`, `let`, and `while` (to name a few) get parsed into special AST symbols, which means that they are not behaving like other symbols. This makes a lot of common metaprogramming tasks unnecessarily complicated. Later on in the compiler this sort of differentation makes sense, but in the evaluator it creates more problems than it solves, in my experience.

My suggestion for now would be to make regular symbols that evaluate to these AST forms during expansion, to get the benefit of both worlds. I’m not sure whether this will always work, though.

What do y’all think?

hellerve:
This PR fixes #1189 by adding symbols for all the keywords that resolve to their special AST types. This means that we can actually treat them as symbols in metaprogramming, and as an added bonus when you use `info` on them you get regular documentation.

It also means that we need to do a bit of a dance in `expand` and `commandBuild` to make sure we don’t emit them and always resolve them. This seems brittle, and if we could get the benefits mentioned above without those drawbacks, I’d be happier. Let me know if you have any ideas!

We also had to touch a bunch of different files, but not really much has changed in the grand scheme of things.

Cheers

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -278,7 +278,7 @@ runExeWithArgs ctx exe args = liftIO $ do
 commandBuild :: VariadicCommandCallback
 commandBuild ctx [] = commandBuild ctx [falseXObj]
 commandBuild ctx [XObj (Bol shutUp) _ _] = do
-  let env = contextGlobalEnv ctx
+  let env = removeSpecials (contextGlobalEnv ctx)
       typeEnv = contextTypeEnv ctx
       proj = contextProj ctx
       execMode = contextExecMode ctx
@@ -343,6 +343,12 @@ commandBuild ctx [XObj (Bol shutUp) _ _] = do
           else case Map.lookup "main" (envBindings env) of
             Just _ -> compile True
             Nothing -> compile False
+  where
+    removeSpecials env =
+      let binds = Map.filterWithKey filterSpecials (envBindings env)
+       in env {envBindings = binds}
+    filterSpecials k _ =
+      not (isSpecialSym (XObj (Sym (SymPath [] k) Symbol) Nothing Nothing))
 commandBuild ctx [arg] =
   pure (evalError ctx ("`build` expected a boolean argument, but got `" ++ pretty arg ++ "`.") (xobjInfo arg))
 commandBuild ctx args =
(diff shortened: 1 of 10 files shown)


src/Commands.hs
```diff
@@ -278,7 +278,7 @@ runExeWithArgs ctx exe args = liftIO $ do
 commandBuild :: VariadicCommandCallback
 commandBuild ctx [] = commandBuild ctx [falseXObj]
 commandBuild ctx [XObj (Bol shutUp) _ _] = do
-  let env = contextGlobalEnv ctx
+  let env = removeSpecials (contextGlobalEnv ctx)
```
eriksvedang, line 281: Why is this needed?

src/Obj.hs
```diff
@@ -199,6 +199,11 @@ isSym :: XObj -> Bool
 isSym (XObj (Sym (SymPath _ _) _) _ _) = True
 isSym _ = False
 
+isSpecialSym :: XObj -> Bool
```
eriksvedang, line 202: A lot of languages use "reserved", but maybe "special" is better in Lisp? I know it's usually called "special form".

src/Parsing.hs
```diff
@@ -412,25 +412,8 @@ symbol = do
             Nothing
         )
     else pure $ case last segments of
-      "defn" -> XObj (Defn Nothing) i Nothing
```
eriksvedang, line 415: Shouldn't the definitions for all these `Obj`:s be removed also?

src/Commands.hs
hellerve: Because otherwise those bindings are emitted as globals, since they’re globally bound syms.

src/Obj.hs
hellerve: I also know it as special, but we can switch to "reserved"; I just feel like that sounds like we’re namesquatting.

src/Parsing.hs
hellerve: No, since the symbols now evaluate to the values shown here, instead of being resolved at parsing. Basically, we just push this behavior back into the expander/evaluator.

src/Obj.hs
eriksvedang: Right, let's keep it as "special" then.

src/Parsing.hs
eriksvedang: Ah, OK – now I get it!

PR #1192, opened 2021-03-29
feat: add setf

hellerve:
This PR adds `setf`, or generalized variable setting. It allows for things such as:

```clojure
(def x 1)
(setf x 10)
x ; => 10

(def x [1 2 3])
(setf (nth &x 1) 10)
x ; => [1 10 3]
```

and is extensible using `register-simple-place`, which registers a “place”, i.e. something that can be set. For example:

```clojure
(deftype T [x Int y Int z Int])

(register-simple-place 'T.x 'T.set-x!)

(let-do [x (T.init 0 1 1)]
  (setf (T.x &x) 1)
  x) ; => (T 1 1 1)
```

More complicated setters that emit arbitrary expressions can be registered using `register-place`.

If this PR looks like something we’d like, I’ll write some docs for it.

Cheers

### DIFF
--- a/core/Core.carp
+++ b/core/Core.carp
@@ -59,6 +59,7 @@
 (load-once "Control.carp")
 (load-once "Opaque.carp")
 (load-once "Unit.carp")
+(load-once "Setf.carp")
 
 (posix-only
  (system-include "sys/wait.h")
--- a/core/Setf.carp
+++ b/core/Setf.carp
@@ -1,0 +1,68 @@
+(doc Setf "is a module that emulates a simple version of
+[`setf`](http://www.lispworks.com/documentation/HyperSpec/Body/m_setf_.htm#setf)
+that allows you to register your own places using
+[`register-place`](#register-place).")
+(defmodule Setf
+  (hidden places)
+  (defdynamic places {})
+
+  (doc register-place "registers a place of name `name` using the expression
+builder `builder`.
+
+Example:
+```
+(register-place 'nth (fn [args] `(Array.aset %@args)))
+```")
+  (defndynamic register-place [name builder]
+    (set! Setf.places (Map.put Setf.places name builder)))
+
+  (doc register-simple-place "registers a place of name `name` that uses the
+setter `setter`.
+
+Example:
+```
+(register-simple-place 'nth 'Array.aset)
+```")
+  (defndynamic register-simple-place [name setter]
+    (Setf.register-place name (fn [args] (cons setter args))))
+
+  (hidden get-place)
+  (defndynamic get-place [n]
+    (Map.get Setf.places n))
+
+  (doc setf "sets the value under the place `place` to the value `val`.
+
+Example:
+```
+(def x 1)
+(setf x 10)
+x ; => 10
+
+(def x [1 2 3]
+(setf (nth x 1) 10)
+x ; => [1 10 3]
+```")
+  (defmacro setf [place val]
+    (let [place (if (symbol? place) `(sym %place) place)
+          key (if (and (list? place) (not (empty? place)))
+                 (car place)
+                 'place-malformed)
+          setter? (Setf.get-place key)]
+      (cond
+        (= key 'place-malformed)
+          (macro-error (list "The `setf` place " place " is malformed. A list or symbol was expected"))
+        (= nil setter?)
+          (macro-error (list "I didn’t find a `setf` place for " key ". Is it defined?"))
+        (setter? (cons-last val (cdr place))))))
+)
+
+(use Setf)
+
+(register-simple-place 'sym 'set!)
+(register-simple-place 'nth 'Array.aset!)
+(register-simple-place 'char-at 'String.string-set!)
+; something is fishy with these
+;(register-place 'car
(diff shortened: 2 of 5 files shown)


hellerve:
I also added `register-struct-places`, which  takes a type name (of a struct-like type) and generates places for all the accessors.

To make this work, I had to fix `Symbol.prefix`, which used to set the symbol type. It now uses the symbol type of the symbol that is being  prefixed instead.

eriksvedang:
This is a very neat feature!

I really dislike these kinds of names from CL though, I think they're very badly designed. (What is the "f" even?) 

Is there a possibility of a nicer name?

eriksvedang:
Maybe it means set "format"? 🤔 (like in `printf`...)

eriksvedang:
Does it work nicely with the borrow checker? (like if the array being worked on has strings for example)
It's a little hard to know if the macro does something that affects the type of the arguments, in regards to memory management. (I have not played around with the code yet)

hellerve:
> This is a very neat feature!
> 
> I really dislike these kinds of names (EDIT: with "random" characters at the end) from CL though, I think they're often quite badly designed. (What is the "f" even?)
> 
> Is there a possibility of a nicer name?

I would love to rename this; I did not do so on my own because I was unsure what a  good name would be. As for the name `setf`, it’s old history, like `car` and `cdr`; AFAIK it comes from L Peter Deutsch’s 1973 paper “A Lisp Machine with Very Compact Programs”, where he introduces `setfq`, `setf`’s predecessor, as setting a *q*uoted *f*unction (because of the call syntax).[1]

Anyway, I’m open to any and all rename suggestions, since it’s a bit of a car crash of a name, even though I respect Deutsch very much.

[1]: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.109.6660&rep=rep1&type=pdf

hellerve:
> Does it work nicely with the borrow checker? (like if the array being worked on has strings for example)
> It's a little hard to know if the macro does something that affects the type of the arguments, in regards to memory management. (I have not played around with the code yet)

It doesn’t do anything by default, this is all down to the handlers you define. Like in `derive`, the meat of the actual implementation happens in the user-supplied code. The fun part, though, is that the arguments are passed as-is, such that `ref`ed names, for instance, persist (see also my array and custom type examples above in which I pass `&x` which is handled as expected).

eriksvedang:
Right – like with car/cdr we can of course just say that it's part of Lisp lore and something you should learn. I don't have a better suggestion for the name right now (and to be useful it has to be quite short). I just really get nervous when I read Lisp code and there's a bunch of weirdly suffixed functions like `setf`, `setq` everywhere :)

hellerve:
> Right – like with car/cdr we can of course just say that it's part of Lisp lore and something you should learn. I don't have a better suggestion for the name right now (and to be useful it has to be quite short). I just really get nervous when I read Lisp code and there's a bunch of weirdly suffixed functions like `setf`, `setq` everywhere :)

Yeah, honestly I’d even much prefer `head` and `tail` over `car` and `cdr` everywhere, and I think `setf` deserves a better name, and should probably suffixed with  a `!` since it does an update in place like `set!` does.

eriksvedang:
`set-at!`

TimDeve:
> And I think it’d be useful if people realized that for updates of bigger structures setf is usually the thing to reach for, instead of having to learn all the updaters that might exist for different types.
>
> Does that make sense?

By "bigger structures" do you mean nested data structures? I'm not sure I get why writing your setter in dynamic land is better than writing it in static land. From your example it feel that instead of learning the function you have to learn the "place", it just feels like more "syntax". I'm probably missing something.

hellerve:
> > And I think it’d be useful if people realized that for updates of bigger structures setf is usually the thing to reach for, instead of having to learn all the updaters that might exist for different types.
> > Does that make sense?
> 
> By "bigger structures" do you mean nested data structures? I'm not sure I get why writing your setter in dynamic land is better than writing it in static land. From your example it feel that instead of learning the function you have to learn the "place", it just feels like more "syntax". I'm probably missing something.

I think the important part of making this worthwhile is making the place identical to the getter you would use, to avoid having to learn anything; if you can access it, you know how to change it.

dleslie:
By convention, Scheme uses a postfix `!` to denote that the function modifies its arguments. Ie, `set!` will modify its arguments.

hellerve:
I’m closing this for now, and will instead make this a library over at carpentry. I think that makes more sense for now to gauge whether this is actually useful and/or desired.

I also have come to the conclusions that this basically does the same thing as the lens library, but slightly inferior. Maybe we should add the lens lib to the standard library instead?

hellerve:
> By convention, Scheme uses a postfix `!` to denote that the function modifies its arguments. Ie, `set!` will modify its arguments.

We also have such a convention in Carp—`!` is used for [mutating functions](https://github.com/carp-lang/Carp/blob/master/docs/Terminology.md#mutating-function), which is basically congruent to the Scheme equivalent!

hellerve:
[It’s online](https://github.com/carpentry-org/setf)!

PR #1200, opened 2021-04-07
refactor: Uses def- & defn- macro in core

TimDeve:
As previously discussed this PR uses the macros introduced in https://github.com/carp-lang/Carp/pull/1174

There is a surprisingly small amount of places we use hidden+private in core.


hellerve (approved):

hellerve:
Thanks!

PR #1201, opened 2021-04-18
feat: add Array.map-reduce

hellerve:
This PR adds `Array.map-reduce`, inspired by Elixir’s `Enum-map-reduce`. It’s for mapping over a collection while holding onto state at the same time.

From its docs:

`(Array.map-reduce f acc a)` reduces an array `a` by invoking the function `f` on each
element, while keeping an accumulator and a list.

Returns a `Pair` where the first element is the mapped array and the second one
is the final accumulator.

The function `f` receives two arguments: the first one is the element, and the
second one is the accumulator. `f` must return `(Pair result accumulator)`.

Example:
```clojure
(map-reduce &(fn [acc x] (Pair.init (+ @x @acc) (* @x 2))) 0 &[1 2 3])
; => (Pair 6 [2 4 6])
```

Test cases are included.

Cheers

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -414,7 +414,32 @@ It will create a copy. If you want to avoid that, consider using [`endo-filter`]
     (let-do [darr (allocate (StaticArray.length sarr))]
       (for [i 0 (StaticArray.length sarr)]
         (aset-uninitialized! &darr i @(StaticArray.unsafe-nth sarr i)))
-      darr)))
+      darr))
+
+  (doc map-reduce "reduces an array `a` by invoking the function `f` on each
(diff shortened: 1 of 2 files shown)


eriksvedang:
Great stuff!

My only suggestion is to add a test for it to the memory-tests to check for leaks. If you don't have time, we can just merge it as-is!

hellerve:
I added a memory test for `map-reduce`. It uses strings as accumulator and mapper to trigger potential memory leaks, but it looks good to me.

eriksvedang (approved):
Thanks, awesome!

PR #1205, opened 2021-04-22
fix: fix String.words for multiple word spearators in a row

hellerve:
This PR fixes `String.words` by  ignoring empty words, i.e. multiple spaces in a row. `String-lines` remains untouched, since empty lines might be meaningful.

A test case for this behavior was included.

Cheers


jacereda (approved):

eriksvedang:
Nice catch!

Issue #1183, opened 2021-03-09
Can't use Types name without fully qualifying


TimDeve:
Tried to build an old project and it now fails because referring to types without fully qualifying them is not possible.

Use to work:
```clojure
(defmodule Mod
 (deftype Ok [a Int])
 (deftype NotOk [a Ok]))
```

What is needed now:
```clojure
(defmodule Mod
 (deftype Ok [a Int])
 (deftype NotOk [a Mod.Ok]))
```

scolsen:
Thanks! This is due to the fact that we actually support module encapsulation of types now. Before the original code was really adding `Ok` and `NotOk` to the global type environment even though the calls occur in `Mod`, whereas now it actually adds it to `Mod`. The only problem is the type lookup isn't taking the current context path into account.

TimDeve:
Ok, so what you are saying is that declaring my types at the top level should actually be equivalent to the previous behaviour?

scolsen:
Yes, if you declare them outside the module it should work as expected.

PR #1207, opened 2021-05-03
Refactor: clean up Env module, store type environments in modules

scolsen:
This PR primarily refactors our handling of environments:

- Remove the lookup module; move lookup functions into Env.hs
- Refactor Env.hs functions:
  - Add an Environment class to handle Type envs and value envs generically
  - Unify and rewrite some lookup functions
  - Add additional utilities
- Store type environments in modules:
  - Store sympaths in concrete type names to point to module locations
  - Search for types via the value environment (resolve to the correct module, then look for the type)
  - Update scoring/emissions accordingly.
- refactor main qualification function.
  
While some issues like #1183 persist, handling types defined in modules in a more principled way should help us start resolving them going forward.


eriksvedang:
Seems like a compiler error snuck in there:

```
/home/runner/work/Carp/Carp/test/TestLookup.hs:24:41: error:
    Not in scope: ‘E.lookup’
    Module ‘Env’ does not export ‘lookup’.
   |
24 | basicLookup = assertNotFound (fmap snd (E.lookup emptyRootEnv (SymPath [] "nonexisting")))
   |                            
```

scolsen:
> Seems like a compiler error snuck in there:
> 
> ```
> /home/runner/work/Carp/Carp/test/TestLookup.hs:24:41: error:
>     Not in scope: ‘E.lookup’
>     Module ‘Env’ does not export ‘lookup’.
>    |
> 24 | basicLookup = assertNotFound (fmap snd (E.lookup emptyRootEnv (SymPath [] "nonexisting")))
>    |                            
> ```

Ah, drat, forgot to update the tests! Should be fixed w/ the latest commit, thanks!

src/Context.hs
```diff
@@ -121,70 +181,110 @@ replaceHistory' = flip replaceHistory
...
+              (E.insert (TypeEnv (contextGlobalEnv ctx)) (SymPath path name) binder)
+                >>= pure . (replaceGlobalEnv ctx) . getTypeEnv
+        )
+
+replaceTypeBinder :: Context -> QualifiedPath -> Binder -> Either ContextError Context
+replaceTypeBinder ctx qpath binder =
+  let (SymPath path name) = unqualify qpath
+   in first
+        (\_ -> trace (show path) (FailedToInsertInTypeEnv (unqualify qpath) binder))
+        ( (E.replaceInPlace (contextTypeEnv ctx) name binder)
```
eriksvedang, line 221: Stylistically, I think this would look better if it's one line and drops a bunch of parenthesis (it must surely be possible, due to the operators being used?)

src/Concretize.hs
```diff
@@ -87,8 +87,8 @@ concretizeXObj allowAmbiguityRoot typeEnv rootEnv visitedDefinitions root =
         let functionEnv = Env Map.empty (Just env) Nothing Set.empty InternalEnv 0
             envWithArgs =
               foldl'
-                ( \e arg@(XObj (Sym (SymPath _ argSymName) _) _ _) ->
-                    extendEnv e argSymName arg
+                ( \e arg@(XObj (Sym path _) _ _) ->
+                    let Right v = insertX e path arg in v
```
eriksvedang, line 91: Are you sure this never fails? (seems kind of off to have an incomplete pattern among all the new safety..!)

src/Concretize.hs
```diff
@@ -436,43 +439,41 @@ concretizeType _ ft@FuncTy {} =
...
       Right (defineStaticArrayTypeAlias arrayTy : concat deps)
--- TODO: handle polymorphic constructors (a b)
-concretizeType typeEnv genericStructTy@(StructTy (ConcreteNameTy name) _) =
-  case lookupInEnv (SymPath lookupPath structName) (getTypeEnv typeEnv) of
-    Just (_, Binder _ (XObj (Lst (XObj (Deftype originalStructTy) _ _ : _ : rest)) _ _)) ->
+concretizeType typeEnv genericStructTy@(StructTy (ConcreteNameTy (SymPath _ name)) _) =
+  -- TODO: This function only looks up direct children of the type environment.
+  -- However, spath can point to types that belong to a module. Pass the global env here.
+  case (getTypeBinder typeEnv name) of
+    Right (Binder _ x) -> go x
```
eriksvedang, line 459: What does `go` stand for?

src/Concretize.hs
```diff
@@ -677,26 +678,29 @@ concretizeDefinition allowAmbiguity typeEnv globalEnv visitedDefinitions definit
...
     predicate (Just t) =
       --trace ("areUnifiable? " ++ show functionType ++ " == " ++ show t ++ " " ++ show (areUnifiable functionType t)) $
       areUnifiable functionType t
     predicate Nothing = error "allfunctionswithnameandsignature"
-    foundBindings = case lookupBinder (SymPath [] functionName) (getTypeEnv typeEnv) of
+    foundBindings = case getTypeBinder typeEnv functionName of
       -- this function is an interface; lookup implementations
-      Just (Binder _ (XObj (Lst (XObj (Interface _ paths) _ _ : _)) _ _)) ->
+      Right (Binder _ (XObj (Lst (XObj (Interface _ paths) _ _ : _)) _ _)) ->
         -- N.B./TODO: There are functions designed for this
```
eriksvedang, line 690: This comment maybe can be changed now? :)

src/Concretize.hs
```diff
@@ -677,26 +678,29 @@ concretizeDefinition allowAmbiguity typeEnv globalEnv visitedDefinitions definit
...
-          Nothing -> (multiLookupEverywhere functionName env)
+        case sequence $ map (\p -> searchValue env p) (paths ++ [(SymPath [] functionName)]) of
+          Right found -> found
+          Left _ ->
+            case findPoly env functionName functionType of
+              Right r -> [r]
+              Left _ -> (lookupEverywhere env functionName)
       -- just a regular function; look for it
-      _ -> (multiLookupEverywhere functionName env)
+      _ -> fromRight [] ((fmap (: []) (Env.getValue env functionName)) <> pure (lookupEverywhere env functionName))
```
eriksvedang, line 703: The old function was just `multiLookupEverywhere functionName env` – is there a chance that this impenetrable line can be replaced by such a function again? Or is that "feature" only used here?

src/Polymorphism.hs
```diff
@@ -1,24 +1,30 @@
...
+--
+-- TODO: Environments are passed in different order here!!!
 nameOfPolymorphicFunction :: TypeEnv -> Env -> Ty -> String -> Maybe SymPath
 nameOfPolymorphicFunction _ env functionType functionName =
-  let foundBinders = multiLookupEverywhere functionName env
-   in case filter ((\(Just t') -> areUnifiable functionType t') . xobjTy . binderXObj . snd) foundBinders of
-        [] -> Nothing
-        [(_, Binder _ (XObj (Lst (XObj (External (Just name)) _ _ : _)) _ _))] ->
+  let foundBinder =
+        (E.findPoly env functionName functionType)
```
eriksvedang, line 22: I see this "pattern" pop up in a bunch of places – does this mean that we do `findPoly` first in the env and then in the global env (progenitor) if nothing was found in env?

eriksvedang (commented):
A very interesting read, I learned a bunch! So much good stuff, this is definitely a step up in terms of Haskell code quality.

I don't think we're far from getting this merged, but there were a few places where things could maybe be improved, and I also have some questions just to understand the code better (to be able to co-maintain it in the future.)

Cheers!

core/Color.carp
```diff
@@ -26,7 +26,7 @@
      BgWhite)
   (use Id)
   (defn hash [k]
-    (get-tag k))
+    (get-tag (the (Ref Color.Id) k)))
```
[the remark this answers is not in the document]
scolsen: So the error was a bit subtle, here was the problem:

1. When we qualify a function, we create two new environments, a function env and a recursive env, respectively, the parent of the recursive env is the original inner env from the context (e.g. current module being evaluated)
2. We then pass this function env to qualify syms, however, I didn't copy over the *Used* modules of the original env.

So the solution is to copy the use'd envs from the module over to the function env. 

Similarly, I noticed a but in primitiveUse that I fixed, and updated primitive use to take account of the context (for this particular case, color `use`s `Id`, but really it means to use `Color.Id`, so Use now accounts for this.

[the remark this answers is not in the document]
scolsen: The reason it was failing is because `using` ID fixes the type correctly via `get-tag`, but since it originally wasn't looking at the use modules correctly, it was falling through to `lookupeverywhere` and keeping the function generic.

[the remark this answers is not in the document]
scolsen: But, I'll have to think about the right approach a bit more, as this introduces some other problematic cases (e.g. `copy`, etc. We probably need to introduce an interface check).

src/Types.hs
```diff
@@ -60,7 +61,7 @@ data Ty
   | RefTy Ty Ty -- second Ty is the lifetime
   | StaticLifetimeTy
   | StructTy Ty [Ty] -- the name (possibly a var) of the struct, and it's type parameters
-  | ConcreteNameTy String -- the name of a struct
+  | ConcreteNameTy SymPath -- the name of a struct
```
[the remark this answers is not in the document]
eriksvedang: Right, makes sense! Thanks for the explanation.

src/SymPath.hs
```diff
@@ -89,3 +90,9 @@ pathToC (SymPath modulePath name) =
 consPath :: [String] -> SymPath -> SymPath
 consPath qualifyers (SymPath stringPaths name) =
   SymPath (qualifyers ++ stringPaths) name
+
+-- | Convert a list of strings into a path.
+fromStrings :: [String] -> SymPath
```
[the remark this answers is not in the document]
eriksvedang: Oh, I see. OK!

src/Polymorphism.hs
eriksvedang: Excellent!

core/Color.carp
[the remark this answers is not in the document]
scolsen: Fixed. I actually just made a mistake. All we needed to do was copy over use module paths to function envs.

src/Concretize.hs
scolsen: Yes. It will never fail because:

1. We are inserting into a brand new functionEnv, declared above.
2. We are inserting binders for the functions arguments, which *must* be unqualified (checked elsewhere).

Though I agree it'd be better to do this in a safe way. I've added a comment to this effect for now.

src/Concretize.hs
scolsen: Nothing in particular. It's just a haskell convention for local function definitions: https://stackoverflow.com/questions/5844653/haskell-why-the-convention-to-name-a-helper-function-go. Admittedly it's also a function of my relative laziness while writing.

src/Concretize.hs
scolsen: Yes, sg!

src/Concretize.hs
scolsen: If we only called `(lookupEverywhere env functionName))` it would amount to the same thing but lead to cases where we process a whole chain of envs when all we really need is the one env (when the function is a direct child)

scolsen:
@eriksvedang pushed some fixes for our discussions. 

There's still quite of bit of HLint related clean up (removing parens, etc.), but otherwise I think the majority of the initial comments should be resolved.

eriksvedang:
> @eriksvedang pushed some fixes for our discussions. 
> 
> 
> 
> There's still quite of bit of HLint related clean up (removing parens, etc.), but otherwise I think the majority of the initial comments should be resolved.

Cool! HLint could also be done in a later PR, I'm sure there is more (unrelated) places that it will complain about.

scolsen:
> > @eriksvedang pushed some fixes for our discussions.
> > There's still quite of bit of HLint related clean up (removing parens, etc.), but otherwise I think the majority of the initial comments should be resolved.
> 
> Cool! HLint could also be done in a later PR, I'm sure there is more (unrelated) places that it will complain about.

Yeah, there's quite a bit to clean up. Up to you, happy to tackle the ones for these files in this PR, otherwise we can do a separate one.

eriksvedang:
> > > @eriksvedang pushed some fixes for our discussions.
> > > There's still quite of bit of HLint related clean up (removing parens, etc.), but otherwise I think the majority of the initial comments should be resolved.
> > 
> > 
> > Cool! HLint could also be done in a later PR, I'm sure there is more (unrelated) places that it will complain about.
> 
> Yeah, there's quite a bit to clean up. Up to you, happy to tackle the ones for these files in this PR, otherwise we can do a separate one.

I think second one is best, keeps it a bit more readable in the history.

src/Qualify.hs
```diff
@@ -291,104 +320,93 @@ qualifyLst typeEnv globalEnv env (XObj (Lst xobjs) i t) =
...
-                  --trace ("Will turn '" ++ show path ++ "' " ++ prettyInfoFromXObj xobj ++ " into an interface symbol.")
-                  createInterfaceSym name
-            _ ->
-              doesNotBelongToAnInterface False localEnv
-        -- Qualified:
-        _ ->
-          doesNotBelongToAnInterface False localEnv
+-- Unqualified path.
+qualifySym typeEnv globalEnv localEnv xobj@(XObj (Sym path@(SymPath _ name) _) i t) =
+  ( ( ( replaceLeft
```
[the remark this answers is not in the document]
scolsen: It's similar to the old version of this function, but I tried to make it slightly cleaner by performing lookups first (what happens in the first part), then based on the result, resolving/qualifying the symbol appropriately.

There's a lot going on here and it can probably be improved even further/made more clear in future commits. 

Here's the gist of what the entire function does:

1. Try the following lookups on the unqualified symbol, taking the first one that succeeds:
    1. Look for the symbol directly in the global type environment.
    2. Look for the symbol directly in the local environment (usually a module or function env)
    3. Look for the symbol directly in the global environment
    4. Look for the symbol in all the `use`d environments of the local environment
    5. Look for the symbol "contextually" (current env + all parents)
    6. Finally, if all else turned up nothing, look for the symbol in every possible env, starting from the global env (lookupEverywhere)
2. Based on which lookup succeeds, we'll either get a single binder and resolve that `resolve`, or we'll get multiple binders (in a list), (i.e. there's more than one possible match) and resolve those `resolveMulti`
    1. `resolve` just performs special checks so that we qualify the binder in the right way in case there's some edge cases:
        1. Check if the binder is an interface, if so, ensure it's not shadowed by some other local binding. If it isn't, return an `InterfaceSym`
        2. Check if the binder is an external override binding
        3. Check if the binding is the name of a module (assume it's naked init of a type)
        4. For all other binders, check the env mode of where the binder is defined and whether or not the binding is captured, and assign the lookup mode appropriately.
    2. `resolveMulti` Attempts to add all the candidate paths to a `MultiSym`, but first it ensures:
        1. The input path wasn't already qualified (since if this is true, we should know precisely what symbol we need)
        2. There are no local definitions among the candidates (none of the possible path matches are defined in an environment with mode local) since this would mean an environment "leaked" into a broader scope.

eriksvedang:
@scolsen Hey, is this ready for review again?

scolsen:
> @scolsen Hey, is this ready for review again?

Yep, ready for another look whenever.

eriksvedang (approved):
Looks great!

src/Scoring.hs
```diff
@@ -79,17 +76,23 @@ depthOfType typeEnv visited selfName theType =
...
+                --where
+                --  moduleDepth = length lookupPath * 1000 -- modules have score 1000
+                Left _ ->
                   --trace ("Unknown type: " ++ name) $
-                  depthOfVarTys -- The problem here is that generic types don't generate
-                  -- their definition in time so we get nothing for those.
-                  -- Instead, let's try the type vars.
+                  -- Two problems here:
+                  --
+                  -- 1. generic types don't generate their definition in time
```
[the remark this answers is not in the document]
eriksvedang: OK, thanks for the explanation!

src/Qualify.hs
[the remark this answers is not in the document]
eriksvedang: Great, I'll refer back to this if/when I have to touch this code :)

PR #1208, opened 2021-05-21
fix: don't crash validation on invalid types

### LINKED ISSUE #1202: Malformed type members can crash the compiler
Sometimes the compiler crashes when type members are malformed.

Example:

```clojure
(deftype X [member []])
```

Here `member` should receive a type, but instead gets an empty array. This should lead to an error, but lead to a crash in `validate` (through use of `fromJust`).

Cheers

scolsen:
This quick fix reorders the ordering of our type def validation checks
to prevent invalid types like [member []] from crashing the compiler.

fixes #1202

### DIFF
--- a/src/Validate.hs
+++ b/src/Validate.hs
@@ -26,7 +26,7 @@ validateMemberCases typeEnv typeVariables rest = mapM_ visit rest
 
 validateMembers :: TypeEnv -> [Ty] -> [XObj] -> Either TypeError ()
 validateMembers typeEnv typeVariables membersXObjs =
-  checkUnevenMembers >> checkDuplicateMembers >> checkKindConsistency >> checkMembers
+  checkUnevenMembers >> checkDuplicateMembers >> checkMembers >> checkKindConsistency
   where
     pairs = pairwise membersXObjs
     -- Are the number of members even?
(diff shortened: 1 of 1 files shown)


src/Validate.hs
```diff
@@ -52,6 +52,8 @@ validateMembers typeEnv typeVariables membersXObjs =
         Left var -> Left (InconsistentKinds var membersXObjs)
         _ -> pure ()
       where
+        -- fromJust is safe here; invalid types will be caught in the prior check.
+        -- todo? be safer anyway?
```
hellerve, line 56: I think it’d be good to be safe just in case anyone thinks of refactoring this in the future.

eriksvedang: Agreed! But let's merge this anyways, right?

hellerve: I think this can be merged unless @scolsen wants to add this to this PR!

scolsen: Let's merge as-is if that's ok. Fixing it in a safe way will require changing the signature of `checkMembers` and the functions it calls ever so slightly which is a bit more involved since we'd need to update call sites, we could also re-evaluate the same check in the body of `checkKinds` but that seems silly.

hellerve: Seems good to me!

Issue #1179, opened 2021-03-04
Dynamic mod can cause hangs


scolsen:
`Dynamic.mod` will cause loops on some arguments, the following code causes a hang:

```
;; assume arr is [x y]
(= 0 (mod (length arr) 2)))
```

Changing `mod` to `imod` fixes the code above. However, the definition of `Dynamic.mod` itself seems fine, so this is likely stemming from some compiler issue.

scolsen:
The underlying problem is coming from the `length` call. Using a plain old number works fine:

```
(= 0 (mod 2 2))
```

scolsen:
Here's the fix: call `eval` on `length`

```
鲤 (= 0 (mod (eval (length ['x 'y])) 2))
=> true
鲤 

```

scolsen:
This is likely a regression from some other change—seems related to some other differences re: the necessity of calling `eval`. What's strange is that it doesn't seem to have affected things uniformly.

scolsen:
Taking a quick look at the code, there's no obvious reason for this issue. `length` is defined as a command, which afaict, is still evaluated immediately in the evaluator 🤔

scolsen:
If I had to guess, we might need to reconsider the evaluator's symbol lookup clause, since this seems to be the only place where the hang might be happening (command handling hasn't changed)—we don't evaluate enough when retrieving some symbol or something.

hellerve:
Is it possible `length` is recomputed every time the first argument is used/looked up (maybe this is related to laziness in a subtle way)?

scolsen:
> Is it possible `length` is recomputed every time the first argument is used/looked up (maybe this is related to laziness in a subtle way)?

Ah that's a good thought. I do think the issue is definitely too much laziness somewhere. Tinkering now.

scolsen:
Ah, it's the final clause in the Evaluator symbol lookup (use modules lookup) that's causing the problem; in rare cases, e.g. this one, the use module name winds up conflicting with a parent module name and causes a loop. Yet another issue related to module relationship mgmt 😢 

```haskell
( foldl                                                                                                                                                                                       
                    (<|>)                                                                                                                                                                                     
                    Nothing                                                                                                                                                                                   
                    ( map                                                                                                                                                                                     
                        ( \(SymPath p' n') ->                                                                                                                                                                 
[...]

PR #1209, opened 2021-05-21
fix: don't type check untyped forms in set!

scolsen:
This is a kludgey temporary fix for #1179. The problem is that set!
occasionally receives untyped forms. In particular, this seems to happen
when the value of the form is determined by a command such as `length`.
This caused hangs when evaluating `Dynamic.mod` in particular.

For now, we just treat these like dynamic values and don't worry about
typechecking them, which seems to do the trick.

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -1048,8 +1048,12 @@ specialCommandSet ctx [orig@(XObj (Sym path@(SymPath _ _) _) _ _), val] =
     evalAndSet :: Binder -> (Context -> Env -> Either EvalError XObj -> Binder -> IO (Context, Either EvalError XObj)) -> Env -> IO (Context, Either EvalError XObj)
     evalAndSet binder setter env =
       case xobjTy (binderXObj binder) of
-        -- don't type check dynamic bindings
+        -- don't type check dynamic or untyped bindings
+        -- TODO: Figure out why untyped cases are sometimes coming into set!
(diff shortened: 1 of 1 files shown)


src/Eval.hs
```diff
@@ -1048,10 +1048,14 @@ specialCommandSet ctx [orig@(XObj (Sym path@(SymPath _ _) _) _ _), val] =
     evalAndSet :: Binder -> (Context -> Env -> Either EvalError XObj -> Binder -> IO (Context, Either EvalError XObj)) -> Env -> IO (Context, Either EvalError XObj)
     evalAndSet binder setter env =
       case xobjTy (binderXObj binder) of
-        -- don't type check dynamic bindings
+        -- don't type check dynamic or untyped bindings
+        -- TODO: Figure out why untyped cases are sometimes coming into set!
         Just DynamicTy ->
           evalDynamic ResolveLocal ctx val
             >>= \(newCtx, result) -> setter newCtx env result binder
```
hellerve, line 1055: Since the two cases for `Just DynamicTy` and `Nothing` are the same, should we pull them into a `where`?

hellerve (approved):
Thanks for looking into this!

src/Eval.hs
scolsen: Nice catch, done!

hellerve (approved):
Looks great now!

PR #1210, opened 2021-05-22
fix: properly assign types to recursive calls

### LINKED ISSUE #348: argument count not checked for recursive calls
E.g.
```
(defn recurse [a b]
  (let [c (+ b 1)
        out (recurse c)]
    (+ out 1)))

(defn main []
  (println* (recurse 1 2)))
```
builds fine but crashes with a segmentation fault.

### LINKED ISSUE #349: return type not checked for recursive calls
This crashes with segmentation fault:
```
(defn recurse [a]
  (if (= a 0)
    ;; We return an Int
    0
    ;; But treat the return value of the recursive call as &Int
    @(recurse (dec a))))

(defn main []
  (println* (recurse 1)))

### PR DESCRIPTION
**Pull request by scolsen**
This commit fixes an issue whereby all recursive calls were given var
types, this resulted in strange behavior such as the following code:

```clojure
(defn recurse [a b]
  (let [c (+ b 1)
        out (recurse c)]
    (+ out 1)))

(defn main []
  (println* (recurse 1 2)))
```

compiling just fine and yielding runtime segfaults. Now, if a recursive
instance of a symbol was previously typed, we use that type--if not, we
assign a fresh type variable (the old behavior). This fixes code like
that above, throwing an error at compile time about an incorrect number
of arguments for `recurse`.

Fixes #348, Fixes #349

### DIFF
--- a/src/InitialTypes.hs
+++ b/src/InitialTypes.hs
@@ -116,11 +116,16 @@ initialTypes typeEnv rootEnv root = evalState (visit rootEnv root) 0
       -- catchall case for exhaustive patterns
       unknown -> pure (Left (InvalidObj unknown xobj))
     visitSymbol :: Env -> XObj -> SymPath -> State Integer (Either TypeError XObj)
-    visitSymbol _ xobj@(XObj (Sym _ LookupRecursive) _ _) _ =
-      -- Recursive lookups are left untouched (this avoids problems with looking up the thing they're referring to)
-      do
-        freshTy <- genVarTy
-        pure (Right xobj {xobjTy = Just freshTy})
+    visitSymbol e xobj@(XObj (Sym name LookupRecursive) _ _) _ =
+        case E.searchValueBinder e name of
+          -- If this recursive symbol is already typed in this environment, use that type.
+          -- This is relevant for, e.g. recursive function calls.
+          -- We need to use search here to check parents as our let-binding handling possibly puts recursive
+          -- environments as the parent of a more local environment for the let bindings.
+          Right (Binder _ found) -> pure (Right xobj {xobjTy = xobjTy found})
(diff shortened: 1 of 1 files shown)


scolsen:
For reference, here's the new behavior:

```
[i] carp ∎ carpdev -x bad.carp     
You used the wrong number of arguments in 'recurse' at line 7, column 9 in 'bad.carp'. I expected 2, but got 1. at bad.carp:6:3.

Traceback:
(defn recurse [a b] (+ (recurse 1) 3)) at bad.carp:6:2.
```

scolsen:
Incidentally (though unsurprisingly), this also fixes issue #349 

```clojure
;; other-bad.carp
(defn recurse [a]
  (if (= a 0)
    ;; We return an Int
    0
    ;; But treat the return value of the recursive call as &Int
    @(recurse (dec a))))

(defn main []
  (println* (recurse 1)))
```

```
[i] carp ∎ carpdev -x other-bad.carp 
I can’t match the types `Int` and `(Ref Int r9)`. within `(copy (recurse (dec a)))`

  (recurse (dec a)) : Int
  At line 6, column 6 in 'other-bad.carp'

  Expected first argument to 'copy' : (Ref Int r9)
  At line 6, column 5 in 'other-bad.carp' at other-bad.carp:1:2.

Traceback:
(defn recurse [a] (if (= a 0) 0 (copy (recurse (dec a))))) at other-bad.carp:1:1.
```

hellerve (approved):
Looks good! This also seems to work fine on weird shadowing `let` bindings, which  was my biggest concern.

🎉 Thank you 🎉 

An example for the kind of `let` bindings that I was concerned about:
```clojure
(defn f []
  (let [f (fn [a b] (+ a b))]
    (f 1 2)))
```

eriksvedang:
Wow, such a clean fix! Let's add regression tests for these too.

PR #1211, opened 2021-05-22
fix: concretize def forms

### LINKED ISSUE #364: Defining functions with def doesn't work
Code:
```
(defn duplicate-arg [f]
  (fn [x] (f x x)))

(def double (duplicate-arg +))

(defn main []
  (println* (double 3)))
```
Error:
```
Unable to concretize '(def double (duplicate-arg +))' at line 4, column 1 in '.../dbg.carp'
```

scolsen:
Most def forms bind values, so their types are wholly determined at
initial assignment time. However, defs that bind to lambdas may have
polymorphic types by virtue of their use of interfaces:

```
(defn duplicate-arg [f]
  (fn [x] (f x x)))

(def double (duplicate-arg +))

(defn main []
  (println* (double 3)))
```

Previously, the above program would yield a concretization error. This
commit adds concretization handling for def forms, and makes it possible
to call lambdas bound to defs as is done in the program above.

Fixes #364

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -638,6 +638,19 @@ concretizeDefinition allowAmbiguity typeEnv globalEnv visitedDefinitions definit
       suffix = polymorphicSuffix polyType concreteType
       newPath = SymPath pathStrings (name ++ suffix)
    in case definition of
+        -- todo: repeat of defn below
+        XObj (Lst (XObj Def _ _ : _)) _ _ ->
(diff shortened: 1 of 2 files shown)


hellerve (approved):
This is awesome! 🎉

eriksvedang:
Yeah, super mega great!!! Can you add a usage of this in somewhere in the test suite to avoid future regressions? (eg in `test/regression.carp`)

PR #1212, opened 2021-05-22
fix: rename type variables during concretization to prevent collisions

### LINKED ISSUE #521: UnificationFailure

scolsen:
Our use of type variables in deftype forms to support parametric
polymorphism relied on precise matching between variables in the head
and the body of the form. However, this could lead to unification
failure when the concretizer encountered a type that mixed two
polymorphic types that use the same type var name:

```
(deftype (HitRecord a) [ t a ])
(deftype (CurrentHit a) [ hr (Maybe (HitRecord a)) ])
```

Concretizing this would previously attempt to bind both Maybe(HitRecord
) and just (HitRecord) to `a`.

To fix this, we temporarily rename all type variables in type
definitions during concretization. Names are preserved in place, so
there's no danger of losing a variable. The code above will now work:

```
=> (CurrentHit.init (Just (HitRecord.init 5)))
(CurrentHit (Just (HitRecord 5)))
=> (HitRecord.init 3)
(HitRecord 3)
```

Fixes #521

(finally got to this)

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -27,6 +27,7 @@ import TypesToC
 import Util
 import Validate
 import Prelude hiding (lookup)
+import InitialTypes
 
 data Level = Toplevel | Inside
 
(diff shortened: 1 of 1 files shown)


hellerve (approved):
That’s a good catch. Also your first paragraph is the most exciting techno-babble for the uninitiated 😸

eriksvedang:
Wow, old one! Very happy you figured this one out!

PR #1216, opened 2021-05-24
fix: don't expand inner module macros on first pass; privacy

### LINKED ISSUE #467: Prevent usage of functions marked as private outside of their module
Currently, marking a function as `private` is purely informative. Ideally, the compiler should enforce that the function isn’t used outside that module. It should also cover interesting cases like:

```clojure
(defmodule T
  (private f)
  (defn f [] (println* "I can’t be called outside!"))
)
(T.f) ; this should fail
(defmodule T
  (defn f2 [] (f)) ; this should work, since it’s inside the module, though in a different form
)
```

Cheers

### LINKED ISSUE #1213: Another way to hang the compiler on recursive functions
I was experimenting with the fix for #349, and came up with this thing that still hangs:

```clojure
鲤 (sig g (Fn [Float] Float))
鲤 (defn g [x] (g))
```

Seems like the addition of the signature makes the difference, without it you get the appropriate warning.

scolsen:
This commit changes the behavior of expansions to avoid expanding module
expressions until we're actually processing the module in question.

Previously, the following form would be entirely expanded at the time of evaluating A:

```clojure
(defmodule A <- current environment

  (some-macro) <- expand

  (defmodule B
    (some-macro f) <- expand, current env is A, *NOT* B.
    so if this expands to
    (private f)
    (defn f ....)
    the f of the expansion is added to *A*, and we have a duplicate
    ghost binder.
  )

  (defn foo [] B.f) <- expand, B.f does not exist yet, any meta on the
  binding will be ignored, permitting privacy errors since expansion
  ignores undefined bindings, instead, we'll look this up at eval time,
  and not check privacy as doing so would cause problems for legitimate
  cases.
)
```

This meant that if the macro happened to have side-effects, e.g. calling
`meta-set!` we'd produce side-effects in the wrong environment, A,
resulting in duplicate bindings, missing bindings at evaluation time,
and other problems.

Now, we instead process the form as follows:

```clojure
(defmodule A <- current environment

  (some-macro) <- expand

  (defmodule B
    (some-macro f) <- wait
  )

  (defn foo [] B.f)
)

;; step 2
(defmodule A

  (foo-bar ) <- previously expanded macro

  (defmodule B <- current environment
    (some-macro f) <- expand
  )
  ....
)
```

In general, this prevents the generation of a bunch of unintentional and
incorrectly added bindings when calling `meta-set!` from various macros.

Additionally, privacy constraints are now carried across nested modules:

```
(defmodule A
  (defmodule B
    (private f)
    (defn f [] 0)
  )
  (defn g [] (B.f)) ;; Privacy error!
)
```

This change also fixed an issue whereby recursive functions with `sig`
annotations could trick the compiler. Again, this had to do with the
unintentionally added bindings stemming from expansion of nested module
expressions via meta-set.

Fixes #1213, Fixes #467

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -102,12 +102,12 @@ eval ctx xobj@(XObj o info ty) preference resolver =
              pure (ctx, Right (resolveDef found))
         getGlobal :: SymPath -> Maybe (Context, Either EvalError XObj)
         getGlobal path =
-          do (Binder _ found) <- maybeId (E.findValueBinder (contextGlobalEnv ctx) path)
-             pure (ctx, Right (resolveDef found))
+          do (Binder meta found) <- maybeId (E.findValueBinder (contextGlobalEnv ctx) path)
+             checkPrivate meta found
         tryDynamicLookup :: Maybe (Context, Either EvalError XObj)
         tryDynamicLookup =
-          do (Binder _ found) <- maybeId (E.searchValueBinder (contextGlobalEnv ctx) (SymPath ("Dynamic" : p) n))
-             pure (ctx, Right (resolveDef found))
+          do (Binder meta found) <- maybeId (E.searchValueBinder (contextGlobalEnv ctx) (SymPath ("Dynamic" : p) n))
+             checkPrivate meta found
         getLocal :: String -> Maybe (Context, Either EvalError XObj)
         getLocal name =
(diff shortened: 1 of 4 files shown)


eriksvedang:
Great stuff!

PR #1218, opened 2021-05-24
test: Regression tests for recent improvements and bug fixes

eriksvedang:


test/regression.carp
```diff
@@ -51,6 +51,35 @@
...
+;; defining function with def (not defn)
+(defn duplicate-arg [f]
+  (fn [x] (f x x)))
+
+(def double (duplicate-arg +))
+
+(defn defining-function-with-def []
+  (double 3))
+
+;; using special symbol as binder name
```
hellerve, line 63: Is the test here just defining it? If so, maybe a mention of the issue number would be good!

test/regression.carp
```diff
@@ -51,6 +51,35 @@
...
+(defndynamic special-symbol-as-binder []
+  (let [c (car (list 1 2 3))]
+     (list c)))
+
+;; dynamic closure referring to itself
```
hellerve, line 68: See above.

hellerve (approved):
Thank you!

test/regression.carp
eriksvedang: Yes, will fix.

scolsen:
Thanks!

PR #1219, opened 2021-05-24
chore: Re-format Haskell code

eriksvedang:


src/Concretize.hs
```diff
@@ -525,9 +526,9 @@ instantiateGenericStructType typeEnv originalStructTy@(StructTy _ _) genericStru
               nameFixedMembers = renameGenericTypeSymbolsOnProduct renamedOrig memberXObjs
               validMembers = replaceGenericTypeSymbolsOnMembers mapp nameFixedMembers
               concretelyTypedMembers = replaceGenericTypeSymbolsOnMembers mappings memberXObjs
-           -- We only used the renamed types for validation--passing the
-           -- renamed xobjs further down leads to syntactical issues.
-           in case validateMembers typeEnv renamedOrig validMembers of
+           in -- We only used the renamed types for validation--passing the
+              -- renamed xobjs further down leads to syntactical issues.
+              case validateMembers typeEnv renamedOrig validMembers of
```
hellerve, line 531: This is a weird reformat for me

src/Eval.hs
```diff
@@ -89,30 +89,33 @@ eval ctx xobj@(XObj o info ty) preference resolver =
         unwrapLookup =
           fromMaybe
             (throwErr (SymbolNotFound spath) ctx info)
-        -- | Try all lookups performs lookups for symbols based on a given
-        -- lookup preference.
+
```
hellerve, line 92: Where did this comment go?

hellerve: It might be because it’s a docstring comment with `|`?

src/Concretize.hs
eriksvedang: It's what Ormolu did...

src/Eval.hs
eriksvedang: Huh, weird. Will add it back – thanks for noticing!

src/Concretize.hs
eriksvedang: I can't find a good way around it, unfortunately 🤷‍♂️

scolsen:
Thanks!

PR #1223, opened 2021-05-26
feat: Remove `address`

eriksvedang:
By @jacereda. I fixed the merge conflicts.


eriksvedang:
The failing test is about taking the address of a non-symbol. With this PR we allow any Ref to be passed to `address` – does that seem OK?

scolsen:
Sounds good to me!

eriksvedang:
Probably makes more sense to move address into the Pointer module also. Will fix.

PR #1225, opened 2021-05-26
fix: fix repl autocomplete

### LINKED ISSUE #1224: Auto completion in modules are broken
To reproduce, in the REPL, type:

`Array.` and then tab. You should get the functions in the Array module but instead you get nothing.

scolsen:
The recent Env refactor commits unwittingly broke the REPL's
autocompletion. This change restores the previous behavior.

Fixes #1224


eriksvedang:
merci beaucoup!

PR #1227, opened 2021-05-26
feat: add Dynamic.get-env and Dynamic.set-env

hellerve:
This PR adds `Dynamic.get-env` and `Dynamic.set-env` for reading and writing environment variables dynamically, as requested by @guberathome.

Cheers


scolsen (approved):
Woo!

Issue #1043, opened 2020-12-01
This error should not crash the compiler - change return type to Either here


charlesetc:
Hi, thanks for working on carp! I found this bug:

```
$ carp
Welcome to Carp 0.4.2
This is free software with ABSOLUTELY NO WARRANTY.
Evaluate (help) for more information.
> (def lines (String.lines (ref (IO.read-file "/tmp/input.txt"))))
> (def numbers (Array.copy-map &(fn [x] (Maybe.unsafe-from (Int.from-string x))) &lines))
> (Array.copy-map &(fn [x] (Pair.init x ())) &numbers)
carp: I can’t use the type `(Ref Int <StaticLifetime>)` as a member type at no info.

Is it defined and captured in the head of the type definition?. This error should not crash the compiler - change return type to Either here.
CallStack (from HasCallStack):
  error, called at src/Deftype.hs:309:30 in CarpHask-0.4.2.0-K2sdzDEAjn25KRaWCaWwhV:Deftype
```

I'm running ubuntu 20.04. Let me know if there's any more information I can help with!

charlesetc:
Ah maybe I misinterpreted a type error as a compilation error? I'm not sure what "this error should not crash the compiler" means when preceding a stack trace...

eriksvedang:
Sorry about this error, it's more of a note to the compiler writers than to the end user. It hints at an actual compiler error though, you currently can't store refs in structs. But it shouldn't crash the compiler, of course.

TimDeve:
Yes, in that case that's a type error that's a bit to noisy. `copy-map` passes a reference to its function so to store that in a pair you need to do a copy:

```clojure
(Array.copy-map &(fn [x] (Pair.init @x @x )) &numbers)
```

eriksvedang:
Oh wait, I actually misread the error. What @TimDeve said.

eriksvedang:
> Ah maybe I misinterpreted a type error as a compilation error? I'm not sure what "this error should not crash the compiler" means when preceding a stack trace...

It should be fixed. On it.

eriksvedang:
Some investigation shows that we shouldn't have to call `concretizeType` in the deps functions for init (this applies to both sumtypes and structs), they could just return an empty list if there's a problem. 

The misuse should in theory be detected by all the calls to `concretizeType` in Concretize.hs, but for some reason this result is thrown away.

eriksvedang:
More information:

We have a lot of this kind of code in Concretize:

[...]

PR #1228, opened 2021-05-27
fix: Don't crash on invalid member in struct / sumtype

eriksvedang:
A potential fix for #1043 .

Before mergeing: 

- [x] Handle all return values from `concretizeTypeOfXObj`.
- [x] Add test that catches `(deftype (Foo a) [x dsfdfsdf])` mistake
- [x] Add test that catches the bug (using ref as member)


src/Validate.hs
```diff
@@ -67,43 +67,45 @@ okXObjForType typeEnv typeVariables xobj =
...
-    --     ((Foo (f a) b) ...)
-    -- and
-    --     ((Foo (f a) (f b)) ...)
-    -- differ.
-    -- Attempt the first, more restrictive formulation first.
-    struct@(StructTy name tyVars) ->
-      checkVar struct <> checkStruct name tyVars
-    v@(VarTy _) -> checkVar v
-    _ -> Left (InvalidMemberType ty xobj)
+  if isTypeGeneric ty
```
eriksvedang, line 70: This ruins this kind of check: 

```
(deftype (Foo a) [x dsfdfsdf])
```

Need to be more clever!

hellerve (approved):
Looks good to me! 🎉 Thanks for implementing this!

scolsen (approved):
Nice! 👏

PR #1229, opened 2021-05-28
PR: additional local documentation as html 

guberathome:
For new Carp users the locally available documentation (in ./docs/ folder) is a natural 1st place to start looking for help.
However the file format (.md) is really meant to be used with github (e.g. a remote web server) since the local files are missing "features" like links to other documents or embedded graphics / schematic drawings.

The proposed solution is to add a new directory "./docs-html/" with html files, because those can be view with standard means (like Mozilla, lync, ...).
As a 1st version a Python script is provided which generates html documentation on demand.

This is my 1st PR so all comments and suggestions are highly welcome.

(the script is complete, but the PR is still WIP).


guberathome:
@eriksvedang: My pleasure. :-)

The script to generate the html docs (if that's what you mean by "new workflows") is in:
./docs-html/zConvertMd2Html.py

Updated documentation (mostly just a bit) is in:
./docs/index.md 
./docs/LanguageGuide.md 
./docs/Libraries.md 
./docs/Embedded.md 
./docs/core/README.md 

There are open ends which I'd like to tie up, but they should not affect the changes so far.
So you can start reviewing as soon as you like.

eriksvedang:
I meant, what part of the PR triggers the GitHub action?

guberathome:
excellent question:
the only code _change_ is in docs/core/generate_core_docs.carp 
It did not seem to get executed by the local build on my machine, so I'd have to guess.
Sorry.

guberathome:
Ok, that should do it, as far as functionality is concerned.
Please review.

The only thing that I don't like is depending on a 3rd language (e.g. Python) with a non-standard external library.
Fortunately there are tons of libs for Haskell and C available, but I'd prefer to leave that for another PR.

docs/index.md
```diff
@@ -1 +1,26 @@
+# Help index
 
+Please start by reading the [Manual](Manual.md)
+
+You may find yourself coming back to [the description of the Carp language](LanguageGuide.md)
+and to [its libraries.](Libraries.md)
+
+Also there is documentation about:
+[C Interop](CInterop.md);
```
TimDeve, line 9: At the moment this list of topics is rendered in html as one line, could you make it a list instead?

TimDeve:
Thanks for your addition to the docs, it will certainly make them easier to navigate.

I am not as keen about the script, especially the fact that it needs an external dependency that needs to be installed separately. I would assume most people have a markdown reader or server if they want to see the docs rendered locally instead of the hosted version. I'm not completely against it, just not sure it's that useful if someone need to understand they need to pull 'markdown' as a dependency.

docs/index.md
guberathome: good point, thank you. The source (md) is now updated.

Sadly, the python lib does not render it correctly, however cmark-cfm (by github) does.
One more reason to replace it very soon (I'm working on it),

guberathome:
As I said, the python script is a 1st working version (mvp you might say).

It is certainly only "valuable" to those, who are not working online and are aiming at a small setup:
a C compiler, a Carp installation and a web browser (since you're lost without one).
The appeal is meant to be "minimalistic": look, you can get the flexibility of Lisp, the speed of C (provided you are willing to deal with types) and still not need to install several GigaBytes (MS-Visual Studio anyone?).

As soon as you're willing to run your own web server, that would certainly be another target group.
However the concept on "offline documentation for minimalists" should fit the project. 

My next task would be to  replace the python script with a Carp program (demonstrating C Interop) and updating any documentation I happen to pass on the way, if that's ok.

docs/index.md
guberathome: index.md is now rendered correctly in html:
that was a limitation in the python markdown library. took only 2 line breaks to work around it. :-/

eriksvedang:
@guberathome OK, so the python script is not for running on the CI, but rather as a helper script for any user who wants to render the md-files as html? In that case I think it's fine and can be merged, though replacing it with a Carp program in the future will be much nicer of course.

docs/Libraries.md
```diff
@@ -1,6 +1,9 @@
 ## Working with libraries and modules
 
-Carp comes with a standard library called [Core](../core/) where only certain files are loaded by default (see [Core.carp](../core/Core.carp)). If your `CARP_DIR` environment variable is [set properly](https://github.com/carp-lang/Carp/blob/master/docs/Install.md#setting-the-carp_dir), these libraries will be easily imported using the `load` command. For example, to get access to the Bench module, do the following:
+Carp comes with a standard library called Core, sources are [here](../core/).
+Documentation is available [online](http://carp-lang.github.io/carp-docs/core/core_index.html) as well as [locally](./core/core_index.html).
+Many of these - 44 of 62 at the time of writing - are loaded by default, see [Core.carp](../core/Core.carp) for details.
```
eriksvedang, line 5: Keeping this number up do date will be obnoxious, so maybe just skip that part?

eriksvedang (approved):

docs/Libraries.md
guberathome: Done.

PR #1230, opened 2021-06-01
fix: support recursive let bindings in static code (Fix 402)

### LINKED ISSUE #402: Is it possible to dynamically create a recursive function?
For the sake of exercise, I'd like to return a recursive procedure based on some input.

```clojure
(use Int)
(defn a-test [n]
    (let [neato (fn [x]
                  (if (<= x 0)
                    1
                    (* n (neato (- x 1)))))] ;; Error here "couldn't find the symbol 'neato'"
      neato))
```

Evaluating this gives me an error, and I'm unsure if this is intended behavior and/or if there is a better or more idiomatic way to do this.

scolsen:
This PR adds support for recursivity in let bindings in static contexts, a.k.a "let-rec" (e.g. `(let (fn [x] (if (= x 1) x (f (dec x)))))`).

We already had support for this in dynamic carp, however, a few additional changes are needed to permit it in static code, mostly related to augmenting the type system to handle such cases.

I've also added a test case for this functionality.

Fixes #402 (not technically  labelled as a bug, but it implements the behavior described in the issue).


src/Concretize.hs
```diff
@@ -136,6 +136,14 @@ concretizeXObj allowAmbiguityRoot typeEnv rootEnv visitedDefinitions root =
                 SymPath spath name = rootDefinitionPath
                 lambdaPath = SymPath spath ("_Lambda_" ++ lambdaToCName name (envFunctionNestingLevel envWithArgs) ++ "_" ++ show (infoIdentifier ii) ++ "_env")
                 lambdaNameSymbol = XObj (Sym lambdaPath Symbol) (Just dummyInfo) Nothing
+                -- Anonymous functions bound to a let name might call themselves. These recursive instances will have already been qualified as LookupRecursive symbols.
+                -- Rename the recursive calls according to the generated lambda name so that we can call these correctly from C.
+                renameRecursives x = case x of
+                                       (XObj (Sym _ LookupRecursive) si st) -> (XObj (Sym lambdaPath LookupRecursive) si st)
+                                       y@(XObj (Lst _) _ _) -> endoMap renameRecursives y
+                                       y@(XObj (Arr _) _ _) -> endoMap renameRecursives y
+                                       _ -> x
```
hellerve, line 145: I feel like this might be readable and more consistent with other parts of the code base if it pattern matched on the argument instead of using `case`?

src/Concretize.hs
```diff
@@ -136,6 +136,14 @@ concretizeXObj allowAmbiguityRoot typeEnv rootEnv visitedDefinitions root =
...
+                recBody = (endoMap renameRecursives okBody)
```
hellerve, line 146: Internal HLint: the parens here are not needed.

src/Obj.hs
```diff
@@ -1121,3 +1121,10 @@ trueXObj = XObj (Bol True) Nothing Nothing
 -- | Dynamic 'false'.
 falseXObj :: XObj
 falseXObj = XObj (Bol False) Nothing Nothing
+
+-- | Applies an XObj transformation over all atomic XObjs in a form, retaining
+-- list and array structure.
+endoMap :: (XObj -> XObj) -> XObj -> XObj
+endoMap f (XObj (Arr xs) i t) = XObj (Arr (map f xs)) i t
+endoMap f (XObj (Lst xs) i t) = XObj (Lst (map f xs)) i t
```
hellerve, line 1129: Should the internal definitions use some form of `endoMap` as well to enable traversing nested structures or is this an explicit non-goal?

hellerve (commented):
I have not actually played around with this, but the code looks good!

src/Concretize.hs
scolsen: Thanks!

src/Obj.hs
scolsen: Good call, I think the more common use case is applying the function to all terminal points in the form, so it does make sense to recursively call `endoMap` over the lists, thanks!

scolsen:
Thanks for the review @hellerve! Ready for another look whenever.

src/Obj.hs
eriksvedang: Does this mean that `(endoMap square [[1 2 3] [4 [5] 6]])` becomes `[[1 4 9] [16 [25] 36]]` ?

That seems quite unexpected to me... Don't we already have a walk?

scolsen: Yes that's right. We likely do already have this function, but I forget where it's defined and what it's called.

I can revert it back to the old definition, or if we know the walk function I can just remove this entirely and use that.

eriksvedang: I guess the walk is in dynamic code though, and this is just a helper function. I'm not sure what's the right call. Maybe rename this one to walk?!

scolsen: Done!

PR #1232, opened 2021-06-03
clean up IO library

guberathome:
The standard library needed a bit of work:
* BREAKING CHANGE: IO.fwrite now returns the number of written items 
* BREAKING CHANGE: IO.read-file renamed to unsafe-read-file and reimplemented in Carp for cleaner error handling and to demonstrate standard IO functionality in standard lib
* implemented IO.write-file to complement IO.read-file to make interface more symmetric
* implemented (System.error-text) to return the description from last error in call to C standard lib
* String.allocate now handles out-of-memory more robustly. use String.not-allocated? to determine if a "NULL" value was returned
* updated documentation for IO (e.g. indicated which functions are (thin) wrappers for standard C)
* error messages in carp_io.h are output to STDOUT instead of STDIN


eriksvedang:
> To avoid a breaking change

I think breaking changes are totally fine, the goal is to get everything into as nice shape as possible

guberathome:
[...]
> I think breaking changes are totally fine [...]

O_k, here goes...

core/IO.carp
```diff
@@ -3,62 +3,64 @@
...
-  (doc fclose "closes a file pointer.")
+  (doc      fclose "closes a file pointer (thin wrapper for the C standard library).")
   (register fclose (Fn [(Ptr FILE)] ()))
-  (doc fgetc "gets a character from a file pointer.")
+  (doc      fgetc "gets a character from a file pointer (thin wrapper for the C standard library).")
   (register fgetc (Fn [(Ptr FILE)] Char))
-  (doc fwrite "writes to a file pointer.")
-  (register fwrite (Fn [a Int Int (Ptr FILE)] ()) "fwrite")
-  (doc fread "reads from a file pointer into a pointer.")
+  (doc      fwrite "writes a C-string to a file and returns the number of item written items (thin wrapper for fwrite(cstr, item-size, items-count, file) from C standard library). Consider using write-file instead.")
```
scolsen, line 41: typo: s/`item written items`/`written items`

core/IO.carp
```diff
@@ -72,9 +74,47 @@
                                (IO.fclose f)
                                (Result.Success (String.from-chars &r)))))))
 
+  (doc      read-file "returns the contents of a file passed as argument as a string.")
+  (register read-file (Fn [&String] String))
+  (doc  safe-read-file "Reads the content of a file into a (Result String String).\nIt is intended for text files, since the way to determine the length of a String is to use strlen() which probably will be inaccurate for binaries:\n(safe-read-file filename).")
```
scolsen, line 79: Typically our convention is to give the safe variants of functions shorter names, reserving `unsafe-*` to denote unsafe versions (see e.g. `Maybe.unsafe-from`).

Maybe call `safe-read-file` just `read-file` and `read-file` `unsafe-read-file`? Another option is to put the unsafe version in an "Unsafe" module so that one has to call it using `IO.Unsafe.read-file`.

core/IO.carp
guberathome: Thank you, good catch. :-)
Done

core/IO.carp
guberathome: Ok, breaking change it is. It's done now.
Thank you!

guberathome:
Ok, I think it's ready.

Please review.

core/IO.carp
```diff
@@ -72,9 +74,47 @@
...
+      (if (Result.error? &finp?)
+        (Result.Error (fmt "Failed to open file='%s', error-number=%d" filename (Result.unsafe-from-error finp?) ))
+        (let [  finput  (Result.unsafe-from-success finp?)
+                length  (do
+                            (fseek finput 0 SEEK-END)
+                            (let-do [ flength (ftell finput) ]
+                              (rewind finput)   ; aka (fseek ofile 0 SEEK-SET) 
+                              flength ))
+                buffer  (String.allocate length \0 )  ]  
+          (if (String.not-allocated? &buffer)
```
eriksvedang, line 91: why not use `null?` here?

test/io.carp
```diff
@@ -1,6 +1,16 @@
 (load "Test.carp")
 (use-all IO Test)
 
+(defn write-then-read [content file-name]
+  (let [ written? (IO.write-file content file-name) ]
+    (if (Result.error? &written?)
+      (Result.unsafe-from-error written?)
+      (let-do [ read? (IO.read-file file-name) ]
+        (IO.unlink @file-name)
```
eriksvedang, line 9: a little scary to use this in a test...

eriksvedang (commented):
Great!

test/io.carp
eriksvedang: maybe just read the file and check its contents

guberathome: The basic idea was to show that the round-trip works (e.g. writing and reading to/from file does not change anything).

And the test works fine, even on Windows.
Would you still like me to reduce it to reading the file?

core/IO.carp
eriksvedang: OK,fair enough. But better call it just `allocated?` then and wrap the form in `not`.

guberathome: Done.

PR #1233, opened 2021-06-06
refactor: move form validation into a separate module

scolsen:
Historically, the evaluator both checked that incoming Carp forms were
well formed and evaluated them when this was in fact the case. This
complicates the evaluator by giving it multiple responsibilities.

This commit introduces a new module, Forms, that is purely responsible
for the validation of the well-formedness of Carp expressions. The
evaluator calls this module to validate forms before evaluation. This
enables us to modularize our code a bit better. In addition, this commit
introduces pattern synonyms for common forms, simplifying pattern
matching in some cases.

This commit also contains a fix for a previously unreported error in the
evaluator that introduced hangs on forms like `((def foo 5) 5)`.


eriksvedang:
This is very cool! The patterns look great.

eriksvedang:
Seems like some error message for misusing the `ref` form is the reason for the tests failing.

scolsen:
Yeah, I have to add arity checks for a couple more forms, I think, to get all of the error tests to pass. I had to run out earlier so I wanted to at least open it for initial comments since the functional tests all pass as of now.

scolsen:
ok, should be good to go now. The error language has changed slightly since we now just use a generic error for all arity related problems.

eriksvedang (approved):

PR #1238, opened 2021-06-08
fix: fix repl function application evaluation

### LINKED ISSUE #1237: Calling functions in the REPL is broken
```
鲤 (defn f [] 123)
鲤 (f)
I did not understand the form `((defn f [] 123))` at REPL:1:2.

Traceback:
  (f) at REPL:1:1.

```

scolsen:
The previous pattern matching changes introduced a subtle error in the
evaluator's handling of function applications, we used the same Resolver
for both lists and symbols, where previously we hadn't. Restoring the
old resolution selection and adding a new clause to capture forms like
((defn f [] 2)) restores the correct functioning in the repl.

Importantly, forms such as ((defn foo [x] x) 3), should only be resolved
when the defn form is the result of expanding a symbol.

Eventually, resolving defns to anonymous functions might simplify this
case.

Fixes #1237

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -204,6 +204,7 @@ eval ctx xobj@(XObj o info ty) preference resolver =
             (WithPat (SymPat sym path) forms) -> specialCommandWith ctx sym path forms
             (DoPat forms) -> evaluateSideEffects forms
             (WhilePat cond body) -> specialCommandWhile ctx cond body
+            (SetPat iden value) -> specialCommandSet ctx (iden:[value])
             -- This next match is a bit redundant looking at first glance, but
             -- it is necessary to prevent hangs on input such as: `((def foo 2)
             -- 4)`. Ideally, we could perform only *one* static check (the one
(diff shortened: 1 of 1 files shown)


src/Eval.hs
```diff
@@ -397,11 +398,11 @@ eval ctx xobj@(XObj o info ty) preference resolver =
...
       case f' of
-        (ListPat l _) -> go l
-        (SymPat sym _) -> go sym
+        (ListPat l _) -> go l ResolveLocal
+        (SymPat sym _) -> go sym resolver
         _ -> pure (evalError ctx (format (GenericMalformed xobj)) (xobjInfo xobj))
-      where go x =
-              do (newCtx, f) <- eval ctx x preference ResolveLocal
+      where go x resolve =
+              do (newCtx, f) <- eval ctx x preference resolve
```
hellerve, line 405: Are these changes part of the fix? They seem unrelated at a first glance

src/Eval.hs
```diff
@@ -204,6 +204,7 @@ eval ctx xobj@(XObj o info ty) preference resolver =
             (WithPat (SymPat sym path) forms) -> specialCommandWith ctx sym path forms
             (DoPat forms) -> evaluateSideEffects forms
             (WhilePat cond body) -> specialCommandWhile ctx cond body
+            (SetPat iden value) -> specialCommandSet ctx (iden:[value])
```
hellerve, line 207: Is this reordering an unrelated cleanup or important for this change?

src/Eval.hs
scolsen: Yes, they are. I didn't think it'd be relevant either, but if we use `ResolveLocal` on resolution of the `Sym` it results in hangs.

src/Eval.hs
scolsen: Yeah, this was just a small change for readability, I can undo it if we want a cleaner history.

hellerve: No, that’s fine with me, I just wanted to check whether I was missing something!

PR #1240, opened 2021-06-08
refactor: Clean up memory management functions

eriksvedang:
I'm trying to make the memory management easier to understand and hopefully less daunting to contribute to. 

The Memory module still need more work, but since the changes already are significant I thought it would be better to get this merged and continue in a follow-up PR.


src/FindFunction.hs
```diff
@@ -0,0 +1,95 @@
+module FindFunction where
```
scolsen, line 1: Maybe we can move these into the polymorphism module since it's pretty small and these functions all seem to deal with finding the right function to use in a polymorphic context?

eriksvedang: Sure, I'll try that!

scolsen:
Looks great!

Issue #1029, opened 2020-11-28
Improve the Language Guide


eriksvedang:
- [ ] Needs an index at the top
- [ ] A more gentle intro on how to call and define functions would be good, early on the page
- [ ] More links to sub pages with more detailed information (like the Macros doc, etc)
- [ ] Some sections are very sparse compared to the nice descriptions in other headings
- [ ] Lambdas are barely mentioned

guberathome:
Here are some things I miss as a newbie:

- a "Hello Carp" either in Manual.md or in Install.md to test that the installation is working
- a short example using (fmt) and (fstr)
- for each standard library (in "Core Docs") a short summary what it is about
	(e.g. why/when to use an Array vs. StaticArray)
- a polite note saying: "Do not be afraid of the type signatures, there're really usefull and not that hard."
	(e.g. if you try to filter an Array with a simple function, the parameter will be borrowed)
- a polite nugde to go read the documentation.
	e.g. point out in LanguageGuide.md that (info) does not only work for dynamic functions but for libraries & functions as well.
- adjust example for structs using a string
	(e.g. (deftype Person [name string, age Int]
- state that the "void" type is: () 
	(e.g. when describing function signatures)
- a short statement WHERE definitions can be put
	(e.g. I found no way to nest one defn inside another defn)
- a short note which forms return a value or (if that is easier) list forms returning () 
	(e.g. set!)
- a brief note that (and) and (or) 	"Special forms during evaluation of dynamic code" take only 2 arguments 
	(e.g. in contrast to Scheme) (and refer to the documenation of the Bool library for the same reason)
- expand explanation on "named holes" or point to implementation (if that should be self explanatory)
	e.g. does is always have to be "?woot" or what else can be used as a named hole?
- a simpler way to use Carp on Windows (for me Zig does the trick and was pretty easy to set up)	
- a complete list of reader macros (e.g. quasiquoting is missing even from description of macros)
- a list of builtins (e.g. what is not included from carp files)
- an example how to process the "single file with C code". One use case might be to use gcc which insists on having the command line argument for libraries AFTER to source file (to be compiled). This is an issue since we alway need to include libm for the trigonometric functions.

eriksvedang:
@guberathome Thanks a lot for this list – great stuff!

guberathome:
My carp installation on Windows 10 setting was set up pretty quickly:

1. Downloaded Zig for Windows (version 0.8.0) and extracted it to: D:\Tools\p\zig
[...]

PR #1243, opened 2021-06-12
IO.Raw

guberathome:
restructured IO module as suggested by @TimDeve:
all wrappers for the standard C library were moved from IO to IO.Raw


core/IO.carp
```diff
@@ -1,14 +1,78 @@
+; TODO: check if documentation needs parameters listed or if they are in function signature
```
hellerve, line 1: The parameters in the documentation are listed. If you mention them, I suggest you use Markdown syntax (i.e. backticks) to refer to them!

core/IO.carp
```diff
@@ -1,14 +1,78 @@
...
+    (doc      fclose "closes a file pointer (thin wrapper for the C standard library).") 
+    (register fclose (Fn [(Ptr FILE)] Int)  "fclose")  
+    (doc      fclose! "same as (fclose) but to be used as a side effect.") 
+    (defn     fclose! [file]
+      (ignore (fclose file)) ) 
+    (private  fwrite-)
+    (hidden   fwrite-)
+    (register fwrite- (Fn [(Ptr CChar) Int Int (Ptr FILE)] Int) "fwrite")  
+    (doc      fwrite "writes a C-string to a file and returns the number of written items (thin wrapper for fwrite(cstr, item-size, items-count, file) from C standard library). Consider using [`write-file`](#write-file) instead.")
+    (defn     fwrite [data item-size items-count file]
```
hellerve, line 38: Should there be an `fwrite!` utility as well?

core/IO.carp
guberathome: What confused me a bit was: Parameters seem to be listed only for functions, not for templates or external.
So I'll update the doc strings for fopen and fread but not for fseek.

hellerve: This is true, since templates and external functions don’t have argument names that are visible to Carp.

core/IO.carp
guberathome: In general it would make more sense to check whether anything was written (e.g. return value > 0) than whether a file was really closed.
Should I add it anyway?

guberathome: Done

guberathome:
Seems complete to me, please review

TimDeve:
Looks good to me. 
Everything in Raw that has Char as a type should probably be CChar instead, no?
Might be good to do something similar in System as well.

guberathome:
> Everything in Raw that has `Char` as a type should probably be `CChar` instead, no?

Good point!
That would be the IO.Raw.EOF constant as well as the (IO.Raw.get-char) function.
Both are really ints in C e.g. "an unsigned char cast to an int, or EOF on end of file or error".
The current code (e.g. throwing away the EOF sign) feels a bit unsafe to me and does not seem to be used in the Carp code base...


What exactly _is_ the difference between Char and CChar (and where would be a good place to put documentation about that?)
Is guess ```CChar``` is the raw 8-bit type from C, but what can a ```Char``` contain: 8bits, 16 bits (UCS-2?), UTF-32, UTF-8??


So far I did not look into why_testing for EOF in ```IO.read->EOF``` seems to work with ```IO.fgetc``` defined like this:
```
char IO_fgetc(FILE *f) {
    return (char)fgetc(f);  // TODO: handle fgetc()==EOF
}
```
My preference would be to remove what is currently in IO.Raw and have IO.get-char return a (Maybe Char): clean but not really "raw" any longer.

What do you think?

core/IO.carp
```diff
@@ -3,12 +3,77 @@
...
+    (doc      stdin "the standard input file (thin wrapper for the C standard library).")
+    (register stdin (Ptr FILE) "stdin")
+    (doc      stdout "the standard output file (thin wrapper for the C standard library).")
+    (register stdout (Ptr FILE) "stdout")
+    (doc      stderr "the standard error file (thin wrapper for the C standard library).")
+    (register stderr (Ptr FILE) "stderr")
+    (doc      get-char "gets a character from stdin (thin wrapper for getchar() from C standard library).")
+    (register get-char (Fn [] Char) "getchar")  ; TODO: handle EOF?!
+    (doc      exit "exit the current program with a return code (thin wrapper for the C standard library).")
+    (register exit (Fn [Int] ()) "exit")  ; TODO: compare to System.exit for and remove 1 of them!
```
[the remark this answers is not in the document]
eriksvedang: Yeah, let's get rid of this one!

eriksvedang:
The Char and CChar are defined in `core/core.h` as follows:

```
typedef uint32_t Char;
typedef char CChar;
```

eriksvedang:
> My preference would be to remove what is currently in IO.Raw and have IO.fgetc return a (Maybe Char): clean but not really "raw" any longer.

Since this will add some fat to each char (keeping track of the tag in the Maybe) I don't think it's a good idea. So yeah, let's keep the raw stuff very raw and make it nice in the IO module.

guberathome:
Ok, returning a Char for raw input proved difficult (e.g. in IO.read->EOF), since what you _really_ do there is read bytes ignoring the text encoding.
Leaving the raw functionality (e.g. returning Int) is conceptually cleaner and has been tested with UTF-8 input.

core/IO.carp
```diff
@@ -19,81 +84,49 @@
...
 
   (doc      unsafe-read-file "returns the contents of a file passed as argument as a string. Note: there is no way to distinguish the output for an empty file and a missing file!")
   (register unsafe-read-file (Fn [&String] String))
   (doc  read-file "Reads the content of a file into a (Result String String).\nIt is intended for text files, since the way to determine the length of a String is to use strlen() which probably will be inaccurate for binaries.")
   (defn read-file [filename]
-    (let [  finp?  (open-file filename "rb") ]
-      (if (Result.error? &finp?)
-        (Result.Error (fmt "Failed to open file='%s', error-number=%d" filename (Result.unsafe-from-error finp?) ))
-        (let [  finput  (Result.unsafe-from-success finp?)
+    (let [  finput?  (open-file filename "rb") ]
```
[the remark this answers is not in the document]
guberathome: done

core/IO.carp
```diff
@@ -102,26 +135,27 @@
...
-  (register getenv- (Fn [String] (Ptr CChar)) "getenv")
+  (register getenv- (Fn [(Ptr CChar)] (Ptr CChar)) "getenv")
   (doc  getenv "gets the value of an environment variable (Carp-style wrapper for the C standard library)")
   (defn getenv [s]
-    (let [e (getenv- s)]
+    (let [e (getenv- (String.cstr s))]
       (if (null? e)
         (Maybe.Nothing)
-        (Maybe.Just (from-cstr e)))))
+        (Maybe.Just (from-cstr e)) )))
```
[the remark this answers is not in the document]
guberathome: That's a personal preference: it separates the brackets for code in the current line and from other indentation levels.
It's removed now to have the code look consistently.

core/IO.carp
```diff
@@ -19,81 +84,49 @@
...
+      (match file?
         (Result.Error x) (Result.Error x)
         (Result.Success f) (let [c (zero)
                                  r []]
                              (do
                                (while (do (set! c (IO.fgetc f))
-                                          (/= c IO.EOF))
-                                 (set! r (Array.push-back r c)))
-                               (IO.fclose f)
+                                          (/= c IO.Raw.EOF))                                 (set! r (Array.push-back r c)))
```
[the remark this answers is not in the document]
guberathome: done

core/IO.carp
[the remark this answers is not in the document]
guberathome: Done

core/IO.carp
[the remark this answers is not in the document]
TimDeve: Ok, thanks

TimDeve:
Think you need to update the SDL examples as well.

Off topic: Did we not update the Windows CI to have the SDL tests? I thought for sure I did...

core/System.carp
```diff
@@ -56,6 +56,9 @@
   (register EROFS Int "EROFS")
   (register EINTR  Int "EINTR")
 
-  (doc exit "exits the program with the supplied exit code (thin wrapper for the C standard library).")
-  (register exit (Fn [Int] ()) "exit")
+  (doc      exit! "exits the program with the supplied exit code (thin wrapper for the C standard library).")
+  (register exit! (Fn [Int] ()) "exit")
```
TimDeve, line 60: I would actually just keep the `exit` one, while the signature is slightly inaccurate, it means you can exit whenever, and it doesn't matter what it returns because it's never used.

TimDeve (changes requested):

core/System.carp
guberathome: Done.

Frankly I don't like the signature of (System.exit) but there's a limit on how long I want to spent on this topic.

TimDeve: Fair enough.

guberathome: Sorry, it's not resolved yet. I stumbled upon the original reason for changing the interface:

With the original (System.exit) I get linker errors on Windows (about missing WinMain):
1) llvm:
```
LINK : fatal error LNK1561: Einstiegspunkt muss definiert werden.
clang: error: linker command failed with exit code 1561 (use -v to see invocation)
```

2) gcc:
```
D:/.../lib/libmingw32.a(lib64_libmingw32_a-crt0_c.o):crt0_c.c:(.text.startup+0x2e): undefined reference to `WinMain'
collect2.exe: error: ld returned 1 exit status
```

Using ```(ignore (System.exit...))``` does not improve things, now Carp compilation fails with:
```

I found an unresolved generic type `t2` for the expression `_` at line 220, column 22 in 'D:/Tools/p/carp/core/Macros.carp'
```

So: how does this look for a solution?
```Clojure
  ; keep (System.exit) as it is
  (doc      exit_ "same as exit, but avoids some obscure errors on Windows (e.g. undefined reference to `WinMain'")
  (register exit_ (Fn [Int] ()) "exit")  
```

TimDeve: Ok, that's interesting. For my curiosity do you still get linker errors if you do `(ignore (the Int (System.exit 0)))` instead?

I'm reticent having two slightly the same `exit`, there should only be one. Maybe keeping the `System` changes out of this PR would make your life easier.

guberathome: > Ok, that's interesting. For my curiosity do you still get linker errors if you do `(ignore (the Int (System.exit 0)))` instead?
Now _that_ works like a charm.

> I'm reticent having two slightly the same `exit`, there should only be one. Maybe keeping the `System` changes out of this PR would make your life easier.
Let me take a look at the SDL examples to see if we can't just have a non-returning exit function, e.g. (Fn [Int] () ).

TimDeve: So it seems we have a bug where if you pass and a function with a generic return at the end of main, we don't generate the main function at all. I would keep the System changes out of this PR and see if we can solve the problem separately. @eriksvedang What do you think?

eriksvedang: Ah, I see. Yeah that would make `main` generic which makes it not be emitted. Seems like a separate issue yes (where we should detect that `main` has become generic and warn about that.)

guberathome: > Ah, I see. Yeah that would make `main` generic which makes it not be emitted. Seems like a separate issue yes (where we should detect that `main` has become generic and warn about that.)

Oh well, I opened request #1246 to handle the problem separately.

However I would prefer wrapping the code in Main in an implicit ```(ignore (the Int ....))``` over displaying a warning.

TimDeve: Warning makes more sense because the two thing we want to accept for `main` is either `()` which means an exit code of 0, or explicitly return an exit code as an `Int`.

It seems the reason you're getting the error is because you're trying to do something like `(System.exit 42)` when you can just replace that with `42` to get the same behaviour.

guberathome: > Warning/Error makes more sense because the two thing we want to accept for `main` is either `()` which means an exit code of 0, or explicitly return an exit code as an `Int`.

Ah, thanks for the pointer
I added it to the list at #1029.

guberathome:
Ok, that should do it.

eriksvedang (approved):
Nice work!

PR #1244, opened 2021-06-12
docs: document managed? primitive

hellerve:
This PR documents the primitive `managed?`.

Cheers


TimDeve:
>  ...because it references an OS resource that needs a cleanup action, like a file or socket.

Not sure if this is in scope for this PR but could we remove this line from the Memory doc and point people to drop instead? Implementing `delete` is super unsafe for type that are already managed.

PR #1247, opened 2021-06-15
fix: check for generic main at build time

### LINKED ISSUE #1246: returning generic (?) function from main causes linker errors
Use this to reproduce the problem:
```
(defmodule NewSystem
  (deftemplate exit (Fn [Int] a) "$a $NAME(int code)" "$DECL { exit(code); }"))

(defn abort [error-code message]
  (do
    (IO.errorln message)
  error-code ))
(defn finish [message]
  (do
    (IO.println message)
  0 ))

(defn process [s1 s2]
  (if (String.= s1 s2)
    (abort 2 "come on, those parameters should not match!")
    (finish  "After lenghty computation we have reached a verdict...") ))

(defn main []
 ; (ignore (the Int
   (NewSystem.exit
    (let [ args (Int.- (StaticArray.length &System.args) 1) ]    ; 1st argument is executable!
    (if (not (Int.= 2 args))
      (abort 1 &(fmt "expecting exactly 2 arguments (got %d)." args))
      (process (StaticArray.unsafe-nth &System.args 1) (StaticArray.unsafe-nth &System.args 2)) ))))
  ; ))
 ```
 Here's the error output on Windows using gcc:
 ```
D:/.../x86_64-w64-mingw32/lib/../lib/libmingw32.a(lib64_libmingw32_a-crt0_c.o):crt0_c.c:(.text.startup+0x2e): undefined reference to `WinMain'
collect2.exe: error: ld returned 1 exit status
carp: callCommand: gcc  -o out\Untitled -I D:\Tools\p\carp/core/  -D_CRT_SECURE_NO_WARNINGS out\main.c (exit 1): failed
 ```

### PR DESCRIPTION
**Pull request by hellerve**
This PR fixes #1246 by inserting a few checks in `commandBuild`.

```
> (defn main [] +)
> :b
`main` was generic at build time, it must return either `Int` or `()`.

Traceback:
  (build) at REPL:1:1.

> (def main 1)
[WARNING] Definition at line 1, column 2 in 'REPL' changed type of 'main' from (Fn [] (Fn [a, a] a)) to Int
> :b
`main` is expected to be a function, but was `Int`.

Traceback:
  (build) at REPL:1:1.
```

Cheers

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -342,7 +342,11 @@ commandBuild ctx [XObj (Bol shutUp) _ _] = do
         if generateOnly
           then pure (ctx, dynamicNil)
           else case Map.lookup "main" (envBindings env) of
-            Just _ -> compile True
+            Just (Binder _ (XObj _ _ (Just (FuncTy _ IntTy _)))) -> compile True
+            Just (Binder _ (XObj _ _ (Just (FuncTy _ UnitTy _)))) -> compile True
+            Just (Binder _ (XObj _ _ (Just (FuncTy _ _ _)))) -> pure (evalError ctx "`main` was generic at build time, it must return either `Int` or `()`." Nothing)
+            Just (Binder _ (XObj _ _ (Just t))) -> pure (evalError ctx ("`main` is expected to be a function, but was `" ++ show t ++ "`.") Nothing)
+            Just (Binder _ (XObj _ _ Nothing)) -> pure (evalError ctx ("The type of `main` is not set at build time. This should not happen, consider raising an issue with the code that caused it.") Nothing)
             Nothing -> compile False
   where
     removeSpecials env =
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -344,6 +344,7 @@ commandBuild ctx [XObj (Bol shutUp) _ _] = do
           else case Map.lookup "main" (envBindings env) of
             Just (Binder _ (XObj _ _ (Just (FuncTy _ IntTy _)))) -> compile True
             Just (Binder _ (XObj _ _ (Just (FuncTy _ UnitTy _)))) -> compile True
+            -- we do not check for non-generic versions of main, these are already caught during concretization
             Just (Binder _ (XObj _ _ (Just (FuncTy _ _ _)))) -> pure (evalError ctx "`main` was generic at build time, it must return either `Int` or `()`." Nothing)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Commands.hs [hunk 1]
```diff
@@ -342,7 +342,11 @@ commandBuild ctx [XObj (Bol shutUp) _ _] = do
         if generateOnly
           then pure (ctx, dynamicNil)
           else case Map.lookup "main" (envBindings env) of
-            Just _ -> compile True
+            Just (Binder _ (XObj _ _ (Just (FuncTy _ IntTy _)))) -> compile True
+            Just (Binder _ (XObj _ _ (Just (FuncTy _ UnitTy _)))) -> compile True
+            Just (Binder _ (XObj _ _ (Just (FuncTy _ _ _)))) -> pure (evalError ctx "`main` was generic at build time, it must return either `Int` or `()`." Nothing)
```
scolsen, line 347: nit: the return type here could be non-generic, e.g. `String`. Maybe change the language to:

``main` did not return an Int or () at build time. It must return either `Int` or `()` but returns: show retT`

scolsen (commented):
One small nit, but looks good otherwise!

src/Commands.hs
hellerve: The non-generic case is actually already handled, as detailed in #1246, by the concretization error.

scolsen: Ah, right—that makes sense. I worry that it's easy to forget that prior checks are performed  before we reach this code, which might lead someone to duplicate the code by adding an additional check (at least that's what I would have done if you didn't point this out!). Maybe we can add a one-line comment explaining this for now? Likely this indicates we should eventually move the check into concretize...but that's not great either imo, since I don't think concretize should really care that a function happens to be called `main`—that feels like we're sliding into ambiguous separation of concerns. Maybe all the checks that are in concretize should be moved here.

hellerve: That’s a good point—I actually just checked and saw that concretization had a check for generic types in `main`. Removing that check sadly breaks the compiler, since it enforces that concretization happens in a single pass, essentially. So I think the problem here is that concretization can‘t enforce specialization, which is why we need the special cases in the build commands. Catching what we can as early as possible (and concretization happens before building) makes sense to me, though, especially in the REPL. What do you think?

Thanks for the input!

hellerve: I did add a comment as you suggested, btw!

scolsen: Ah, gotcha—catching early makes sense too. Comment looks great, thank you!

scolsen (approved):

PR #1248, opened 2021-06-15
Up down case

guberathome:
added simple case conversion for ASCII characters

### DIFF
--- a/core/String.carp
+++ b/core/String.carp
@@ -1,4 +1,5 @@
 (system-include "carp_string.h")
+(system-include "ctype.h")
 
 (defmodule String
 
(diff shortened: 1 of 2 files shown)


core/String.carp
```diff
@@ -169,6 +170,19 @@
 
   (doc contains? "Checks whether the string `s` contains the character `c`.")
   (defn contains? [s c] (> (index-of s c) -1))
+
+  (private  tolower-)
+  (hidden   tolower-)
+  (register tolower- (Fn [Byte] Byte) "tolower")
+  (private  toupper-)
+  (hidden   toupper-)
+  (register toupper- (Fn [Byte] Byte) "toupper")
```
hellerve, line 179: I would argue that this shouldn’t be part of `String`, but may be better suited for the `Byte` module. It could still be `hidden`, just not `private`.

guberathome: Yes, this is the _ugly_ bit, so I tried to hide it as much as possible...

But since it's about converting characters, wouldn't the Char module be a better fit?

hellerve: In a way that would make sense, but I would go by the type here usually. We could make this a function that operates on `CChar`, but that would probably just complicate matters even more.

guberathome: done

hellerve (approved):

core/String.carp
hellerve: Thanks!

test/string.carp
```diff
@@ -323,4 +323,12 @@
                 &[\� \�]
                 &(chars &(from-bytes &[255b 255b]))
                 "check for invalid UTF-8 sequences")
+  (assert-equal test
+                "hällo wörld"
+                &(ascii-to-lower "HälLo WöRld")
+                "ascii-to-lower works for valid UTF-8" )
```
scolsen, line 329: I find the test case here a bit confusing. The function purportedly only works on ASCII values, but we test the function against extended ASCII values here and mention UTF8 values in this test?

guberathome: Well hopefully not _that_ confusing:
The sources are encoded in UTF-8, so only the ASCII-compatible bits are affected by the transformation, the rest is not changed. It's really just the underlying C library at work (with all its quirks).

scolsen: Ah, that makes sense, just threw me for a loop on a first skim. Thanks!

Issue #1249, opened 2021-06-16
Enable documentation emission for top-level bindings


eriksvedang:
Currently you can only specify modules to generate docs for, so top-level bindings are left out (which is a shame, especially when they have nice documentation.)

The whole situation is a bit tricky because we can't just treat the global env as a big module – that would emit *everything*  (and that's not what you want if you make a library, for example).

PR #1253, opened 2021-06-16
feat: Emit docs for top level bindings

eriksvedang:
Here's a rough implementation for solving #1249.

It allows you to generate docs the following way:

```
(save-docs-ex ['SomeModule 'SomeOtherModule] ["Macros.carp" "Int.carp"])
```

This would create the normal docs for SomeModule and SomeOtherModule, but also create a faux module called "Global" which contains all global bindings in "Macros.carp" and "Int.carp".

Ideas for improvements are welcome!


hellerve:
Does filtering through the whole global environment for eligible bindings impose a significant performance overhead?

eriksvedang:
@hellerve The search only happens when one saves the docs, so it's not super critical to be fast. My experience was that it's more or less instantaneous though.

eriksvedang:
I changed the command for saving docs to take two arrays, to enable better programatic support. The old macro `save-docs` still auto quotes its input.

PR #1256, opened 2021-06-18
use MD4C C-library to convert Markdown to html

guberathome:
* removes the python library for conversion of markdown to html
* needs PR #1257 to work

Issue #1226, opened 2021-05-26
Ability to configure how are arguments are passed to external compiler


eriksvedang:
Different compilers require different flags, order of arguments, etc.
It would be super handy to be able to define this per compiler/platform, something like:

```clojure
(defbackend osx-tcc
  {"compiler" "tcc"
   "lib" "-l%"
   "include" "-i%"})
```

(very rough sketch)

guberathome:
There already seems to be a project setting:
```
(Project.config "libflag" "...")
```

If these were to be added _after_ the file name in the compiler command, the major obstacle for using gcc should be gone.
Not having to configure profile.carp for this would indeed be _nice_to_have.

eriksvedang:
Yeah, basically this would move all of those settings into a data structure and make it easy to switch between different sets of them.

guberathome:
Just to illustrate my current issue with gcc (on Linux), which might provide additional motivation for the suggested solution:

Supposing you had any Carp source file like this...
```
; content of hello.carp 
(Project.config "echo-compiler-cmd" true)
(Project.config "compiler" "gcc")

(defn main []
	(IO.println &(fmt "Hello there: it took us only %d iterations to approximate e=%f" 123 2.71828 )) )
```

... then  the current output from Carp fails since ld can not link functions from libmath (or libm):
```
gcc  -o out/Untitled -I /home/ptokremin/.local/carp//core/  -fPIC -g -std=c99 -D_DEFAULT_SOURCE -Wall -Werror -Wno-unused-variable -Wno-self-assign -lm out/main.c
/usr/bin/ld: /tmp/ccisakli.o: in function `Double_acos':
/home/ptokremin/.local/carp//core/carp_double.h:68: undefined reference to `acos'
/usr/bin/ld: /tmp/ccisakli.o: in function `Double_asin':
/home/ptokremin/.local/carp//core/carp_double.h:72: undefined reference to `asin'
[...]
/usr/bin/ld: /tmp/ccisakli.o: in function `Float_ceil':
/home/ptokremin/.local/carp//core/carp_float.h:123: undefined reference to `ceilf'
/usr/bin/ld: /tmp/ccisakli.o: in function `Float_floor':
/home/ptokremin/.local/carp//core/carp_float.h:127: undefined reference to `floorf'
/usr/bin/ld: /tmp/ccisakli.o: in function `Float_mod':
/home/ptokremin/.local/carp//core/carp_float.h:131: undefined reference to `fmodf'
[...]

PR #1262, opened 2021-06-22
declare to use c11 

guberathome:
feat: builds on OS other than Windows now use ```-std=c11``` instead of ```-std=c99```

The use case is to be able to forward declare functions which use the String type, which in turn needs to be defined and thus redefined (e.g. PR #1256).
This causes a warning (and thus failed builds) on Linux when ```-std=c99``` is supplied to the C compiler.
The problem is gone with C11, which explicitly permits this.

This has been tested on Linux with Clang, tcc (which ignores the parameter), and Zig.
(note: gcc supports the flag, but needs the ```-lm```  flag positioned after the source file name) which is separate issue).

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -30,7 +30,7 @@ defaultProject =
         _ ->
           [ "-fPIC",
             "-g",
-            "-std=c99",
+            "-std=c11",
             -- , "-pedantic"
             "-D_DEFAULT_SOURCE",
             "-Wall",
@@ -157,13 +157,14 @@ main = do
   pure ()
 
 -- | Options for how to run the compiler.
-data FullOptions = FullOptions
-  { optExecMode :: ExecutionMode,
-    optOthers :: OtherOptions,
-    optPreload :: [String],
-    optPostload :: [String],
-    optFiles :: [FilePath]
-  }
+data FullOptions
+  = FullOptions
+      { optExecMode :: ExecutionMode,
+        optOthers :: OtherOptions,
+        optPreload :: [String],
+        optPostload :: [String],
+        optFiles :: [FilePath]
+      }
   deriving (Show)
 
 parseFull :: Parser FullOptions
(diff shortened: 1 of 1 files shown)


guberathome:
Ready for review. :-)

eriksvedang:
I'm not good at C versions pros/cons. My gut feeling is that we should stay as old as possible to increase the likelihood of people being able to compile the C code we emit/supply in core. 

Ergonomics for us (people working on the Carp core libs) is not a priority – ideally we only use very simple C features to avoid any trouble when people want to build on weird platforms with ancient compilers. 

This is not a good trade off in most projects but is important when using the language mainly as a code generation target. Basically we take the hit of readability and maintainability as a service to the Carp users.

Maybe @jacereda (or someone else!) has some input on why we set it to c99, and if going up to c11 is a good or bad idea.

guberathome:
Fair enough, but this is currently a show stopper for #1256.

Just to clarify:
* this is _not_ about generating code use features from C11, only to declare it's usage as valid.
* GCC, CLang and Zig explicitely support C11, TCC ignores it.
* MSVC is not relevant, since Windows is treated separately.
* this should no longer be relevant, once PR #1226 is implemented.

Let me describe the use case a bit more, maybe I'm missing something:
1) In ./docs-html/lib/Md2Html.c:137 the result has been calculated in a buffer and is to be returned: 
```result = String_from_MINUS_cstr( buf_out.data );```
2) The result struct (containing the buffer) is not compatible with Carp and needs to be free, so the result has to be copied.
3) Returning a raw ```(Ptr CChar)``` and then using ```(String.from-cstr)``` would internally call ```(String-from-cstr)``` and thus copy again. This is not desirable.
4) Thus the C library uses ```(String.from-cstr)``` directly, which is fine as far as memory management in Carp is concerned.
5) In order to do that, the Carp function needs to be forward declared, code is in ./docs-html/lib/Md2Html.h:6
6) That in turn need to know the String type, which is defined in ./docs-html/lib/Md2Html.h:5
7) So effectively String gets defined twice, which is fine in C11, but raises a warning in C99, which due to the C compiler settings (for os/=Windows) gets promoted to an error.

What would be the clean Carp way to solve this?

eriksvedang:
Could the Md2Html.h import carp's Core.h?

eriksvedang:
Alternatively, avoid depending on Carp functions in the C code.

guberathome:
Nah, much simpler: renamed the String type to Carp_String, thus its no longer duplicated.

Let's scratch this.

guberathome:
This feature poses some risks and is in essence a bit misleading, since we don't actually emit code for C11.
Further PR #1256 has been solved without it.

eriksvedang:
Ok, great!

Issue #1263, opened 2021-06-26
tcc support on Windows


guberathome:
Motivation: tcc seems like a good way to speed up the Carp repl, which can feel a bit sluggish at times.
This works fine under Linux but on Windows things are more difficult.

Suppose we had a simple Carp program with UTF-8 output:
```
(defn main []
	(IO.println &(fmt 
		"Hällo there: üt took us only %d iterations to approximate e=%f" 
		123 
		(Double.exp 0.999999) )) )
```

In principle this works, e.g. using gcc+glibc:
```
D:\tmp\bastel>carp -x hello.carp
gcc  -o out\Untitled -I D:\Tools\p\carp/core/  -D_CRT_SECURE_NO_WARNINGS out\main.c
Hello there: it took us only 123 iterations to approximate e=2.718280
```

Let's verify that msvc and the native libC support this too
(profile.carp checks CARP_CC and sets Project.config accordingly):
```
D:\tmp\bastel>set carp_cc=msvc

D:\tmp\bastel>carp -x hello.carp
cl  -o out\Untitled -I D:\Tools\p\carp/core/  -D_CRT_SECURE_NO_WARNINGS out\main.c
Microsoft (R) C/C++-Optimierungscompiler Version 19.16.27045 für x64
Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.

cl : Befehlszeile warning D9035 : Die Option "o" ist veraltet und wird in einer der nächsten Versionen entfernt.
main.c
Microsoft (R) Incremental Linker Version 14.16.27045.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:main.exe
/out:out\Untitled.exe
main.obj
Hello there: it took us only 123 iterations to approximate e=2.718280
```

Great, let's turn to tcc:
```
D:\tmp\bastel>tcc -version
tcc version 0.9.27 (x86_64 Windows)
D:\tmp\bastel>set carp_cc=tcc

D:\tmp\bastel>carp -x hello2.carp
tcc       -o out\Untitled -I D:\Tools\p\carp/core/  -D_CRT_SECURE_NO_WARNINGS out\main.c
out/main.c:3426: error: 'CP_UTF8' undeclared
carp: callCommand: tcc    -o out\Untitled -I D:\Tools\p\carp/core/  -D_CRT_SECURE_NO_WARNINGS out\main.c (exit 1): failed
```

[...]

PR #1266, opened 2021-06-30
to be deleted (wrong branch): TCC on windows 32bit

guberathome:
TCC is an incredibly fast C compiler.
Unfortunately on Windows the latest TCC release 0.9.27 does not work out of the box, since 2 minor definitions are missing to compile a simple HelloWorld.Carp.

* related to issue #1263, but does not generally fix it.


hellerve:
This looks great! However it looks like some stray commits made their way into this branch! I assume only `d421d8c` is relevant here?

guberathome:
Oops, that was not branched of the master branch as intended, hence the "stray commits". 
Let me fix that.

guberathome:
branching error, redo from start

Issue #1264, opened 2021-06-30
pattern matching too greedy


guberathome:
Assume we want to match markdown references like ```[description](URl)```.
Note there is no character between both parts (description and URl).

The current implementation delivers false matches, when the parts are separated by whitespace:
```
Welcome to Carp 0.5.1
This is free software with ABSOLUTELY NO WARRANTY.
Evaluate (help) for more information.
> (Pattern.match-all-groups  #"\[(.-)\]\((.-)\)"  "icke [dette] (xyz) [kieke]\n\t(...) [mal](oochen,) fleesch un beene")
[[@"dette] (xyz) [kieke]
        (...) [mal" @"oochen,"]]
>
```

The correct results would be:
```
Python 3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 03:37:03) [MSC v.1900 64 bit (AMD64)] on win32
Type "copyright", "credits" or "license()" for more information.
>>> import re
>>> pat = re.compile("\[(.*?)\]\((.*?)\)")
>>> re.findall( pat, "icke [dette] (xyz) [kieke]\n\t(...) [mal](oochen,) fleesch un beene" )
[('mal', 'oochen,')]
```

guberathome:
Since carp_pattern.h is derived from lua, let's check if they have the same problem in the newest version:
```
Lua 5.4.3  Copyright (C) 1994-2021 Lua.org, PUC-Rio
> cap1, cap2 = string.match( "icke [dette] (xyz) [kieke]\n\t(...) [mal](oochen,) fleesch un beene", "%[(.-)%]%((.-)%)" )
> cap1
dette] (xyz) [kieke]
	(
	...) [mal
> cap2
oochen,
```
So the problem persists, updating to a newer lua version is not a viable path.

hellerve:
I don’t think this is a bug, since the matches are valid in a non-backtracking pattern system (note that this is not the same as a regex, which does backtrack, which is required to satisfy “as few characters matched as possible” with the preceding matches in mind).

Now, a simple way to fix this would be to make the requirement “do not match brackets” explicit, like so:

```clojure
> (Pattern.match-all-groups  #"ups  #"\[([^\[\]]-)\]\((.-)\)"  "icke [dette] (xyz) [kieke]\n\t(...) [mal](oochen,) fleesch un beene")
[[@"mal" @"oochen,"]]
```

[...]

PR #1268, opened 2021-07-01
docs: add pattern limitations

hellerve:
This PR adds some documentation on the limits of patterns, as discussed in #1264. It’s very terse, but hopefully explains enough to make sense.

Cheers

### DIFF
--- a/docs/LanguageGuide.md
+++ b/docs/LanguageGuide.md
@@ -499,7 +499,7 @@ Here is a little overview of the API:
 
 #### Limitations of Patterns
 
-As mentioned above, patterns are not as expressive as Regular Expressions. The
+As mentioned above, patterns are not as expressive as regular expressions. The
 fundamental difference is that patterns do not backtrack. This means that they
 cannot express alternation (because we can’t go back to where we branched) and
 we cannot reduce non-greedy matches on the left. The latter point might not be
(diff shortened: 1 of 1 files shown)


hellerve:
CCing @guberathome; does this explanation make sense?

docs/LanguageGuide.md
```diff
@@ -496,3 +496,22 @@ Here is a little overview of the API:
...
+we cannot reduce non-greedy matches on the left. The latter point might not be
+obvious, so let us look at an example:
+
+```
+(Pattern.match-all-groups #"1.-2" "1 1 2") ; => [[@"1 1 2"]]
+```
+
+A valid, less greedy match would have been `"1 2"`, but since it would have
+required us to go back to the left after we had started matching to reduce the
+match size, this is not done. As such, while `-` is similar to `*?`, it is not
```
guberathome, line 514: Great, that certainly helps!

I'd add 
```
is similar to `*?` from RegularExpressions
```
to point out, where this syntax originated and where it can (not) be used.

hellerve: Thanks for the feedback, done!

hellerve:
It looks like the Windows CI build failed due to a network issue again.

eriksvedang:
Very nice!

PR #1269, opened 2021-07-02
feat: add machine-info primitive draft

hellerve:
This PR adds `machine-info`, meant as a verison of `info` for machines or programs. The name and behavior are up for debate, but I wanted to provide a prototype to get the conversation going.

## The Proposal

A machine-readable version of the compiler-visible information associated with a symbol would be great. `machine-info` attempts to provide that by returning a list of the following form: `(type|(), info|(), metadata)`, where `type` is an S-expression of the type, `info` is a tuple of `(file, line, column)`, and `metadata` is a list of key-value pairs.

## The Value

Allowing for this would, among other things, allow us to express `info` as a Carp expression. It would also allow us to inquire IDE-usable informations from within a REPL.

## Open Questions

Of course the name and format of the return value are questions of taste. I don’t particularly like the name and am open to suggestions. As for the return value, I’d like the format to be easily traversable, so the sublists for optional information made sense. As for the ordering I don’t have any opinions, and if additional info could or should be provided let me know.

Any and all feedback welcome!

Cheers

### DIFF
--- a/src/Obj.hs
+++ b/src/Obj.hs
@@ -9,7 +9,6 @@ import Control.Monad.State
 import Data.Char
 import Data.Hashable
 import Data.List (intercalate)
-import Data.List.Split (splitOn)
 import Data.Maybe (fromMaybe)
 import GHC.Generics (Generic)
 import Info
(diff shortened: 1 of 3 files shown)


eriksvedang:
Great idea! I'll think about the name...

eriksvedang:
The current name fooled me into thinking it was about the host machine somehow, so it's a bit misleading in that way.

`structured-info` maybe?

hellerve:
Works for me!

src/Obj.hs
```diff
@@ -1150,3 +1151,42 @@ walk :: (XObj -> XObj) -> XObj -> XObj
 walk f (XObj (Arr xs) i t) = XObj (Arr (map (walk f) xs)) i t
 walk f (XObj (Lst xs) i t) = XObj (Lst (map (walk f) xs)) i t
 walk f x = f x
+
+tyToObj :: Ty -> Obj
```
scolsen, line 1155: Ideally, I think we could define this as:

```haskell
tyToObj = xobjObj . reify
```

And perform further transformations (if any are needed) downstream. (e.g. calling `simpleSym`)

scolsen: See `Reify.hs`

scolsen (commented):
Nice! You might be able to simplify the definition of `tyToObj` since we have a function that performs the type to xobj conversion, but other than that looks great!

src/Obj.hs
scolsen: We'd also have to move the definition of this function into `Reify` to define it this way (to avoid import cycles)—but that seems ok to me since the responsibility of `Reify` is to provide conversions from internal Type representations to AST things

hellerve: Actually, I just completely forgot about `reify`, and the usecase can be expressed completely in terms of it. This means I can get rid of this entirely! Thanks!

hellerve: Fixed in 04588784

scolsen: Nice! Thanks!

scolsen (approved):

eriksvedang:
Is this ready?

hellerve:
Sure is!

PR #1270, opened 2021-07-03
fix: set correct info on defmodule

hellerve:
This PR fixes the info set on `defmodule`.

Example:

```
> ; before
> :i Quasiquote
...
  Defined at line 0, column 0 in 'dummy-file'
...
> ; after
> :i Quasiquote
...
  Defined at line 5, column 12 in '/Users/veitheller/Documents/Code/Github/carp/carp/core/Quasiquote.carp'
...
```

Cheers


eriksvedang:
Nice catch!

PR #1273, opened 2021-07-06
docs: set infoFile on primitives and commands

hellerve:
This PR sets the `infoFile` property on commands and primitives in the compiler. This mostly means that `info` is changed:

```
> ; before
> :i Dynamic.info
Dynamic.info : Dynamic
  Defined at line 0, column 0 in '<dummy file>'
[...]
> ; after
> :i Dynamic.info
Dynamic.info : Dynamic
  Defined at line 0, column 0 in 'Core Primitives'
[...]
```

This is mostly so people know where things come from.

Cheers

### DIFF
--- a/src/Commands.hs
+++ b/src/Commands.hs
@@ -64,7 +64,7 @@ addCmd path callback doc example =
               XObj (Arr args) Nothing Nothing
             ]
         )
-        (Just dummyInfo)
+        (Just dummyInfo {infoFile = "Core Commands"})
         (Just DynamicTy)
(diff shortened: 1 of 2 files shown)


hellerve:
I don’t know why whitespace changes keep appearing in my commits, I think this might be `ormolu` at work.

eriksvedang:
Awesome! This could perhaps be used to generate a api docs page for the commands.

hellerve:
They are already created in [`Dynamic`](https://carp-lang.github.io/carp-docs/core/Dynamic.html)!

eriksvedang:
Ah, ok 👍

PR #1276, opened 2021-07-06
feat: allow for multibranches in case

hellerve:
This PR changes `case`—backwards-compatible change!—to allow for what I call multibranches, i.e. branches that are reachable by multiple values. It achieves this by using list syntax:

```clojure
(case (+ 10 1)
  (11 12) (println* "its either 11 or 12")
  (println* "its something else")
 )
```

I also took the liberty of optimizing it—and fixing a potential problem with side effects—using `gensym`. Before, something like the above would desugar to:

```clojure
(if (or (= (+ 10 1) 11) (= (+ 10 1) 12) ; => (+ 10 1) is duplicated here, side effects would potentially be executed twice
  (IO.println (ref (str "its either 11 or 12")))
  (IO.println (ref (str "its something else"))))
```

Now, using `gensym` we can write:

```clojure
(let [gensym-generated1037 (+ 10 1)]
  (if (or (= gensym-generated1037 11) (= gensym-generated1037 12))
    (IO.println (ref (str "its either 11 or 12")))
    (IO.println (ref (str "its something else")))))
```

i hope this all makes sense!

Cheers

### DIFF
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -169,7 +169,9 @@ Example:
 )
 ```")
 (defmacro case [form :rest branches]
-  (case-internal form branches))
+  (let [name (gensym)]
+    (list 'let [name form]
+      (case-internal name branches))))
 
 (defmodule Dynamic
   (doc flip
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -149,20 +149,22 @@ Example:
       (macro-error "case has even number of branches; add an else branch")
       (if (= (length xs) 1)
         (car xs)
-        (list 'if
-              (list '= name (car xs))
-              (cadr xs)
-              (case-internal name (cddr xs)))))))
+        (let [values (if (list? (car xs)) (car xs) (list (car xs)))]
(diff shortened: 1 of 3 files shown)


eriksvedang:
Cool! But doesn't case work in dynamic carp? In that case this interferes a bit with matching on actual lists, no? Maybe a more clear and (slightly) safer design would be to have the matching case be a list starting with `or`, eg `(or 10 11)`?

hellerve:
Mhm, this definitely introduces a problematic case I hadn’t thought about (could be rectified by wrapping the list again, but that’s a little iffy). The thing with `or` is that it shifts the problem around, since I can see a case where people want to match  against a list with `or` in the  first position as well. We can document the double-wrapping technique in the docstring? I’m also happy to change it to `or`, I just don’t know how much it helps.

eriksvedang:
Yeah, tricky! Maybe take a hint from :rest for rest args and use :or (if that even parses correctly?)

hellerve:
Done! Works! Awesome!

Issue #1280, opened 2021-07-13
Let binding in macro that shadows macro argument behaves weirdly


hellerve:
This is a weird one. Consider this macro:

```clojure
(defmacro x [y]
  (let [y y]
    y))
```

Ideally, the `let` binding shouldn’t do anything. Instead, the following happens when we use it:

```clojure
(x 'a) ; => y
```

That’s weird. What about more complicated expressions?

```clojure
(defmacro x [y]
  (let [y (if (symbol? y) 10 11)]
    y))
(x 'a) ; => (if (symbol? y) 10 11)
```

Apparently if a `let` binding is introduced and it shadows a macro argument—and only then—the right hand side of the let doesn’t get evaluated?! I’m not quite sure what the heck is going on here.

Cheers

scolsen:
One hypothesis on what might be happening here: one could get this effect if the macro environment and let environment collide. My guess is there's an environment mismanagement somewhere where in rather than making the let a child of the macro environment, we're just replacing the `y` binding in the macro environment with the right hand side without calling the evaluator. Replacing `y` is correct in that case, but we should be calling evaluate somewhere that we aren't.

hellerve:
Another fun fact: if I wrap this code in an `eval` it hangs!

scolsen:
I found the problematic lines. When we resolve let's, we add a flag to make sure the evaluator knows to prefer local definitions over ones of wider scope with the same name.

```haskell
-- src/Eval. line 278
(finalCtx, evaledBody) <- eval newCtx body (PreferLocal (map (\(name, _) -> (SymPath [] name)) binds)) ResolveLocal
```

This then enters this branch in the evaluator:

```haskell
-- line 100
tryAllLookups (PreferLocal shadows) = (if spath `elem` shadows then (getLocal n) else (getDynamic)) <|> fullLookup
```

Since this is a shadow, we call out to `getLocal`

```haskell
-- line 118
getLocal name =
          do
            internal <- contextInternalEnv ctx
            (Binder _ found) <- maybeId (E.getValueBinder internal name)
            pure (ctx, Right (resolveDef found))
```

which then calls `resolveDef` after getting the binding (in this case the forms `y, (if... etc.`

```haskell
-- line 184
resolveDef (XObj (Lst [XObj DefDynamic _ _, _, value]) _ _) = value
    resolveDef (XObj (Lst [XObj LocalDef _ _, _, value]) _ _) = value
    resolveDef x = x
```

[...]

PR #1281, opened 2021-07-14
fix: fix dynamic let bindings recursion and binder leaks

hellerve:
This PR fixes two issues. The first being that dynamic let binders were filled with an unevaluated right hand side as dummy value to enable recursion:

```clojure
; this is allowed
(let [x (fn [n] (if (= n 0) n (x (- n 1))))] (x 10)) ; => 0

; but sadly this also happens
(let [x x] x)
```

The first fix is thus to make sure that the binder trick only happens when the right hand side is a `fn`, which is the only case where recursion might make sense.

The second issue was that binders leaked out of their `let` scope in the global scope:

```clojure
(let [x 1] x) ; => 1

x ; => 1
```

This is because we replaced the internal environment with its parent when leaving the `let` block, _or with itself if it doesn’t have a parent_. The parenthetical does not make sense, especially since internal environments are completely optional. This behavior was dropped alltogether.

I hope all of this makes sense!

Cheers

_Fixes #1280_

### DIFF
--- a/src/Context.hs
+++ b/src/Context.hs
@@ -2,6 +2,7 @@ module Context
   ( ContextError (..),
     replaceGlobalEnv,
     replaceInternalEnv,
+    replaceInternalEnvMaybe,
     replaceTypeEnv,
     replaceHistory,
     replaceProject,
@@ -117,6 +118,11 @@ instance Contextual QualifiedPath where
 replaceInternalEnv :: Context -> Env -> Context
 replaceInternalEnv ctx env =
   ctx {contextInternalEnv = Just env}
+
+-- The previous environment is completely replaced and will not be recoverable.
+replaceInternalEnvMaybe :: Context -> Maybe Env -> Context
+replaceInternalEnvMaybe ctx env =
+  ctx {contextInternalEnv = env}
 
 -- | Replace a context's global environment with a new environment.
 --
--- a/src/Context.hs
+++ b/src/Context.hs
@@ -119,6 +119,8 @@ replaceInternalEnv :: Context -> Env -> Context
 replaceInternalEnv ctx env =
   ctx {contextInternalEnv = Just env}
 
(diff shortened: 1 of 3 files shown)


hellerve:
Oh! This fixes #1280. Thanks to @scolsen for talking through the problem with me!

eriksvedang (approved):
Great!!

eriksvedang (approved):
Great!!

src/Context.hs
```diff
@@ -118,6 +119,11 @@ replaceInternalEnv :: Context -> Env -> Context
 replaceInternalEnv ctx env =
   ctx {contextInternalEnv = Just env}
 
+-- The previous environment is completely replaced and will not be recoverable.
```
eriksvedang, line 122: This comment seems a little off since the above function also completely replaces the env?

src/Context.hs
```diff
@@ -118,6 +119,11 @@ replaceInternalEnv :: Context -> Env -> Context
...
+replaceInternalEnvMaybe :: Context -> Maybe Env -> Context
```
eriksvedang, line 123: This name is a bit confusing since it sounds like it will maybe replace the env (but it's always replaced). I don't really have a better suggestion though, I'm afraid...

eriksvedang (commented):
Re-read the PR and came up with some tiny tiny feedback. I'll merge this tonight unless someone raises concern over it before then!

src/Context.hs
hellerve: It turns out I just half-took the comment from the function above. Will fix.

src/Context.hs
hellerve: Hopefully the type of the function helps clear up the confusion?

hellerve:
Cabal seems to fail on the Windows CI build, but I think this is unrelated—if this issue persists, we will need to investigate!

PR #1282, opened 2021-07-22
fix: respect line number in repl

hellerve:
This PR fixes the behavior of the REPL to respect line numbers.

Example:

```
> (def x 1)
> :i x
x : Int
  Defined at line 1, column 2 in 'REPL'
> (def x 2)
> :i x
x : Int
  Defined at line 3, column 2 in 'REPL' ; <= this used to say line 1 before!
```

Cheers


scolsen (approved):

eriksvedang:
Oh, really nice catch! And a great fix!

PR #1283, opened 2021-07-26
refactor: Cleanup Concretize module

scolsen:
This PR performs a few refactors in efforts to "cleanup" the `Concretize` module, whereby I mean it:

- Removes a bunch of local definitions to make functions easier to modify (no chance of accidentally altering function-local internal state)
- Adds an explicit export list to the module
- Tidies up some monadic usages to make them more terse and remove unneeded function calls.

I've also added some more pattern synonyms.

This is purely a refactor. This shouldn't result in any behavioral changes.

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -3,7 +3,24 @@
 -- | Module Concretize determines the dependencies of polymorphic objects and
 -- resolves the object into a "concrete" version, where its types are no longer
 -- variables.
-module Concretize where
+module Concretize
+  ( concretizeXObj,
+    concretizeType,
+    depsForCopyFunc,
+    depsForPrnFunc,
+    depsForDeleteFunc,
+    depsForDeleteFuncs,
+    depsOfPolymorphicFunction,
+    typesStrFunctionType,
+    concreteDelete,
+    memberDeletion,
+    memberRefDeletion,
+    concreteCopy,
+    tokensForCopy,
+    memberCopy,
+    replaceGenericTypeSymbolsOnMembers,
+  )
+where
 
 import AssignTypes
 import Constraints
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -1,14 +1,20 @@
 {-# LANGUAGE LambdaCase #-}
 
+-- | Module Concretize determines the dependencies of polymorphic objects and
+-- resolves the object into a "concrete" version, where its types are no longer
+-- variables.
 module Concretize where
 
 import AssignTypes
 import Constraints
+import Control.Applicative
 import Control.Monad.State
+import Data.Either (fromRight)
 import Data.List (foldl')
 import Data.Maybe (fromMaybe)
 import Debug.Trace
-import Env (envIsExternal, getTypeBinder, insert, insertX, searchValue)
+import Env (EnvironmentError, empty, envIsExternal, getTypeBinder, insert, insertX, searchValue)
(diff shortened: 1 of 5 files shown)


hellerve:
I haven’t checked the PR yet. When does the new error (about type instantiation) kick in?

scolsen:
> I haven’t checked the PR yet. When does the new error (about type instantiation) kick in?

It's in `instantiateGenericStructType` -- we use to just throw `error "instantiategenericstructtype"`

eriksvedang:
Oh, awesome. I only did a quick read-through but it looks very promising! Will do a proper review as soon as possible.

hellerve (commented):
The diff is a little hard to read, but this looks like a great refactor! Thank you so much for making the codebase more habitable 🎉

src/Concretize.hs
```diff
@@ -36,347 +59,398 @@ data Level = Toplevel | Inside
...
+-- These functions take a state and supporting information and gradually
+-- convert generically typed xobjs into concretely typed xobjs.
+--
+-- The functions prefixed "visit" primarily just recur into the component parts
+-- of different Carp forms while the "concretizeType..." functions perform the
+-- actual type conversion work.
+
+-- | The type of visit functions. These functions convert the types of the
+-- components of a form into concrete types and take the following arguments:
+--   - A bool indicating whether or not type variables are allowed
```
eriksvedang, line 78: The first argument ([SymPath]) isn't covered in this description.

src/Concretize.hs
```diff
@@ -36,347 +59,398 @@ data Level = Toplevel | Inside
...
-                          XObj (Arr structMemberPairs) Nothing Nothing
-                        ]
+    go b = pure [defn, name, args, b]
+visitDefn _ _ Inside _ _ x@(ListPat (DefnPat _ _ _ _)) =
+  pure (Left (DefinitionsMustBeAtToplevel x))
+visitDefn _ _ _ _ _ x = pure (Left (CannotConcretize x))
+
+-- | Concretely type a program entry point. Can only return Int or Unit.
+visitMain :: Visitor
+visitMain vs _ Toplevel tenv env (ListPat (DefnPat defn name@(SymPat (SymPath [] "main") _) args@(ArrPat []) body)) =
```
eriksvedang, line 152: The `vs` variable name is very cryptic to me, what does it stand for?

eriksvedang (commented):
I didn't have time to completely go through all code right now, but I had some small initial comments.

src/Concretize.hs
```diff
@@ -36,347 +59,398 @@ data Level = Toplevel | Inside
...
+  do
+    c <- concretizeTypeOfXObj tenv body
+    v <- visit vs False Inside tenv env body
```
eriksvedang, line 155: It really looks like `v` is related to `vs` here. I think more descriptive names for these visited forms would be very good for readability.

src/Concretize.hs
scolsen: good catch! fixed

src/Concretize.hs
scolsen: it's just the visitedDefinitions -- it's not used in most cases but needs to be threaded through for the one case in which it matters (mkLambda)

src/Concretize.hs
scolsen: Good call, updated

PR #1286, opened 2021-07-27
fix: make Symbol.prefix work on qualified symbols

### LINKED ISSUE #1285: `Symbol.Prefix` can't be called with qualified symbols as prefixes.
Example:

```clojure
鲤 (Symbol.prefix 'Foo.Bar 'baz)
Can’t call `prefix` with Foo.Bar at REPL:1:17.

Traceback:
  (Symbol.prefix (quote Foo.Bar) (quote baz)) at REPL:1:1.

鲤 (Symbol.prefix 'Foo 'baz)
=> Foo.baz
```

I'd expect the first case to produce `Foo.Bar.baz`

scolsen:
This commit makes it so that `Symbol.prefix` can take qualified symbols
as prefix arguments, e.g.

`(Symbol.prefix 'Foo.Bar 'baz) => 'Foo.Bar.baz`

Fixes #1285


hellerve (approved):
Good catch, and sorry for the bug 🎉

PR #1287, opened 2021-07-28
fix: escape quotes in String.prn

hellerve:
This PR fixes `String.prn`, escaping quotes. The function was changed in a way that should facilitate adding more escaped or specially treated character later on if desired.

Cheers


core/carp_string.h
```diff
@@ -8,7 +8,7 @@ String String_allocate(int len, char byte) {
      * String_alloc(10, "a") == "aaaaaaaaaa"
      */
     String ptr = CARP_MALLOC(len + 1);
-    if( ptr!=NULL) {  
+    if (ptr != NULL) {
```
hellerve, line 11: This was automatically picked up by the formatter that runs on test execution.

PR #1294, opened 2021-08-01
fix: fix nested polymorphic types

### LINKED ISSUE #1293: Scoring: Another problem, scoring is incorrect for polymorphics in modules
When one defines a polymorphic type in a module, it is scored incorrectly. Here's an easy reproducible example:

```clojure
(defmodule Foo (deftype (Bar a) [it a]))
(defn bar [x] (Foo.Bar.init x))
(bar 3)
=> out/main.c:1527:46: error: unknown type name 'fooBar__int'
typedef String(*Fn__fooBar__int_MUL__String)(fooBar__int*);
                                             ^
out/main.c:1530:22: error: type specifier missing, defaults to 'int' [-Werror,-Wimplicit-int]
typedef fooBar__int(*Fn__int_fooBar__int)(int);
                     ^
out/main.c:1530:9: error: type specifier missing, defaults to 'int' [-Werror,-Wimplicit-int]
typedef fooBar__int(*Fn__int_fooBar__int)(int);
~~~~~~~ ^
out/main.c:1530:20: error: function cannot return function type 'int (int)'
typedef fooBar__int(*Fn__int_fooBar__int)(int);
                   ^
out/main.c:1641:21: error: function cannot return function type 'fooBar__int' (aka 'int (int *)')
fooBar__int bar__int(int x);
                    ^
out/main.c:3437:30: error: function cannot return function type 'fooBar__int' (aka 'int (int *)')
fooBar__int foo_Bar_init__int(int f);
                             ^
out/main.c:10521:21: error: function cannot return function type 'fooBar__int' (aka 'int (int *)')
fooBar__int bar__int(int x) {
                    ^
out/main.c:10523:12: error: incompatible pointer to integer conversion returning 'fooBar__int' (aka 'int (int *)') from a function with result type 'int' [-Werror,-Wint-conversion]
    return _6;
           ^~
out/main.c:10530:30: error: function cannot return function type 'fooBar__int' (aka 'int (int *)')
fooBar__int foo_Bar_init__int(int f) {
                             ^
out/main.c:10532:13: error: member reference base type 'fooBar__int' (aka 'int (int *)') is not a structure or union
    instance.f = f;
    ~~~~~~~~^~
out/main.c:10533:12: error: incompatible pointer to integer conversion returning 'fooBar__int' (aka 'int (int *)') from a function with result type 'int' [-Werror,-Wint-conversion]
    return instance;
           ^~~~~~~~
out/main.c:10542:19: error: member reference base type 'fooBar__int' (aka 'int (int *)') is not a structure or union
  temp = Int_prn(p->f); 
                 ~^ ~
out/main.c:10553:19: error: member reference base type 'fooBar__int' (aka 'int (int *)') is not a structure or union
  temp = Int_prn(p->f);
```

scolsen:
This PR fixes an issue where polymorphic types defined in modules were not being emitted. To do so, we need to change the signatures of a few functions to take account of the global environment (since types can be defined in modules). I've also added regression tests to catch this issue in the future.

N.B. Depends on PR #1283 

Fixes #1293

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -18,6 +18,12 @@
                 body)
              (set! %variable (+ %variable %step))))))))
 
+(doc Array "is the indexable collection data structure.
+
+Its literal uses brackets, so an array of integers would look like this:
+`[1 2 3]`. It provides both a functional and an imperative API, and is one of
+the core data structures of Carp. It is heap-allocated, for a stack-allocated
+variant you might want to check out [`StaticArray`](./StaticArray.html).")
 (defmodule Array
 
   (doc reduce "will reduce an array `xs` into a single value using a function `f` that takes the reduction thus far and the next value. The initial reduction value is `x`.
--- a/core/Bench.carp
+++ b/core/Bench.carp
@@ -2,6 +2,9 @@
 
 (system-include "carp_bench.h")
 
+(doc Bench "is a module for benchmarking. The most important function of the
+module is [`Bench.bench`](#bench), which takes a function, benchmarks it, and
+prints the results.")
 (defmodule Bench
   (def- min-runs 50)
 
--- a/core/Binary.carp
+++ b/core/Binary.carp
@@ -29,6 +29,7 @@
     significant byte occurs first in a given byte sequence.")
   (deftype ByteOrder LittleEndian BigEndian)
 
+(doc Binary "provides various helper functions to work with bits and bytes.")
 (defmodule Binary
   (register to-int16 (λ [Byte Byte] Uint16))
   (register to-int32 (λ [Byte Byte Byte Byte] Uint32))
--- a/core/BoolExtras.carp
+++ b/core/BoolExtras.carp
@@ -1,0 +1,3 @@
+; this requires doc to work
+(doc Bool "is the basic boolean data type, inhabited by the values represented
+by the literals `true` and `false`.")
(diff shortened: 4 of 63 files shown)


scolsen:
@eriksvedang Marking this as a draft for now. It has the changes from #1283, so that should be merged first.

hellerve:
Let me see whether I understood this right: `scoreTypeBinder` needed to be updated to receive an `Env` so it could lookup the type binder. The rest of the change is basically “just” to thread down the environment all the way from the initial call to concretization. Is that right?

scolsen:
> Let me see whether I understood this right: `scoreTypeBinder` needed to be updated to receive an `Env` so it could lookup the type binder. The rest of the change is basically “just” to thread down the environment all the way from the initial call to concretization. Is that right?

Yes, exactly. The only places that really need the env that didn't have it are concretize and scoring--everything else is just to make sure it gets passed along.

One thing we could consider is whether or not it really makes sense to have separate envs for types and values now that we have specialized functions for finding each. We'd just need some internal means of separating type keys from value keys so that we can look up the appropriate one based on context in the case a value and a type share a name (e.g. we'd have in our map of binders something like `data Foo = Value String | Type String` as the key instead of just `String`.

hellerve:
(I fat-fingered this, sorry!)

eriksvedang:
Always fun to find those "todo:s" :) 
Great work!

PR #1300, opened 2021-08-22
feat: Add additional ignore macros

scolsen:
This commit adds two new ignore macros, ignore*, which wraps an
arbitrary number of forms in calls to `ignore` and ignore-do, which
wraps an arbitrary number of forms in ignore, then bundles the whole in
a do call, effectively executing each form only for side effects and
ignoring all results.

### DIFF
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -133,6 +133,13 @@ Example:
 (defmacro forever-do [:rest forms]
   (list 'while true (cons 'do forms)))
 
+ (doc ignore-do 
+ ("Wraps side-effecting `forms` in a `do`, " false)
+ "ignoring all of their results."
+ "In other words, executes `forms` only for their side effects.")
+ (defmacro ignore-do [:rest forms]
+   (cons 'do (expand (apply ignore-all forms))))
+
 (doc when "is an `if` without an else branch.")
 (defmacro when [condition form]
   (list 'if condition form (list)))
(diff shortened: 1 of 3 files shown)


core/ControlMacros.carp
```diff
@@ -133,6 +133,13 @@ Example:
 (defmacro forever-do [:rest forms]
   (list 'while true (cons 'do forms)))
 
+ (doc ignore-do 
+ ("Wraps side-effecting `forms` in a `do`, " false)
+ "ignoring all of their results."
+ "In other words, executes `forms` only for their side effects.")
+ (defmacro ignore-do [:rest forms]
+   (cons 'do (expand (apply ignore-all forms))))
```
eriksvedang, line 141: Shouldn't this use `ignore*` instead? I can't find `ignore-all`. (Maybe add a test case to catch this also.)

eriksvedang (changes requested):

core/Macros.carp
```diff
@@ -258,6 +258,10 @@ the filename and module name are the same.")
 (defmacro ignore [form]
   (list 'let (array '_ form) (list)))
 
+(doc ignore* "Wraps all forms passed as an argument in an ignore call.")
```
hellerve, line 261: How about we link to the docs for `ignore` here?
```suggestion
(doc ignore* "Wraps all forms passed as an argument in a call to [`ignore`](#ignore).")
```

scolsen: thanks!

core/ControlMacros.carp
scolsen: Good catch--forgot to update after a rename.

scolsen: Fixed! + Added a test.

eriksvedang (approved):
Lovely!

PR #1303, opened 2021-08-22
fix: Allow types defined in modules to be members

### LINKED ISSUE #1302: Types defined in modules cannot be used as members
To reproduce:

```clojure
鲤 (defmodule Foo (deftype Bar))
鲤 (deftype Baz [x (Array Foo.Bar)])
I can’t find a definition for the type `Foo.Bar` at line 2, column 17 in 'REPL'.

Was it registered?

Traceback:
  (deftype Baz [x (Array Foo.Bar)]) at REPL:2:1.
 at REPL:2:10.

Traceback:
  (deftype Baz [x (Array Foo.Bar)]) at REPL:2:1.
```

Expected:

```clojure
鲤 (defmodule Foo (deftype Bar))
鲤 (deftype Baz [x (Array Foo.Bar)])
鲤 :i Baz
Baz : Type
Baz : Module = {
    copy : (Fn [(Ref Baz q)] Baz)
    delete : (Fn [Baz] ())
    init : (Fn [(Array Foo.Bar)] Baz)
    prn : (Fn [(Ref Baz q)] String)
    set-x : (Fn [Baz, (Array Foo.Bar)] Baz)
    set-x! : (Fn [(Ref Baz q), (Array Foo.Bar)] ())
    str : (Fn [(Ref Baz q)] String)
    update-x : (Fn [Baz, (Ref (Fn [(Array Foo.Bar)] (Array Foo.Bar) fq) q)] Baz)
    x : (Fn [(Ref Baz q)] (Ref (Array Foo.Bar) q))
}
```

The fix is pretty straightforward. We just need to update `Validate.hs` to account for types defined in modules.

scolsen:
This commit fixes issue #1302, whereby types defined in modules were not
recognized as valid member types by our validation routines.

We simply need to account for types defined in modules in the global
env, threading the global env along through validation (previously we
only passed the top-level type env, which contains no modules).

Fixes #1302

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -612,7 +612,7 @@ instantiateGenericStructType typeEnv env originalStructTy@(StructTy _ _) generic
       let nameFixedMembers = renameGenericTypeSymbolsOnProduct renamedOrig memberXObjs
           validMembers = replaceGenericTypeSymbolsOnMembers mappings' nameFixedMembers
           concretelyTypedMembers = replaceGenericTypeSymbolsOnMembers mappings memberXObjs
-      validateMembers AllowAnyTypeVariableNames typeEnv renamedOrig validMembers
+      validateMembers AllowAnyTypeVariableNames typeEnv env renamedOrig validMembers
       deps <- mapM (depsForStructMemberPair typeEnv env) (pairwise concretelyTypedMembers)
       let xobj =
             XObj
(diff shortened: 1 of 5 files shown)


eriksvedang (approved):
Ah, makes sense! Having the environments in a "bundle" would really make this kind of fix a lot easier...

PR #1305, opened 2021-08-23
fix: elide definitions of type `Unit`

### LINKED ISSUE #1296: Void defs produce invalid C
It's possible to assign a `def` to a form of type void. The value for the def is then elided in the emitted C, but the def itself isn't:

```clojure
鲤 (def x (Array.push-back! &[1 2 3] 4))
鲤 x
out/main.c:1656:6: error: variable has incomplete type 'void'
void x;
     ^
out/main.c:3507:13: error: expected expression
        x = ;
```

We should either elide these entirely or cast them to void pointers.

hellerve:
This PR fixes #1296 by eliding definitions of `Unit` type. The body still gets executed for side effects, but does not get assigned to a variable.

A regression test is included.

Cheers


scolsen:
Awesome, thanks for this!

scolsen (approved):

PR #1306, opened 2021-08-27
feat: allow C types to satisfy all type constraints

scolsen:
This commit builds on the emit-c feature by permitting C typed values to
be used anywhere in Carp code.

For example, if one wants to use the literal C macro `EDOM`:

```clojure
(register EDOM C "EDOM")

(Int.+ 1 EDOM)
=> 34
```

when compiled, this will produce the call:

```c
Int__PLUS(1, EDOM)
```

So it provides a quite flexible means of using C macros directly. It is,
of course, also radically unsafe. Anyone registering and using values of
the C type better be cautious.

One can get pretty crazy with this feature:

```clojure
(register comment-it C "// commented out;")

(Int.+ 1 comment-it)
=> int _11 = Int__PLUS_(1, // commented out;)
   int* _12 = &_11; // ref
   String _13 = IntRef_str(_12);
```

### DIFF
--- a/src/Constraints.hs
+++ b/src/Constraints.hs
@@ -177,6 +177,8 @@ solveOneInternal mappings constraint =
             Left err -> Left err
             Right ok -> foldM (\m (aa, bb) -> solveOneInternal m (Constraint aa bb i1 i2 ctx ord)) ok (zip args [b, ltB])
     -- Else
+    Constraint _ CTy _ _ _ _ -> Right mappings
+    Constraint CTy _ _ _ _ _ -> Right mappings
     Constraint aTy bTy _ _ _ _ ->
       if aTy == bTy
         then Right mappings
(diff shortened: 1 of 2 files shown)


eriksvedang:
This change to the type checker makes sense to me, if you have arbitrary C code you want to be able to use it wherever (and be prepared to face the consequences.)

But I think the particular situation would be better suited by making the current code do the right thing (since what you have is clearly an integer, and you should be able to type it as such.) Not sure how hard the fix is though.

TimDeve:
@scolsen You're right, I thought for sure I had tried to take a ref and it worked anyway. The alternative is to do something like this but you certainly lose some of the flexibility of the original macro:

```clojure
(deftemplate edom Int "static int $NAME=EDOM" ""))
```

PR #1310, opened 2021-09-14
feat: add expand-compiled

hellerve:
This PR adds `expand-compiled`, which expands and annotates an AST and returns the result. It can be used to inspect the result of macro-expansion, desugaring, and typechecking at the same time:

```
> (expand-compiled '(+ (* 3 4) (inc @&2)))
=> (Int.+ (Int.* 3 4) (Int.inc (Int.copy (ref 2))))
```

Cheers

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -986,6 +986,18 @@ commandC ctx xobj = do
             liftIO (putStr c)
             pure (newCtx, dynamicNil)
 
+-- | This function will return the compiled AST.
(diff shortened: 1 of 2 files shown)


hellerve:
It looks like `unlink` has been deprecated on Windows. I’ll try to fix that in a separate PR.

eriksvedang (approved):

hellerve:
No idea why Nix is failing. Looks like a GH actions snafu?

PR #1311, opened 2021-09-14
fix: use non-deprecated unlink function on windows

hellerve:
This PR fixes our `unlink` varieties on Windows by using the non-deprecated `_unlink` function instead of `unlink`.

Cheers

### DIFF
--- a/core/IO.carp
+++ b/core/IO.carp
@@ -1,5 +1,7 @@
 (system-include "carp_io.h")
 
+(load-once "Platform.carp")
+
 (register-type FILE)
 
 (doc IO "is a module for performing I/O operations. Most functions found in this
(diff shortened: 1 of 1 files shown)


hellerve:
This fixes the problem, but there is a followup issue with the address sanitizer on Windows. I don’t think I’ll be able to debug this; I don’t currently have access to a Windows machine.

TimDeve:
The ASAN error is definitely an odd one, I've found a few similar errors online but nothing concrete, I'm wondering if this is related to that specific build/packaging of llvm. I'll try a Windows build on my local machine.

eriksvedang:
> The ASAN error is definitely an odd one, I've found a few similar errors online but nothing concrete, I'm wondering if this is related to that specific build/packaging of llvm. I'll try a Windows build on my local machine.

Yes, please – this is really strange. I'll hold off mergeing until we know a bit more.

TimDeve:
It is kinda weird that we're seeing some changes here given that we're using the same llvm version in the last passing Windows build and this one. Maybe a change in the underlying Windows image if the problems come from the system headers?

TimDeve:
I think it's still fine to merge this as the deprecation and the Address Sanitizer issues seem separate. See https://github.com/carp-lang/Carp/pull/1314

core/IO.carp
```diff
@@ -1,5 +1,7 @@
 (system-include "carp_io.h")
 
+(load-once "Platform.carp")
```
eriksvedang, line 3: I don't think this should be necessary.

(it makes the `Core.carp` file harder to reason about)

eriksvedang (approved):
Very nice!

core/IO.carp
hellerve: That’s right, this is left over from the previous commit. Sorry about that!

PR #1314, opened 2021-09-14
ci: Uses older windows version for CI

TimDeve:
Checking if the current Windows CI errors are related to changes in the underlying image...

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


TimDeve:
This seems to confirm there is something different about the new windows-latest, I'm not sure if we should stick to 2016 for now or try to dig deeper. Running the latest Windows update on my local machine I do not get any ASAN errors but it might just be that there is something on the windows-latest image that I don't have locally.

eriksvedang:
Let's stay at 2016 for now, to get the other PRs working. Thanks for fixing this!

Issue #1064, opened 2020-12-11
Allocation made in match-ref branch do not get freed properly 


TimDeve:
When we create a value in the branch of a `match-ref`, delete is not called on that value.

Example with a string:

```clojure
(deftype Sum One Two)

(defn main []
    (let [state @"Ok" sumt &(Sum.One)]
     (match-ref sumt
      Sum.One (println* &@&state)
      Sum.Two ())))
```

<details>
<summary>Emitted C code</summary>

```c
int main(int argc, char** argv) {
    carp_init_globals(argc, argv);
    /* let */ {
        static String _6 = "Ok";
        String *_6_ref = &_6;
        String _7 = String_copy(_6_ref);
        String state = _7;
        Sum _11 = Sum_One();
        Sum* _12 = &_11; // ref
        Sum* sumt = _12;
        if(sumt->_tag == Sum_One_tag) {
            Sum* _15_temp = sumt;
            // Case expr:
            String* _24 = &state; // ref
            String _25 = String_copy(_24);
            String* _26 = &_25; // ref
            String _27 = String_str(_26);
            String* _28 = &_27; // ref
            IO_println(_28);
        }
        else if(sumt->_tag == Sum_Two_tag) {
            Sum* _15_temp = sumt;
            // Case expr:
            /* () */
        }
        else {
          // This will not be needed with static exhaustiveness checking in 'match' expressions:
          fprintf(stderr, "Unhandled case in 'match' expression at line 7, column 6 in '/home/tim/dev/sandbox/matc/main.carp'\n");
          exit(1);
        }
        String_delete(state);
        Sum_delete(_11);
    }
    return 0;
}
```

</details>

Example with a lambda:

```clojure
(deftype Sum One Two)

(defn main []
  (let [state 0]
   (match-ref &(Sum.One)
    Sum.One (println* ((fn [] @&state)))
    Sum.Two ())))
```

<details>
<summary>Emitted C code</summary>

```c
int main(int argc, char** argv) {
    carp_init_globals(argc, argv);
    /* let */ {
        int state = 0;
        Sum _10 = Sum_One();
        Sum* _11 = &_10; // ref
        if(_11->_tag == Sum_One_tag) {
            Sum* _11_temp = _11;
            // Case expr:
            // This lambda captures 1 variables: state
            _Lambda_NAKED_LAMBDA_23_env *_23_env = CARP_MALLOC(sizeof(_Lambda_NAKED_LAMBDA_23_env));
[...]

PR #1321, opened 2021-10-05
Solution for #1064 and #843

leblowl:
Removes broken fix for #843 in Emit.hs, thus fixing #1064. And then
this commit focuses on fixing the memory management side of things,
so that we don't add deleters for symbols in the left-hand-side of
match case expressions if we are  matching on a ref (e.g. using
match-ref).

There are several test cases in the #1064 thread. Should I add some of these to the Carp test suite? I don't know what we would assert on, but if someone ran a profiler with the test suite they might be able to detect memory leaks as a way of determining whether the tests are passing or not. Also might be able to get fancy and parse the C code and assert on it.


eriksvedang:
Awesome stuff!

> Should I add some of these to the Carp test suite?

Yes, this feels very important!

There are currently two main ways to test these things:

1. In `test/memory.carp` there are tests that check that nothing leaks, so these are great for all the cases where we want to make sure that correct code does the right thing. There are actually a bunch of tests for match already, named `sumtype-1`, etc.

2. In `test/test_for_errors/` we test code that should generate errors. These are usually a bit tedious to write, but very good for making these bugs not pop up again in the future.

Issue #919, opened 2020-08-24
String.allocate with NULL as the initail value causes unexpected behaviors.


Liorst4:
* The assert that checks out of bound access in `String_string_MINUS_set_BANG_` uses `strlen` instead of the value given to `malloc`. This makes the function unusable on empty strings.

* In the core implementation the length of a string is the `strlen` of it, and not the initial value given to `malloc`, this should be documented.

```clojure
(let [my-string (String.allocate 10 (Char.from-int 0))]
     (do
      (println* (String.length &my-string)) ;; Length is 0, not 10
      (String.string-set! &my-string 1 'a') ;; Assertion fails (Thinks its out of bounds because it checks that the index is less than `strlen`
      ))
```

leblowl:
Running this example as-is with latest Carp from the master branch, resulted in:
```
[PARSE ERROR] "REPL" (line 4, column 43): unexpected ")" expecting "&", "~", "@", "'", "`", "%@", "%", "(", "$[", "[", "{", "-", digit, "0b", "0x", "#\"", "r\"", "\"", "\\" or letter
```

Changing the code code to (use `\a` instead of `'a'`):
```
(let [my-string (String.allocate 10 (Char.from-int 0))]
     (do
      (println* (String.length &my-string))
      (String.string-set! &my-string 1 \a)))
```

Got rid of the parse error, but resulted in:
```
I did not understand the form `(external String.allocate (Fn [Int Char] String))` at /home/dev/src/leblowl/Carp/core/String.carp:35:22.

Traceback:
  (String.allocate 10 (Char.from-int 0)) at REPL:5:17.
(let [my-string (String.allocate 10 (Char.from-int 0))] (do ...) at REPL:5:1.
```
So maybe I am running into a separate bug or misusing the language, but this works for demonstrating the index out of bounds:
```
(def my-string (String.allocate 10 (Char.from-int 0)))
(println* (String.length &my-string))
(String.string-set! &my-string 1 \a)

/home/dev/src/leblowl/Carp/core/carp_string.h:24: bad index: 1 < 0
[RUNTIME ERROR] 'out/Untitled' exited with return value -6.
```
and:
```
(String.string-set! &my-string 0 \a)

/home/dev/src/leblowl/Carp/core/carp_string.h:24: bad index: 0 < 0
[RUNTIME ERROR] 'out/Untitled' exited with return value -6.
```

scolsen:
[...]

PR #1322, opened 2021-10-09
Categorize more static calls correctly

hellerve:
This PR fixes the categorization of static calls directly in calling position. It makes some of the recent problems in #919 work.

Cheers


eriksvedang:
@hellerve Awesome!!!

PR #1327, opened 2021-10-11
fix: Don't pass 'If' to InvalidObj when Obj actually is 'Mod'

### LINKED ISSUE #1325: I didn’t understand the `if` statement at line 1, column 12 in 'core/Blitable.carp'.
```
(defn main []
  (IO.println &(str (width &Int))))
```
```
$ carp width.carp -x
I didn’t understand the `if` statement at line 1, column 12 in 'C:/Users/richie/carp-v0.5.3-x86_64-windows/core/Blitable.carp'.

Is it valid? Every `if` needs to follow the form `(if cond iftrue iffalse)`. at C:/Users/richie/Documents/org/projects/handmade-carp/width.carp:1:2.
```

eriksvedang:
Closes #1325.


scolsen:
> Hmm, actually `root` isn't quite right either. I'll ponder this a bit before mergeing.

The proper `Info` is probably something between `root` and `xobj`

Issue #1323, opened 2021-10-09
Doesn't work with mingw64 clang (edit: fixed)


rgkirch:
It doesn't work if I use clang from mingw but it works with clang from scoop. I got the idea to try installing with scoop because that's what the github action for windows uses. I ran the same command `scoop install llvm` and then just reordered the windows path.

```
$ echo %CARP_DIR%
C:\Users\richie\carp-v0.5.3-x86_64-windows
```

```
$ where clang
C:\msys64\mingw64\bin\clang.exe
C:\Users\richie\scoop\shims\clang.exe

$ carp handmade.carp -x
clang-cl: warning: unable to find a Visual Studio installation; try running Clang from a developer command prompt [-Wmsvc-not-found]
out\main.c(1,10): fatal error: 'assert.h' file not found
#include <assert.h>
         ^~~~~~~~~~
1 error generated.
carp: callCommand: clang-cl.exe  -o out\Untitled -I C:\Users\richie\carp-v0.5.3-x86_64-windows/core/  -D_CRT_SECURE_NO_WARNINGS out\main.c (exit 1): failed

$ clang --version
clang version 12.0.1
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: C:/msys64/mingw64/bin

$ carp
> native-triple
=> ["x86_64" "mingw32" "unknown"]
```

`C:\msys64\clang64\bin\clang.exe` which I neglected to test until now gives the same error output above. This is the one that I thought I was using until I started comparing clangs. I was surprised to find `C:\msys64\mingw64\bin\clang.exe` 

I'm almost sure `msys64/clang64` came from installing `mingw-w64-clang-x86_64-clang`.

In total I've installed 3 things but idk what the difference is between `clang-x86_64-clang` and `x86_64-clang`

`pacman -S mingw-w64-clang-x86_64-clang mingw-w64-x86_64-clang mingw-w64-i686-clang`

```
$ where clang
C:\Users\richie\scoop\shims\clang.exe
C:\msys64\mingw64\bin\clang.exe

$ carp handmade.carp -x
hello world

$ clang --version
clang version 12.0.1
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Users\richie\scoop\apps\llvm\current\bin

$carp
> native-triple
=> ["x86_64" "mingw32" "unknown"]
```

[...]

PR #1332, opened 2021-10-12
feat: register-type improvements

scolsen:
This PR includes a couple of improvements to smooth over some of the rough edges around `register-type` discussed in issue #1323 

- Don't instantiate the `__dummy` field for external (registered) types. We instantiate a dummy field for empty types (no members) defined in Carp for ANSI C compatibility reasons. We also generated these for external type initializers, however, this assumes the external type defines this dummy field, which is unlikely. Instead we no longer instantiate this for external types, and leave all their implementation details up to the user/C code that defines them.
- Automatically implement the `prn` and `str` interfaces for registered types. We generate `prn` and `str` functions for registered types, but they didn't implement the corresponding interfaces. Now they do. 
- Updated C Interop docs.

Here's some examples of previous and new behaviors:

```clojure
;; old way, no auto implementation of prn or str
鲤 (relative-include "../b.c")
鲤 (register-type B [])
鲤 (B.init)
I can’t find any implementation for the interface `str` of type (Fn [(Ref B t7)] String) at line 503, column 10 in '/Users/scottolsen/dev/Carp/core/String.carp'.

None of the possibilities have the correct signature:
    Array.str : (Fn [(Ref (Array a) q)] String)
    StaticArray.str : (Fn [(Ref (StaticArray a) q)] String)
    Function.Arity0.str : (Fn [(Ref (Fn [] z) q)] String)
    .....

;; new way, auto-impl of prn and str
鲤 (relative-include "../b.c")
鲤 (register-type B [])
鲤 (B.init)
(B)
```

```clojure
;; old way, dummy field initialized in memberless registered type 
鲤 (register-type B [])
鲤 (implements str B.str)
鲤 (B.init)
out/main.c:4992:14: error: no member named '__dummy' in 'B'
    instance.__dummy = 0;
    ~~~~~~~~ ^
;; new way, don't initialize dummy fields
鲤 (register-type B [])
鲤 (B.init)
(B)
```

### DIFF
--- a/docs/CInterop.md
+++ b/docs/CInterop.md
@@ -494,11 +494,13 @@ the generated initializer might cause errors.
 The `prn` and `str` functions for the type will also automatically implement
 their corresponding interfaces.
 
-Be mindful that Carp *does not manage the memory associated with external
-types!* Unlike types defined in Carp, Carp will not generate `copy` and `delete`
-functions for registered types. So, even if you have Carp generate initializers
-for convenience, remember that you still need to manage the memory associated
-with registered types manually.
+Be mindful that Carp *does not manage the memory associated with external types
+by default!* Unlike types defined in Carp, Carp will not generate `copy` and
+`delete` functions for registered types. If you use generated initializers for
+a registered type for convenience, remember that you still need to manage the
+memory associated with values of the type manually. If you want Carp to manage
+the memory for a registered type, you can provide implementations of the `copy`
+and `delete` interfaces.
 
 If needed, you can override the name Carp emits for a registered type by
 providing an additional string argument. This comes in handy when the type's


eriksvedang:
@scolsen Great stuff!

Does this consolidate `(register-type A)` and `(register-type A [])` to mean the same thing (or is there actually a good reason not to?)

scolsen:
It might also be interesting to note that the initializer for `(register-type A [])` will only ever with one possible type: the empty struct. Using a scalar typedef would result in incorrect code because of the definition of generated inits:

```c 
typedef struct {} C;                                                                                                                                                                      
                                                                                                                                                                                          
typedef int D; 
```

```clojure
(register-type C [])
(C.init)
=> (C)
(register-type D [])
(D.init)
=> out/main.c:7150:18: error: scalar initializer cannot be empty
    D instance = {};
```

scolsen:
Another note: 

```c
typedef struct {
  int foo; 
} B;
```

```clojure
(register-type B [])
(B.init)
```

Is also valid, with the result that the returned struct has all of its members initialized to zero values.

eriksvedang:
@scolsen That makes sense, thanks! All of this grew kind of organically based on various feature requests, but I think what we have after this PR looks pretty nice and flexible actually.

eriksvedang:
@scolsen I just got one more thought – is it really necessary to handle the `__dummy` variable in the Deftype module? Since it's only there for C compatibility reasons, ideally we would detect the need for it much later, in Emit.hs, which would keep it more contained (and easier to deal with if we want more backends in the future). Let me know what you think!

PR #1353, opened 2021-11-02
feat: Adds flag to always output C id with headerparse

TimDeve:
Adds -c|--emitcname flag to headerparse to always emit the C identifier
in `register` definitions (useful when the register will be in a
module).

Fixes a bug where the kebab case flag would not output C identifiers
making the emitted C identifiers not match with the ones in the headers.

Adds docs entry about headerparse in CInterop doc.

Makes headerparse emit `CChar` instead of `Char` when encountering a
signature containing `char`.

### DIFF
--- a/docs/CInterop.md
+++ b/docs/CInterop.md
@@ -16,6 +16,7 @@ This is an extension of what is covered in the [Language Guide](./LanguageGuide.
   - [`preproc`](#unsafe-preproc)
   - [Registering Types](#register-types)
 - [Callbacks](#callbacks)
+- [Headerparse](#headerparse)
 
 
 ## How Carp generates identifiers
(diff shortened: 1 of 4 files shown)


eriksvedang (approved):
Looks great!

Can you clarify with a small example what difference the `-c` flag actually does?

TimDeve:
```c
// header.h
void MyFunc();
```

Without `-c`:
```clojure
(register MyFunc (λ [] ()))
```

With `-c`:
```clojure
(register MyFunc (λ [] ()) "MyFunc")
```

It's mainly to avoid problem with the way Carp generate C identifiers, for example if I was to move the first declaration in a module called `Mod` the emitted C identifier would be `Mod_MyFunc` breaking the binding.

eriksvedang:
@timdeve OK, great – thanks for the explanation!

PR #1354, opened 2021-11-04
feat: add bytes->hex-string

scolsen:
Converts an array of bytes to a string of their hexadecimal
representation, for example:

```clojure
(Binary.bytes->hex-string &(String.to-bytes "string to hex :)"))
=> 73 74 72 69 6E 67 20 74 6F 20 68 65 78 20 3A 29
```

### DIFF
--- a/core/Binary.carp
+++ b/core/Binary.carp
@@ -326,4 +326,54 @@
   (defn int64-seq->bytes [order is]
     (let [f (fn [i] (int64->bytes order @i))]
       (Array.copy-map &f is)))
+
+  (defn to-hex-str [b]
+    (let [hi (Byte.bit-and b (from-int 0xF0))
+          lo (Byte.bit-shift-left b (from-int 4))
+          nib-one  @""
+          nib-two  @""]
+      (do (case hi
+            (from-int 0x00) (set! nib-one @"0")
+            (from-int 0x10) (set! nib-one @"1")
+            (from-int 0x20) (set! nib-one @"2")
+            (from-int 0x30) (set! nib-one @"3")
+            (from-int 0x40) (set! nib-one @"4")
+            (from-int 0x50) (set! nib-one @"5")
+            (from-int 0x60) (set! nib-one @"6")
+            (from-int 0x70) (set! nib-one @"7")
+            (from-int 0x80) (set! nib-one @"8")
+            (from-int 0x90) (set! nib-one @"9")
+            (from-int 0xA0) (set! nib-one @"A")
+            (from-int 0xB0) (set! nib-one @"B")
+            (from-int 0xC0) (set! nib-one @"C")
+            (from-int 0xD0) (set! nib-one @"D")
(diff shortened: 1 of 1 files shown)


core/Binary.carp
```diff
@@ -326,4 +326,54 @@
...
+            (from-int 0x80) (set! nib-two @"8")
+            (from-int 0x90) (set! nib-two @"9")
+            (from-int 0xA0) (set! nib-two @"A")
+            (from-int 0xB0) (set! nib-two @"B")
+            (from-int 0xC0) (set! nib-two @"C")
+            (from-int 0xD0) (set! nib-two @"D")
+            (from-int 0xE0) (set! nib-two @"E")
+            (from-int 0xF0) (set! nib-two @"F")
+            (set! nib-two @"FATAL ERROR IN BIT LAND! ALL IS LOST"))
+          (String.concat &[nib-one nib-two]))))
```
hellerve, line 371: We could get rid of mutation by inlining the case statements:
```suggestion
  (defn to-hex-str [b]
    (let [hi (Byte.bit-and b (from-int 0xF0))
          lo (Byte.bit-shift-left b (from-int 4))
          nib-one  (case hi
            (from-int 0x00) @"0"
            (from-int 0x10) @"1"
            (from-int 0x20) @"2"
            (from-int 0x30) @"3"
            (from-int 0x40) @"4"
            (from-int 0x50) @"5"
            (from-int 0x60) @"6"
            (from-int 0x70) @"7"
            (from-int 0x80) @"8"
            (from-int 0x90) @"9"
            (from-int 0xA0) @"A"
            (from-int 0xB0) @"B"
            (from-int 0xC0) @"C"
            (from-int 0xD0) @"D"
            (from-int 0xE0) @"E"
            (from-int 0xF0) @"F"
            @"FATAL ERROR IN BIT LAND! ALL IS LOST")
          nib-two  (case lo
            (from-int 0x00) @"0"
            (from-int 0x10) @"1"
            (from-int 0x20) @"2"
            (from-int 0x30) @"3"
            (from-int 0x40) @"4"
            (from-int 0x50) @"5"
            (from-int 0x60) @"6"
            (from-int 0x70) @"7"
            (from-int 0x80) @"8"
            (from-int 0x90) @"9"
            (from-int 0xA0) @"A"
            (from-int 0xB0) @"B"
            (from-int 0xC0) @"C"
            (from-int 0xD0) @"D"
            (from-int 0xE0) @"E"
            (from-int 0xF0) @"F"
            @"FATAL ERROR IN BIT LAND! ALL IS LOST")]
     (String.concat &[nib-one nib-two])))
```

scolsen: thanks!

hellerve (commented):
Should we add a test case here?

eriksvedang:
Oops, missed the comment about test case. That'd be great of course.

eriksvedang:
Just had a thought that there must be a more direct way of converting int to the correct hex char? At the very least an index into a string. Should be faster too.

PR #1337, opened 2021-10-14
feat: Recursive datatypes

scolsen:
This PR is a WIP implementation of recursive data type support in Carp. At the moment, it only supports recursive product types, which are backed by structs with fields that point to values of themselves.

Currently, the following sample code will work:

```clojure
(deftype IntList [head Int tail IntList])                                                                                                                                                                                        
                                                                                                                                                                                                                                 
(defn main []                                                                                                                                                                                                                    
  (let [is (IntList.init 3 (IntList.init 2 (IntList.make 1)))]                                                                                                                                                                   
    (do (IO.println &(str (IntList.head &is)))                                                                                                                                                                                   
        (IO.println &(str (IntList.head (IntList.tail &is))))                                                                                                                                                                    
        (IO.println &(str (IntList.head (IntList.tail (IntList.tail &is)))))                                                                                                                                                     
        0)))
```

Here's a list of what's been implemented thus far, and what remains:

- [x] Support for referring to type names in member fields
- [x] Support for concrete recursive product types
- [x] `make` function for initializing a recursive type with a null recursive part (the end of the recursion chain). 
- [x] Support for recursive sum types
- [ ] Support for generic (type variables in the head of the type) recursive data types

PR #1358, opened 2021-11-19
feat: add box type

scolsen:
This PR adds a Box type to Carp using the initial approach in #1337. 

One important difference is that the box is implemented as a pointer, instead of a struct. While this is simpler, it doesn't play quite as well with Carp's overall programming model, since we need to treat the type specially in `TypesToC` as a result--but then again, we already do this for some other compiler defined types so it's not too strange.

### DIFF
--- a/CarpHask.cabal
+++ b/CarpHask.cabal
@@ -18,6 +18,7 @@ library
   hs-source-dirs:      src
   exposed-modules:     ArrayTemplates,
                        AssignTypes,
+                       BoxTemplates,
                        ColorText,
                        Commands,
                        Concretize,
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -36,7 +36,6 @@ defaultProject =
             "-Wall",
             "-Werror",
             "-Wno-unused-variable",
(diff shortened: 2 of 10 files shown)


TimDeve:
So you've split out the implementation of Box from the recursive type implementation, correct? It looks good to me. I would maybe just add something like `Box.peek : (Fn [(Ref (Box a))] (Ref a))` so you can do things like this:

```clojure
(deftype Foo [bar Int])
(defn main []
  (let-do [b (Box.init (Foo.init 0))]
    (Foo.set-bar! (Box.peek &b) 42)
    (println* b)))
```

But I'm also ok with keeping this PR MVPish and add stuff later if we need.

scolsen:
That's correct! I'll try to add in `Box.peek`!

scolsen:
Ok, added `Box.peek`!

scolsen:
The nix failure stems from a `nix-install` script failure, not from the tests.

eriksvedang:
Very nice, let's ship it!

PR #1361, opened 2021-12-06
refactor: add type candidates and template generators

scolsen:
This PR adds two new modules that help with managing user defined types:

- TypeCandidate: defines a representation of user defined types that aren't yet fully defined. It's used as input for type validation and function/binding generation for user defined types. It simplifies managing user defined types by containing all the information needed for the type in a single structure (e.g., no more need to pair up field xobjs throughout the codebase)
- TemplateGenerator: Defines an approach for standardizing the generation of function templates based on the contents of type candidates. It abstracts away some of the details involved in defining templates and makes their definition slightly more declarative. It also enables us to unify much of the generic and concrete template code.

This PR also removed `SumtypeCase` which is no longer needed after the introduction of `TypeCandidate`. TypeCandidates have `TypeFields` which do the old job of sumtype cases (and similarly of XObj pairings on the struct/product side).


scolsen:
Again, the nix build failure seems to stem from nix related set up and is not the result of a failed build or failing tests.

eriksvedang:
From reading your description of the PR, this sounds fabulous! I have yet to dig into the code, might take a few days to get through it.

scolsen:
Sounds good! Yeah the diff is pretty large. No functionality changes though, so hopefully the new code is easy enough to understand. Let me know if you have any questions!

eriksvedang (commented):
@scolsen I've read through once now, It looks very good! I want to go over some parts a second time to understand them better before mergeing. Sorry for the review taking a while, last week was super hectic.

scolsen:
No problem! Let me know if I can help at all.

eriksvedang (changes requested):
Alright, I feel like I've got a pretty good overview now! It was a lovely read, really.

Lots of new types, which is great :D 

I do worry a little about the abstraction tower "TemplateGenerator > TemplateCreator > Template"; it's not very intuitive. But it definitely does the job, and also brings some more structure to the process.

I have one request before we merge – could you run `./scripts/format.sh` on the code?

Apart from that, this is awesome and more than ready for `master`. Let's kick off a celebratory release once it's in!

eriksvedang:
@scolsen Alternatively, I can make a follow up PR where I run the formatter. Might actually be better, the diff from this merge commit will be neater. If that sounds good, I'll just merge as-is.

eriksvedang (approved):

scolsen:
> @scolsen Alternatively, I can make a follow up PR where I run the formatter. Might actually be better, the diff from this merge commit will be neater. If that sounds good, I'll just merge as-is.

Sorry for missing this--I was away! Thanks for applying the formatting!

PR #1362, opened 2021-12-08
fix: permit registering types in modules

scolsen:
Historically, we did not support calling (register-type) within a
module--while there were instances of such calls (even in our core
library!) they were not properly handled in two ways:

1. They weren't added to the right place in the type environment.
2. Their corresponding emitted code wasn't correct.

This commit enables registering types in modules by making a few fixes:

1. Fix the logic in environment mutations -- before, adding a type to an
   environment would result in traversing the environment chain
   incorrectly. This is now fixed (we traverse modules and retrieve a
   type environment only at the end of the traversal).
2. Fix the typedef emission for registered types--it previously emitted
   the C code for the type path, not for the type definition (it called
   pathToC not tyToC).

This will now allow authors of wrapper libraries to add more structure
to their Carp wrapper APIs.

This commit also adds a test to check the behavior.

### DIFF
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -921,7 +921,7 @@ toDeclaration (Binder meta xobj@(XObj (Lst xobjs) _ ty)) =
     XObj (ExternalType Nothing) _ _ : _ ->
       ""
     XObj (ExternalType (Just override)) _ _ : XObj (Sym path _) _ _ : _ ->
-      "typedef " ++ override ++ " " ++ pathToC path ++ ";"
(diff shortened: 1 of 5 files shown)


scolsen:
Again, the nix CI is currently flaky/broken.

eriksvedang:
Very nice! Can this be merged independently of your other PR?

scolsen:
@eriksvedang yes, should be fine to merge!

eriksvedang (approved):
Excellent work, this is great! 💪

PR #1367, opened 2021-12-28
docs: Instructions to ensure correct handling of utf-8

IrrenWirr:
Should also update the release page with :
* For POSIX : ensure that the `LC_CTYPE` locale has an UTF-8 aware value

It would also be nice to implement something along the lines of this elisp snippet in `app/Main.hs` to warn the user. (Couldn't do it myself because I don't know haskell)

```elisp
(defun lc-ctype-is-utf8-compatible? (lc-ctype-value-str)
  (when (and
         ;; In Unix systems LC_CTYPE has always a value,
         ;; but Windows 10 doesn't have value for LC_CTYPE (and doesn't need it to handle UTF-8 correctly)
         lc-ctype-value-str
         ;; When lc-ctype-value-str has a value,
         ;; it is a very reliable heuristic to detect a POSIX system.
         (string-match
          ;; regex to match
          ;; either “.utf8” or “.utf-8”
          ;; at the end of the string
          "\\.utf-?8$"
          ;; downcase because the locale is case insensitive
          (downcase lc-ctype-value-str)))
    (message "Your `LC_CTYPE` environment variable isn't UTF-8 compatible.
This may cause odd behaviours in the repl. Please export it to a UTF-8 compatible value like `C.UTF-8`")))
```

### DIFF
--- a/docs/Install.md
+++ b/docs/Install.md
@@ -29,6 +29,20 @@ You should now be able to start Carp from anywhere:
 $ carp
 ```
 
+## Ensuring an UTF-8 aware LC_CTYPE locale in POSIX environments
+
+To be able to handle UTF-8 correctly when using `carp`'s interactive repl (binaries from `carp` always handle UTF-8 correctly),
+all POSIX aware environments (Linux, MacOs or even Emacs's eshell inside a Windows 10)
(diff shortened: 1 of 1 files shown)


eriksvedang (approved):
Awesome, thanks for this!

Issue #1317, opened 2021-09-16
A brief explanation of the lifetimes implementation


eriksvedang:
## How the Carp lifetimes work right now

1. Lifetime variables are unified just like type variables (see https://github.com/carp-lang/Carp/blob/5a449b09191594111b8792cc74d6183c55f498f4/src/Constraints.hs#L158)

2. When analyzing the memory management code, when something has a lifetime in its type, that lifetime gets associated with the original location being referred to (see https://github.com/carp-lang/Carp/blob/5a449b09191594111b8792cc74d6183c55f498f4/src/Memory.hs#L589)

3. For the rest of that function definition, when a reference is encountered, those mappings are then checked against the live set of variables (referred to as "deleters" in the code base) to make sure that whatever is being referenced is still alive (see https://github.com/carp-lang/Carp/blob/5a449b09191594111b8792cc74d6183c55f498f4/src/Memory.hs#L542)

## Regarding scopes

The reason we're referring to variables (and not to scopes) is that a scope is too coarse of a unit. Within a scope multiple variables can get created in sequence with functions being called in-between. That could be solved by creating one scope per variable but still, we need to also know exactly when a variable is dead -- which might not be at the end of the scope (if it is given away before that.)

## Problems / improvements

The problem with point 2 above is that it only assigns the mapping on the *first* occurence. A more correct behaviour (I think) would be to have a set of variables that each lifetime depends upon. This solves bugs like the second example in this issue: https://github.com/carp-lang/Carp/issues/470

```
(defn main []
  (let-do [x ""]
    (let [a [@"hi"]]
      (set! x (Array.unsafe-nth &a 0)))
    (println* x))) ; a was already dropped!
```

Currently this fools the memory system because everything unifies to a single lifetime variable, and that lifetime only depends on `x`, not on `a`. Having the lifetime depend on a set solves this neatly and the bug is caught -- it leads to some code in Core not compiling though. Would be nice to re-implement that and look at the problem together!

A second problem with the whole setup is that a bunch of lifetimes can be unified with the `static` lifetime which just makes the memory system give up and not check those references. I'm kind of at a loss how to fix that one, so would be nice to talk it through.

redbar0n:
maybe this could serve as some inspiration (grouped allocations/lifetimes): https://www.sophiajt.com/search-for-easier-safe-systems-programming/

hellerve:
[...]

PR #1368, opened 2021-12-29
feat: stack-allocated lambdas

eriksvedang:


TimDeve:
Very cool, I've been waiting for this, I think some of my programs can be alloc-free now.

TimDeve:
Pulled the PR to play with it, found a couple of problem:

The first one might be linked to #1317, is that you can return a (Ref Fn) meaning that one that Fn capture is now out of the scope.

The second one is related to the Lambda deleter which assumes what's in it's env is owned by the Lambda (which sounds correct) so it deletes things, but it's not actually taking ownership properly so the normal deleters for the variables still run when the scope ends.

src/Concretize.hs
```diff
@@ -327,7 +327,7 @@ mkLambda visited allowAmbig _ tenv env root@(ListPat (FnPat fn arr@(ArrPat args)
                     modify (deleterDeps ++)
                     modify (copyFn :)
                     modify (copyDeps ++)
-            pure (Right [XObj (Fn (Just lambdaPath) (Set.fromList capturedVars)) (xobjInfo fn) (xobjTy fn), arr, recBody])
+            pure (Right [XObj (Fn (Just lambdaPath) (Set.fromList capturedVars)) (xobjInfo fn) (xobjTy fn), arr, body])
```
scolsen, line 330: Was the change from `recBody` to `body` intentional? I think this was needed to appropriately handle recursive anonymous functions in let bindings, e.g. `(let [f (fn [x] (if (= x 0) x (f (- x 1)))])`, since we rename the lambda (see renameRecursives above).  It doesn't seem like these changes would alter that behavior but maybe I'm missing something.

scolsen (commented):
Very nice! I'm wondering, if one should need a `malloc`'d lambda for some reason--would it be possible after this change? How would one do that? Not sure if it'd ever be necessary though.

eriksvedang:
@scolsen The idea is that you copy the lambda and that will put its environment on the heap, just like before.

TimDeve:
Yes, this seems to be working, it's just that it's hitting the double free problem from above:

```clojure
(defn gen []
  (let [s1 @"Wow"
        s2 @"Lambda"
        lam (fn [] (IO.println &(String.join "" &[s1 s2])))]
    @lam))

(defn main []
  ((gen)))
```

eriksvedang:
@TimDeve will look into that asap!

hellerve:
Should I pick this up?

PR #1372, opened 2022-01-07
fix: Fixes nix install by using correct pkg-configDepends config key

TimDeve:
As reported by Hrafn Blóðbók, nix updated references to pkgconfig in its codebase to use the `pkg-config` spelling, this seems to break the installation of Carp using nix.

See this PR: https://github.com/NixOS/nixpkgs/commit/9bb3fccb5b55326cb3c2c507464a8a28d44d1730

The current error when installing:
```
error: anonymous function at /nix/var/nix/profiles/per-user/root/channels/nixos/pkgs/development/haskell-modules/generic-builder.nix:13:1 called with unexpected argument 'pkgconfigDepends'

       at /nix/var/nix/profiles/per-user/root/channels/nixos/lib/customisation.nix:69:16:

           68|     let
           69|       result = f origArgs;
             |                ^
           70|
```

We might need to update the nix action as well, just pushing it like this for now.

### DIFF
--- a/.github/workflows/nixpkgs.yml
+++ b/.github/workflows/nixpkgs.yml
@@ -12,7 +12,7 @@ jobs:
         os: [ubuntu-latest]
     steps:
     - uses: actions/checkout@v2.3.4
-    - uses: cachix/install-nix-action@v15
+    - uses: cachix/install-nix-action@v16
       with:
-        nix_path: nixpkgs=channel:nixos-20.09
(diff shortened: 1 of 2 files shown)


TimDeve:
Updated the Github Actions so it uses 21.05, guess it means this now the minimum Nix version to install Carp with, hope that doesn't break anyone workflows.

PR #1373, opened 2022-01-07
fix: #1347 by  ignoring generically typed symbols on printing C

### LINKED ISSUE #1347: carp: Can't emit symbol of generic type: e : a at line 4, column 3 in
```
(Project.config "echo-c" true)

(defn identity [e]
  e)

(defn main []
  (ignore (identity 1)))
```
```
$ carp no.carp -x
carp: Can't emit symbol of generic type: e : a at line 4, column 3 in 'C:/Users/richie/Documents/org/projects/handmade-carp/no.carp'
  error, called at src\Emit.hs:210:15 in CarpHask-0.5.3.0-ApN9rIbXYwcLYnQzPVPog2:Emit
```
The below code is fine.
```
(defn identity [e]
  e)

(defn main []
  (ignore (identity 1)))
```

hellerve:
This PR fixes #1347 by ignoring generically typed symbols when we echo C output. That should usually be safe, since the specializations will still be printed when the forms require them.

Cheers

### DIFF
--- a/src/Primitives.hs
+++ b/src/Primitives.hs
@@ -9,7 +9,7 @@ import Control.Applicative
 import Control.Monad (foldM, unless, when)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Bifunctor
-import Data.Either (fromRight, rights)
+import Data.Either (fromRight, isRight, rights)
 import Data.Functor ((<&>))
 import Data.List (foldl')
 import Data.Maybe (fromJust, fromMaybe)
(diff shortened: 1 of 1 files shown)


scolsen (approved):
Nice!

src/Primitives.hs
```diff
@@ -182,10 +182,15 @@ define hidden ctx qualifiedXObj =
     defineInTypeEnv = pure . fromRight ctx . (insertTypeBinder ctx qpath)
     defineInGlobalEnv :: Binder -> IO Context
     defineInGlobalEnv newBinder =
-      when (projectEchoC (contextProj ctx)) (putStrLn (toC All (Binder emptyMeta annXObj)))
+      when (projectEchoC (contextProj ctx) && canBeEmitted annXObj) (putStrLn (toC All (Binder emptyMeta annXObj)))
         >> case (lookupBinderInGlobalEnv ctx qpath) of
           Left _ -> pure (fromRight ctx (insertInGlobalEnv ctx qpath newBinder))
           Right oldBinder -> redefineExistingBinder oldBinder newBinder
+    canBeEmitted :: XObj -> Bool
```
eriksvedang, line 189: There is already the function `checkForUnresolvedSymbols` in Emit.hs, would that work instead of defining this new one? (Since that function is much more complicated, I also wonder if it could be simplified, or if this one is too simplistic...)

hellerve: Ah, I forgot about that one. It looks a little more robust than my simplistic attempt at rewriting it, so I’ll change the PR to use it!

eriksvedang (approved):
Awesome!

PR #1379, opened 2022-01-14
fix: mangle field names in setter templates

### LINKED ISSUE #1378: regression Struct field names are no longer mangled in field mutators
example problem output:

```c
CligFlag CligFlag_set_MINUS_default_MINUS_value(CligFlag p, String newValue) {
    String_delete(p.default-value);
    p.default_MINUS_value = newValue;
    return p;
}   
```

scolsen:
This fixes a regression whereby the names of struct fields weren't
mangled in calls to setters/mutators, resulting in invalid C code if the
user happened to use reserved words in their struct field names such as
"short" or a disallowed character like a dash.

Fixes #1378

### DIFF
--- a/src/Deftype.hs
+++ b/src/Deftype.hs
@@ -288,7 +288,7 @@ setterGenerator = TG.mkTemplateGenerator tgen decl body deps
     body GeneratorArg {tenv, env, instanceT = (FuncTy [_, ty] _ _), value = (TC.StructField name _)} =
       multilineTemplate
         [ "$DECL {",
-          memberDeletion tenv env (name, ty),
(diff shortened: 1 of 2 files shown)


TimDeve:
Should we add unit tests for when there's code changes that fix regressions/bugs?

scolsen:
> Should we add unit tests for when there's code changes that fix regressions/bugs?

Definitely, I can add some to this PR

eriksvedang (approved):
Nice!

scolsen:
Added a regression test

eriksvedang:
Nice!

PR #1398, opened 2022-03-15
feat: add c-name meta field

### LINKED ISSUE #1389: permit overriding emitted C names
On some occasions, C output from Carp may be used as a library called from another language in cases where it's not feasible to write an entire application in Carp (e.g. some part of your application is written in another language that supports C interop but not carp interop)

To make the experience with this use case a bit nicer, we should allow users to hardcode emitted C names for binders to give their public API nice names. We could do this using meta set, e.g.

`(meta-set Foo.bar "cname" "fooify")`

Would cause the `Foo.bar` binding to be emitted as `fooify` in emitted c code.

scolsen:
This commit adds a new special compiler meta key, c-name, that enables
users to explicitly c the C identifier Carp should emit for a given
symbol. For now, it is only explicitly supported for Def and Defn forms.

For example:

```clojure
(defn foo-bar [] 2)
(c-name foo-bar "foo_bar")
```

Will cause foo-bar in emitted C code to be emitted as `foo_bar` instead
of `foo_MINUS_bar`.

I've also refactored some of the meta code to be a bit more principled
about keys that are understood by the compiler.

fixes #1389


scolsen:
I'm not sure why the commit history here contains a bunch of old stuff. afaict the change set is correct. 

I can re-branch and cherrypick if we want to discard these old commits.

eriksvedang (approved):
Looks good! Nice feature 💪

eriksvedang:
@scolsen I'll squash merge so I think it's fine to leave the history. Do you know if the Windows build failing is a real problem?

TimDeve:
> @scolsen I'll squash merge so I think it's fine to leave the history. Do you know if the Windows build failing is a real problem?

Seems to be unrelated, I'll try to raise a PR for that separetely.

hellerve (approved):
Thanks especially for the doc rewrite! <3

PR #1399, opened 2022-03-17
ci: Fixes scoop install as admin user

TimDeve:
This fixes error in Windows pipelines by [forcing the scoop installer](https://github.com/ScoopInstaller/Install#for-admin) to run.

I don't believe there is much point in not-running as Admin given the environment the CI runs in.

Edit: had to upgrade the Windows version because 2016 is getting deprecated but it comes with ASAN errors we've seen before.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,7 +39,8 @@ jobs:
       - if: matrix.os == 'Windows'
         name: Install Scoop
         run: |
-          iwr -useb get.scoop.sh | iex
+          iwr -useb get.scoop.sh -outfile 'install-scoop.ps1'
+          ./install-scoop.ps1 -RunAsAdmin
           echo "~\scoop\shims" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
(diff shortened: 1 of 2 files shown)


eriksvedang:
Looks good, but the CI still fails on Windows?

TimDeve:
Yes this is the error I mentioned in the edit, still trying to fix it.

eriksvedang:
@TimDeve Oh, sorry I read sloppily. Thanks for looking into it!

TimDeve:
So I've disabled sanitize-addresses because I can't find a solution to the problem, it seems to come from the clang that comes with these newer windows version but even using our scoop's clang it doesn't work.

I believe it needs a specific combination of flags (maybe even passing the asan libs directly if some of the issues I found are to be believed) but the fact I can't reproduce locally makes it quite hard to debug.

I would merge like this as I think it's more useful to have a partially working Windows CI than none at all, I will raise an issue to track the problem and try to make a minimal CI pipeline to reproduce the issue.

eriksvedang:
@TimDeve Sounds good, thanks again!

PR #1405, opened 2022-03-27
feat: add IO.fgetc

scolsen:
IO defined a function, fgetc (distinct from IO.Raw.fgetc) which actually
produced invalid code, since the name was not overridden and C does not
define IO_fgetc.

There was also a TODO to handle EOF conditions; so, I've implemented the
function, checking for EOF and error conditions using the Raw stdio
wrappers. IO.fgetc returns a Char in Success on success and an error
string on failure.

This PR also registers `feof` and `ferror` -- `IO.fgetc` is implemented in pure Carp by calling these wrapper functions.

### DIFF
--- a/core/IO.carp
+++ b/core/IO.carp
@@ -77,6 +77,8 @@ module are wrappers around the C standard library.")
     (register SEEK-END Int "SEEK_END")
     (doc      ftell "gets the position indicator of a file (thin wrapper for the C standard library).")
     (register ftell (Fn [(Ptr FILE)] Int) "ftell")
+    (register feof (Fn [(Ptr FILE)] Bool) "feof")
+    (register ferror (Fn [(Ptr FILE)] Bool) "ferror")
   )
 
   (doc      println "prints a string ref to stdout, appends a newline.")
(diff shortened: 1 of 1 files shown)


core/IO.carp
```diff
@@ -90,7 +92,11 @@ module are wrappers around the C standard library.")
   (doc      get-line "gets a line from stdin.")
   (register get-line (Fn [] String))
   (doc      fgetc "gets a character from a file pointer (thin wrapper for the C standard library).")
-  (register fgetc (Fn [(Ptr FILE)] Char)) ; TODO: check EOF handling (see carp_io.h)!
+  (defn fgetc [file]
+    (let [char (IO.Raw.fgetc file)]
+      (if (or (IO.Raw.feof file) (IO.Raw.ferror file))
```
hellerve, line 97: Should those two errors be distinct?

scolsen: yeah, they should--another thing I was wondering about is if we should define an actual Error type for these errors rather than just a string.

hellerve: I’m usually too lazy for this, but it’s probably the correct thing to do :)

scolsen: added! -- eventually we'll want to check `errno` and report on what the error was, but for now we at least distinguish between EOF and other errors.

hellerve (approved):

eriksvedang (approved):

PR #1407, opened 2022-04-01
feat: don't manage blittable types

scolsen:
The `blit` interface, when implemented, indicates that a type is freely
copyable and does not require deletion. It is similar to Rust's Copy
trait and indicates that a type should have copy instead of "move"
semantics. This commit updates the memory management system to ignore
such types, since we do not need to worry about deleting them.

Thus, when using a blittable type, users can freely use a variable of
the type in more than one place.

This commit also fixes a small typo and adds a TODO to warn users when
delete and blit are both implemented (as typically one shouldn't need
delete if a type is truly blittable).

In other words, the borrow checker will no longer complain about the following:

```clojure
(deftype Thing Foo Bar)

(defmodule Thing
  (defn blit [x] (the Thing x))
  (implements blit blit)
)

(defn multi-use-thing [] 
  (let-do [thing (Thing.Foo)
           other-thing thing]
    (f thing)
    (g thing)
    (h other-thing)
    (k other-thing)))
```

That is, there's no need to borrow types that indicate blit. They are assumed to have copy semantics and not require deletion.

This PR also includes a test for this behavior.


scolsen:
if this merges I'll open up an issue to produce a warning for types that implement both `blit` and `delete` (there's a TODO in this commit for this)

TimDeve:
It would be useful to expose `interfaceImplementedForTy` as a Dynamic function so that people can derive blit rather than implementing it themselves as it's unsafe. Could be a follow up PR though.

test/memory.carp
```diff
@@ -568,4 +594,5 @@
   (assert-no-leak test sumtype-12 "sumtype-12 does not leak")
   (assert-no-leak test box-1 "box-1 does not leak")
   (assert-no-leak test box-2 "box-2 does not leak")
+  (assert-no-leak test blit-1 "blit-1 does not leak")
```
TimDeve, line 597: Do we need the opposite test (non-blit give error when moving twice) to make sure there is no regression or is that covered by existing tests?

TimDeve:
> ...perhaps we should add something about manual implementations being unsafe somewhere in the docs

For sure, something like "This is a marker interface. DO NOT implement yourself"

test/memory.carp
eriksvedang: It should be covered.

eriksvedang (approved):
Pretty straight-forward!

test/memory.carp
scolsen: sg!

eriksvedang:
@scolsen I'll merge this now if that's OK?

scolsen:
Go for it!

PR #1408, opened 2022-04-04
chore: Updates Stackage version to 19.2

TimDeve:
Comes with GHC 9.0.2.

Some tests in `test/dynamic_map.carp` were updated to not be order dependent.

### DIFF
--- a/test/dynamic_map.carp
+++ b/test/dynamic_map.carp
@@ -1,189 +1,206 @@
 (load "Test.carp")
 (use Test)
 
+(doc list-equal-unordered "Checks two lists have the same values, not necessarily in the same order")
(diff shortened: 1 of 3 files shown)


test/dynamic_map.carp
```diff
@@ -1,189 +1,206 @@
 (load "Test.carp")
 (use Test)
 
+(doc list-equal-unordered "Checks two lists have the same values, not necessarily in the same order")
+(defndynamic list-equal-unordered [lx ly]
```
eriksvedang, line 5: np. I think we usually call these `xs` and `ys`.

eriksvedang (approved):
Lovely!

PR #1412, opened 2022-04-07
feat: register MAX and MIN macros for stdint types

scolsen:
Adds MAX and MIN for each INT<N> type and MAX for each UINT<N> type.
These are macros defined by stdint.h and are sometimes useful for bounds
determinations and conversions and such.

### DIFF
--- a/core/Double.carp
+++ b/core/Double.carp
@@ -6,6 +6,7 @@
   (implements pi Double.pi)
   (def e 2.718281828459045)
   (register MAX Double "CARP_DBL_MAX")
+  (implements MAX MAX)
   (register = (Fn [Double Double] Bool))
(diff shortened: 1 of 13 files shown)


eriksvedang (approved):
Looks good! Should these be interfaces?

scolsen:
> Looks good! Should these be interfaces?

Good point! They probably should since we have MAXes for the other integer types too. I can make the change.

scolsen:
Added interfaces!

scolsen:
so it turns out these interfaces are breaking tests because our interface resolution doesn't currently overwrites symbol lookup modes! I've got a fix.

scolsen:
Once #1415 merges the tests will pass again and we can merge this.

PR #1413, opened 2022-04-07
fix: respect let binding shadowing in memory management

### LINKED ISSUE #597: Type checker can behave weirdly when using the name of the functions as a binder
Example:

```
(defn n [xs]
  (let [xs [1 2 3]
        n &xs]
    n)) ;; The type of 'n' here is wrong, plus this should be a type error!
```

scolsen:
Previously, we didn't account for shadowing in let bindings in our
memory management routines. This led to rare situations in which
multiple deleters might be added for a single variable name, for
example:

```clojure
(defn n [xs]
  (let [xs [1 2 3]
        n &xs]
    n))
```

The borrow checker would fail on this code since it would assign `xs`
two deleters, one for the untyped argument and another for the let
binding.

Instead, we now perform *exclusive* ownership transfer for the duration
of the let scope--when a shadow is introduced, the previous deleters for
that variable name are dropped until the end of the let scope, since we
evaluate all instances of the shadowed name to the more local binding.
At the end of the let scope, the original deleter is restored.

The code in the issue now produces: 

```clojure
The reference 'n' (depending on the variable 'xs') isn't alive at line 4, column 5 in '/Users/scottolsen/shadow.carp'. at /Users/scottolsen/shadow.carp:1:2.

Traceback:
  (defn n [xs] (let [xs [1 2 3] n (ref xs)] n)) at /Users/scottolsen/shadow.carp:1:1.
(load "../../shadow.carp") at REPL:1:1.
```

fixes #597

### DIFF
--- a/src/Info.hs
+++ b/src/Info.hs
@@ -12,12 +12,14 @@ module Info
     makeTypeVariableNameFromInfo,
     setDeletersOnInfo,
     addDeletersToInfo,
+    uniqueDeleter,
   )
 where
 
 import Path (takeFileName)
 import qualified Set
 import SymPath
+import Data.List (unionBy)
 
 -- | Information about where the Obj originated from.
 data Info = Info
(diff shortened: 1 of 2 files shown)


scolsen:
As discussed in the issue, I originally thought we could just ignore type variables wrt to memory mgmt, but this doesn't work. 

Luckily, we can use the state monad to selectively use deleters for shadows then restore the original deleters once we leave the let scope, which works quite nicely. An alternative would be to apply alpha renaming so that all bound variables are unique according to their scope, but this would be a significant effort.

eriksvedang:
Nice!

Seems like the failing test is just due to an improved error message that needs adjusting in test/output/tes
t/test-for-errors, right?

Alpha renaming would make the memory management code cleaner, but also make error messages harder (since they would refer to variable names that don't exist in the source.) I'll try to think a little about if there's anything else we can do but if I can't come up with anything, I think this solution here makes sense.

Thanks a lot for solving the mystery of this bug!

scolsen:
Fixed!

eriksvedang (approved):
Thinking a bit more about it, I think this solution is the best one!

Before we merge, can you add a regression test for this bug (with a reference to the issue).

scolsen:
> Thinking a bit more about it, I think this solution is the best one!
> 
> Before we merge, can you add a regression test for this bug (with a reference to the issue).

added!

eriksvedang (approved):

eriksvedang:
Lovely!

PR #1417, opened 2022-04-18
feat: add fputc wrapper to IO

scolsen:
fputc is a low-level function that writes a single C character to a
file. We can use this as a basis for more elegant APIs.

The signature matches that of the C standard library function for
compatibility, which takes an Int instead of a char. However, the
documentation notes that the int argument is converted to an unsigned
char.


scolsen:
Looks like the Linux and MacOS CI errors were transient errors related to downloading GHC

core/IO.carp
```diff
@@ -18,6 +18,8 @@ module are wrappers around the C standard library.")
     (register get-char (Fn [] Int) "getchar") 
     (doc      fgetc "gets a character from file (thin wrapper for the from C standard library).")
     (register fgetc (Fn [(Ptr FILE)] Int) "fgetc") 
+    (doc fputc "writes a character to a file (thin wrapper for the from C standard library).")
```
eriksvedang, line 21: "for the from" seems to be missing a word?

scolsen: good catch, fixed!

eriksvedang (approved):

PR #1421, opened 2022-04-22
refactor: make fread more general

scolsen:
fread can technically accept any object -- it takes a void pointer. Our
current implementation assumes the object is a String, which limits the
applicability of this function. This change alters the definition to
accept any type.


scolsen:
If anyone has good ideas around how to do this in a way that won't break existing callers, please share!

PR #1424, opened 2022-08-06
replace lambda with flip

mm1995tk:
hlint says

```
 Avoid lambda. Replace with: flip (execStr i)
```

so I replace lambda with flip.


mm1995tk:
Because I missed the contributing guide, i closed

PR #1429, opened 2022-08-26
feat: add bag data structure

hellerve:
This PR adds a `Bag` data structure, which acts as a set (only storing each thing once) that keeps track of the number of times an item was put in the structure. It’s thus a deduplicated, unordered collection.

Test cases are included.

Cheers

### DIFF
--- a/core/Map.carp
+++ b/core/Map.carp
@@ -832,8 +832,8 @@ collision.")
   "Implementation notes: it is a map from elements to number of occurrences"
   "under the hood.")
 (defmodule Bag
-  ;(private internal)
-  ;(hidden internal)
+  (private internal)
+  (hidden internal)
   (private set-internal)
   (hidden set-internal)
   (private update-internal)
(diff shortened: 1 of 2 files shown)


scolsen:
Nice!

core/Map.carp
```diff
@@ -806,3 +821,94 @@ collision.")
...
+  internal (Map a Int)
+])
+
+(doc Bag
+  "is an unordered datatype that only stores all its equal elements once while preserving size by storing the count of each element."
+  ""
+  "Implementation notes: it is a map from elements to number of occurrences"
+  "under the hood.")
+(defmodule Bag
+  ;(private internal)
```
scolsen, line 835: Should these comments be removed?

scolsen (approved):

core/Map.carp
hellerve: They should be re-added, actually :)

eriksvedang (approved):

PR #1433, opened 2022-09-27
refactor: remove unnecessary solveOneInternal function

scolsen:
Since solveOne was defined as `= solveOneInternal`, we can just call it directly.


eriksvedang:
Nice!

PR #1441, opened 2022-10-31
chore: disable unused-but-set-variable for clang

scolsen:
At some point (I believe version 13.0.0?) clang added a warning that catches variables that were assigned but unused. This version of clang (or later) is now bundled w/ github's macos images and is causing our tests to fail in continuous integration. We can currently generate C code that trips this warning, so for now I've disabled it as we do some other warnings related to variable usages.


scolsen:
Seems like the linux and nix errors were transient. This does fix the macOS CI again.

PR #1442, opened 2022-10-31
Document the memory management system further

scolsen:
Adds some details about the internals of the memory management system.

### DIFF
--- a/docs/Memory.md
+++ b/docs/Memory.md
@@ -192,6 +192,77 @@ to. The anonymous linear value `[1 2 3]` will be deleted at the end of the
 function scope, but the reference will be returned from the function, so its
 lifetime is potentially greater than that of the value it points to. 
 
+The memory management system performs two key checks around ref usage:
+
+1. Check that a newly created reference doesn't point to a linear value binding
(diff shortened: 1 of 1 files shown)


hellerve (approved):

eriksvedang (approved):
Nice writeup! I'm a little unsure of putting this into docs given that I don't think it's the entirely correct way of doing it, see https://github.com/carp-lang/Carp/issues/1317 for some ideas for improvements. But it's probably a nice help for some people, we just need to try to keep it updated if things change.

Also, I'm very open to any other ideas for making the whole "algorithm" nicer.

scolsen:
> Nice writeup! I'm a little unsure of putting this into docs given that I don't think it's the entirely correct way of doing it, see #1317 for some ideas for improvements. But it's probably a nice help for some people, we just need to try to keep it updated if things change.
> 
> Also, I'm very open to any other ideas for making the whole "algorithm" nicer.

That makes sense -- we could possibly put it in `hacking.md` instead so that it's clearer that it might be changed, but even then it's probably best to hold off on this until we've got an implementation that we feel is relatively stable.

PR #1451, opened 2023-01-27
refactor: change project flag configuration behavior

scolsen:
This PR changes the behavior of the project flag setting dynamic functions. Instead of appending a single string to the flagset, they now take a list of strings, which serve as the new flagset:

```
(Project.get-config "cflag")
=> (...a bunch of default platform flags; "-wall" ...)
(Project.config "cflag" '("-my" "-new" "-flags"))
(Project.get-config "cflag")
=> ("-my" "-new" "-flags")
```

In order to append flags, users now need to `cons` to a `config-get`:

```
(Project.config "cflag" (cons "-new" (Project.get-config "cflag")))
```

This is a **breaking change**, as passing a single string is now an error.

I considered overloading the behavior as an alternative so that we wouldn't break existing callers, but I feel this would ultimately lead to unintuitive behavior (calling with a single string appends while calling with a list of strings overrides ... users might wonder why the list doesn't append all the strings in the list instead).

This PR also fixes a bug w/ fetching project C flags and provides helper macros for appending flags to existing project flag sets given the new behavior.

### DIFF
--- a/core/Project.carp
+++ b/core/Project.carp
@@ -13,3 +13,13 @@
 Example usage: `(save-docs Int Float String)`")
 (defmacro save-docs [:rest modules]
   (eval (list 'save-docs-ex (list quote (collect-into modules array)) [])))
+
+(defmodule Project
+  (hidden append-flag)
+  (defndynamic append-flag [flag-set flag]
+    (Project.config flag-set (cons flag (Project.get-config flag-set))))
+
+  (defmacro append-cflag [flag] (Project.append-flag "cflag" flag))
+  (defmacro append-libflag [flag] (Project.append-flag "libflag" flag))
+  (defmacro append-pkgflag [flag] (Project.append-flag "pkgconfigflag" flag))
+)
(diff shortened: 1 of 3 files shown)


hellerve (approved):
🎉

eriksvedang (approved):
Looks good! Seems to break the tests though?

scolsen:
I'm wondering if we should just drop `-Wall` for our tests at this point in favor of an explicitly set list of reasonable warnings we want to make sure our emitted C complies with.

hellerve:
Possibly a good idea, but maybe we should also just remove usages of `sprintf`. I’ve done that in #1453 for your convenience :)

scolsen:
> Possibly a good idea, but maybe we should also just remove usages of `sprintf`. I’ve done that in #1453 for your convenience :)

Even better!

eriksvedang:
Does this need to pull in master for the tests to pass?

eriksvedang:
@scolsen Should we try to get this merged?

Issue #1427, opened 2022-08-15
(()) seems to cause an infinite loop on the latest release


Gavinok:
Haven't tried on master yet but just typing in `(())` into the repl causes an infinite loop for some reason.

PR #1456, opened 2023-02-03
fix: Infinite loop when evaluating `(())` (#1427)

Gavinok:
Evaluation of `(())` is treated as a malformed expression rather than causing an infinite loop.

### DIFF
--- a/src/Eval.hs
+++ b/src/Eval.hs
@@ -394,6 +394,7 @@ eval ctx xobj@(XObj o info ty) preference resolver =
     evaluateApp :: Evaluator
     evaluateApp (AppPat f' args) =
       case f' of
+        (ListPat []) -> pure (evalError ctx (format (GenericMalformed xobj)) (xobjInfo xobj))
         l@(ListPat _) -> go l ResolveLocal
         sym@(SymPat _ _) -> go sym resolver
         _ -> pure (evalError ctx (format (GenericMalformed xobj)) (xobjInfo xobj))


hellerve:
Thanks, this is super useful! 🎉 

Can we add a test for this? I suggest we put one in `test/test_for_errors` where we keep the expected failing code snippets. Then we add the expected output to `test/output`. See [this commit](https://github.com/carp-lang/Carp/commit/4e02c452dc44f901a03e3c548a0e579f083f935a) for an example.

Gavinok:
No problem I am just happy to help. Should be good to go. Let me know if I missed anything

hellerve:
Super! Maybe we should introduce a special error type to improve the error message—or maybe that would be overkill. What does everyone else think?

Gavinok:
Not exactly sure what a better fitting existing error message would look like. Could try something like the error message of used for applying non functions.
```
You are trying to call the non-function `()` at line 4, column 2 in 'REPL'. at REPL:4:1.
```
That's just an idea for reusing an existing error. Don't have any clever unique ideas personally.

hellerve:
Of the two, I personally prefer the non-function error message. But I’d wait for feedback from the others :)

eriksvedang:
Very nice! I also prefer the non-function error.

Gavinok:
@hellerve is there anyone else we are waiting to hear from?

hellerve:
No, I think we’re good to merge. @eriksvedang is the only one with those rights, though :)

eriksvedang:
The test suite fails on this branch – does it help if you merge in latest `master`?

PR #1458, opened 2023-02-26
fix: mangle member name in deletion generator

hellerve:
Before, the member deletion did not mangle the member name, so members named things like `default` (as is the case in https://github.com/carpentry-org/clig) would emit faulty C.

Cheers


scolsen:
Nice catch!

scolsen (approved):

PR #1467, opened 2025-09-11
fix(core): fix calls to snprintf

ghost:
The n argument to snprintf refers to the buffer's size. [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/fprintf.html)
The fault here is that it shrunk by one byte.

snprintf with n=1 will just result in a null byte

found this accidentally when compiling the output .c code statically
clang version 20.1.8 threw a warning that gcc 15.2.0 did not

while in search of this issue showing up, I figured out that the issue was not isolated to that singular snprintf call, but all of them in this commit.
Otherwise the only other snprintf usage (in SDLHelper.h) is fine.

bug landed in ca5774b1aeeab2bc4d6fa901b5a55682ebc2d539 - after the 0.5.5 release, which is still the latest

### DIFF
--- a/core/carp_pattern.h
+++ b/core/carp_pattern.h
@@ -568,13 +568,13 @@ Array Pattern_match_MINUS_all_MINUS_groups(Pattern *p, String *s) {
 String Pattern_internal_add_char(String a, Char b) {
     if (!a) {
         String buffer = CARP_MALLOC(2);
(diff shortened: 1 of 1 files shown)


eriksvedang:
@CorruptedVor Nice work, thanks!

PR #1470, opened 2025-09-11
fix(core): System_nanotime incorrectly returns double

CorruptedVor:
The cast to double leads to a potential loss of significant digits.
64-bit long should be enough, overflow in 2262-04-11 from 23:47:16 to 23:47:17
`0x7fffffffffffffff - (9223372037 * 1000000000) < 0, 9223372037 corresponds to 2262`

uint64_t is more appropriate, considering CLOCK_REALTIME is going to be returning timestamps after 1970 (positive), 32-bit long (the case on **32-bit platforms**) **suffer**s `(2^32 / 1000000000) < 5`, always **overflow**ing in today's age

from-type and to-type conversions:
only some of them are missing, it's trivial to add all of them
perhaps we should make the carp compiler output the definitions automatically


eriksvedang:
Hi, sorry I missed the notifications for your PRs. Will look into them asap! Cheers

PR #1518, opened 2026-03-25
feat(core): add more IO extensions (exists?, mkdir, rmdir, cwd)

sqrew:
Adds several cross-platform IO extensions to the standard library:
- `IO.file-exists?` and `IO.dir-exists?`
- `IO.create-dir` (wraps mkdir)
- `IO.remove-dir` (wraps rmdir)
- `IO.get-cwd` (wraps getcwd)
- `IO.set-cwd` (wraps chdir)

Includes corresponding cross-platform C wrappers in `core/carp_io.h` and comprehensive tests in `test/io.carp`.


core/carp_io.h
```diff
@@ -1,3 +1,13 @@
+#include <sys/stat.h>
+#include <dirent.h>
+#ifdef _WIN32
+#include <direct.h>
+#define getcwd _getcwd
+#define chdir _chdir
+#else
```
hellerve, line 9: It looks like Windows is not happy with this change quite yet.

hellerve (changes requested):

sqrew:
Added cross-platform Windows implementation for `list-dir` and `get-cwd`, and fixed the missing headers in `carp_io.h`. Consolidated tests into `test/io.carp`. Thanks!

core/carp_io.h
```diff
@@ -53,3 +66,99 @@ String IO_unsafe_MINUS_read_MINUS_file(const String* filename) {
...
+            }
+            String name = CARP_MALLOC(strlen(dir->d_name) + 1);
+            strcpy(name, dir->d_name);
+            result.len++;
+            result.data = CARP_REALLOC(result.data, sizeof(String) * result.len);
+            ((String*)result.data)[result.len - 1] = name;
+        }
+        closedir(d);
+    } else {
+        result.len = -1;
```
hellerve, line 129: I don’t think returning an array of length -1 is safe or sensible here. We should find a different error condition (also for all the functions on top of this one).

core/IO.carp
```diff
@@ -204,8 +222,85 @@
...
+    (register get-cwd (Fn [] (Ptr CChar)) "IO_Raw_get_MINUS_cwd")
+    (register set-cwd (Fn [&String] Int) "IO_Raw_set_MINUS_cwd")
   )
 
+  (doc file-exists? "Checks if a file exists at the given path.")
+  (register file-exists? (Fn [&String] Bool))
+
+  (doc dir-exists? "Checks if a directory exists at the given path.")
+  (register dir-exists? (Fn [&String] Bool))
+
```
hellerve, line 252: Those should probably also be in `IO.Raw`.

core/IO.carp
```diff
@@ -204,8 +222,85 @@
...
+  (doc get-cwd
+    "Returns the current working directory."
+    ""
+    "Returns a (Result String String) indicating success or failure.")
+  (defn get-cwd []
+    (let [res (IO.Raw.get-cwd)]
+      (if (null? res)
+        (Result.Error (fmt "Failed to get current working directory: %s" &(System.error-text)))
+        (let [s (String.from-cstr res)]
```
hellerve, line 261: The C side already returns a `String`. We should fix the function signature on the C side to avoid a double allocation.

core/IO.carp
```diff
@@ -329,6 +424,32 @@
...
+    (if (Int.= (IO.Raw.unlink filename) 0)
+      (Result.Success true)
+      (Result.Error (fmt "Failed to remove file '%s': %s" filename &(System.error-text)))))
+
+  (doc copy-file
+    "Copies a file from `src` to `dest`. Overwrites `dest` if it already exists."
+    ""
+    "Returns a (Result Bool String) indicating success or failure.")
+  (defn copy-file [src dest]
+    (Result.and-then (read-file src)
```
hellerve, line 450: This reads the complete file into memory. We should probably do a chunked read-write.

core/carp_io.h
```diff
@@ -53,3 +66,99 @@ String IO_unsafe_MINUS_read_MINUS_file(const String* filename) {
...
+    }
+    return result;
+}
+#endif
+
+int IO_mkdir(const String* path) {
+#ifdef _WIN32
+    return _mkdir(*path);
+#else
+    return mkdir(*path, 0777);
```
hellerve, line 139: Is using `0777` really desirable here?

core/IO.carp
```diff
@@ -238,7 +333,7 @@
 
   (doc open-file
     "Opens a [FILE](#file) with the given name using a designated mode"
-    "(e.g. [r]ead, [w]rite, [a]ppend),  [rb] read binary...)."
+    "(e.g. [r]ead, [w]rite, [a]ppend),  [rb] read binary...."
```
hellerve, line 336: We open a paren here, so we should also close it.

test/io.carp
```diff
@@ -21,20 +22,94 @@
                "getenv works on existant variable"
   )
   (let [ data "Icke\n\tdette\n\tkieke mal,\nOochen, Flesch und Beene." ]  ; include \n in test data!
-    (assert-equal test
-                  data
-                  &(write-then-read data "io_carp_testdata.txt")
-                  "write-file then read-file"  ))
+    (let [res (write-then-read data "io_carp_testdata.txt")]
```
hellerve, line 25: Why the `let` change here? Should be fine without?

test/io.carp
```diff
@@ -21,20 +22,94 @@
...
       (ignore (IO.append-file second file-name))
       (let-do [ result (IO.read-file file-name) ]
-        (IO.Raw.unlink! file-name)
-        (assert-equal test
-                      "hello world"
-                      &(Result.unsafe-from-success result)
-                      "append-file appends to existing file"))))
+        (do
+          (IO.Raw.unlink! file-name)
+          (let [s (Result.unsafe-from-success result)]
```
hellerve, line 39: Same let-change here. Why? We also don’t need a `do` inside a `let-do`.

hellerve (changes requested):

core/IO.carp
```diff
@@ -204,8 +222,97 @@
...
+  (doc dir-exists? "Checks if a directory exists at the given path.")
+  (defn dir-exists? [path]
+    (IO.Raw.dir-exists? path))
+
+  (doc get-cwd
+    "Returns the current working directory."
+    ""
+    "Returns a (Result String String) indicating success or failure.")
+  (defn get-cwd []
+    (let [res (IO.Raw.get-cwd)]
```
hellerve, line 272: Ths can return `NULL`. We _need_ to check for that.

core/IO.carp
```diff
@@ -318,17 +425,67 @@
...
+      (Result.Success true)
+      (Result.Error (fmt "Failed to remove file '%s': %s" filename &(System.error-text)))))
+
+  (doc copy-file
+    "Copies a file from `src` to `dest`. Overwrites `dest` if it already exists."
+    ""
+    "Returns a (Result Bool String) indicating success or failure.")
+  (defn copy-file [src dest]
+    (let [f-src? (open-file src "rb")
+          f-dest? (open-file dest "wb")]
```
hellerve, line 463: Ths will create and open the file before we validated the source is valid.

core/IO.carp
```diff
@@ -318,17 +425,67 @@
...
+      (if (Result.error? &f-src?)
+        (Result.Error (fmt "Could not open source file '%s': %s" src &(Result.unsafe-from-error f-src?)))
+        (if (Result.error? &f-dest?)
+          (do
+            (ignore (IO.Raw.fclose (Result.unsafe-from-success f-src?)))
+            (Result.Error (fmt "Could not open destination file '%s': %s" dest &(Result.unsafe-from-error f-dest?))))
+          (let-do [f-src (Result.unsafe-from-success f-src?)
+                   f-dest (Result.unsafe-from-success f-dest?)
+                   buf-size 4096
+                   buffer (String.allocate buf-size \0)
```
hellerve, line 473: Can we validate the allocation like we do in `read-file`?

core/carp_io.h
```diff
@@ -53,3 +68,98 @@ String IO_unsafe_MINUS_read_MINUS_file(const String* filename) {
...
+
+#ifdef _WIN32
+int IO_Raw_list_MINUS_dir(const String* path, Array* out_result) {
+    WIN32_FIND_DATA fd;
+    char search_path[MAX_PATH];
+    snprintf(search_path, MAX_PATH, "%s\\*", *path);
+    HANDLE hFind = FindFirstFile(search_path, &fd);
+    if (hFind != INVALID_HANDLE_VALUE) {
+        out_result->len = 0;
+        out_result->capacity = 0;
```
hellerve, line 96: We need to update capacity!

core/carp_io.h
```diff
@@ -53,3 +68,98 @@ String IO_unsafe_MINUS_read_MINUS_file(const String* filename) {
...
+        out_result->data = NULL;
+        while ((dir = readdir(d)) != NULL) {
+            if (strcmp(dir->d_name, ".") == 0 || strcmp(dir->d_name, "..") == 0) {
+                continue;
+            }
+            String name = String_from_MINUS_cstr(dir->d_name);
+            out_result->len++;
+            out_result->data = CARP_REALLOC(out_result->data, sizeof(String) * out_result->len);
```
hellerve, line 128: Don’t reallocate on every directory, this is quadratic. Grow geometrically like arrays do.

hellerve (changes requested):

sqrew:
IO is more complicated than I anticipated so I will close this in favor of someone more qualified attempting it

PR #1522, opened 2026-04-02
perf: make String.lines faster

hellerve:
This PR makes `String.lines` faster by not using patterns but a regular string indexing.

Using this script:

```clojure
(load "Bench.carp")

(def big-string @"")

(defn-do setup []
  (set! big-string @"")
  (for [i 0 1000]
    (set! big-string (String.append &big-string "this is a line of moderate length for benchmarking\n"))))

(defn-do main []
  (setup)
  (Bench.bench (fn [] (String.lines &big-string))))
```

I observe these numbers:

```
# before
Total time elapsed: 37.8041s
Best case: 373.605ms
Worst case: 386.365ms
Standard deviation: 1.09158ms

# after
Total time elapsed: 904.533ms
Best case: 1.23942ms
Worst case: 1.2958ms
Standard deviation: 4.62482µs
```

Or a 373x speedup.

Cheers

### DIFF
--- a/core/Pattern.carp
+++ b/core/Pattern.carp
@@ -180,24 +180,4 @@ list of those characters.")
   (doc collapse-whitespace "collapses groups of whitespace into single spaces.")
   (defn collapse-whitespace [s]
     (Pattern.substitute #"\s+" s " " -1))
-
-  (doc split-by "splits a string by separators.")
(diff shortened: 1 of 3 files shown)


eriksvedang (approved):

eriksvedang:
Great work!

PR #1523, opened 2026-04-02
Port #1440: ensure captured values can’t be given away

hellerve:
This PR picks up #1440 by @scolsen to our new system (thank you!). The cool thing is that it’s now much simpler, boiling down to a literal 3 line fix! We only have to check for captures in `unmanage`, and that’s it!

Cheers


eriksvedang (approved):
YES!!

PR #1524, opened 2026-04-08
Fix/fix eval primitives

hellerve:
This PR moves primitives over into commands where applicable to reduce the amount of `eval` calls we need to get anywhere.

Cheers


hellerve:
I’m redoing this.

hellerve:
I’m redoing this from scratch.

PR #1527, opened 2026-04-09
feat(stdlib): improve String module with functional API and optimizat…

sqrew:
- Added String.map, String.filter, String.any?, String.all?, String.reduce
- Added String.replace utility
- Optimized starts-with? and ends-with? with efficient C implementations
- Integrated tests for new functionality into test/string.carp
- Updated CHANGELOG.md

### DIFF
--- a/core/String.carp
+++ b/core/String.carp
@@ -50,7 +50,7 @@
   (doc from-chars "Creates a string from an array of characters.")
   (register from-chars (Fn [&(Array Char)] String))
   (doc tail "Returns the tail of a string (all characters except the first).")
-  (register tail (λ [(Ref String)] String))
+  (register tail (Fn [&String] String))
   (doc format "Formats a string.")
   (register format (Fn [&String &String] String))
   (implements format String.format)
--- a/core/String.carp
+++ b/core/String.carp
@@ -133,15 +133,9 @@
     (from-chars &(Array.suffix &(chars s) b)))
 
   (doc starts-with? "Check if the string `s` begins with the string `sub`.")
-  (defn starts-with? [s sub]
-    (let [ls (length sub)]
-      (and (>= (length s) ls) (= sub &(prefix s ls)))))
-
+  (register starts-with? (Fn [&String &String] Bool))
   (doc ends-with? "Check if the string `s` ends with the string `sub`.")
-  (defn ends-with? [s sub]
-    (let [ls (length s)
-          lsub (length sub)]
-      (and (>= ls lsub) (= sub &(suffix s (- ls lsub))))))
+  (register ends-with? (Fn [&String &String] Bool))
 
   (doc zero "The empty string.")
   (defn zero [] @"")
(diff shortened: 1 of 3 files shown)


core/String.carp
```diff
@@ -152,7 +146,7 @@
     (let-do [sum 0
              lstrings (Array.length strings)]
       (for [i 0 lstrings]
-        (set! sum (+ sum (String.length (Array.unsafe-nth strings i)))))
+        (set! sum (+ sum (length (Array.unsafe-nth strings i)))))
```
hellerve: Why did you change that? Will this genericize the function?

core/String.carp
```diff
@@ -240,6 +234,47 @@
   (doc lines "splits a string into lines.")
   (defn lines [s]
     (split-by s &[\newline]))
+
+  (doc map "Maps a function over the characters of a string, returning a new string.")
+  (defn map [f s]
```
hellerve: Should we call this `copy-map`, to be explicit like in the array module?

core/String.carp
```diff
@@ -240,6 +234,47 @@
...
+
+  (doc all? "Checks if all characters in a string satisfy a predicate.")
+  (defn all? [f s]
+    (Array.all? f &(chars s)))
+
+  (doc reduce "Reduces a string using a function and an initial value.")
+  (defn reduce [f init s]
+    (Array.reduce f init &(chars s)))
+
+  (doc replace "Replaces all occurrences of a substring in a string with another string.")
```
hellerve: Shouldn’t we leave this to the Pattern module?

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@
 - fix: #1347 by ignoring generically typed symbols on printing C (#1373)
 - fix: nix install by using correct pkg-configDepends config key (#1372)
 - fix: type signature of Array.unsafe-raw (#1375)
+- feat: add map, filter, any?, all?, reduce, and replace to String (#XXXX)
```
hellerve: The changelog entries are in the wrong version, and shouldn’t be part of the PR.

hellerve (changes requested):

core/String.carp
```diff
@@ -240,6 +234,47 @@
...
+  (doc all? "Checks if all characters in a string satisfy a predicate.")
+  (defn all? [f s]
+    (Array.all? f &(chars s)))
+
+  (doc reduce "Reduces a string using a function and an initial value.")
+  (defn reduce [f init s]
+    (Array.reduce f init &(chars s)))
+
+  (doc replace "Replaces all occurrences of a literal substring in a string with another string.")
+  (defn replace [s needle replacement]
```
hellerve: This is quadratic and allocation-heavy. Can we optimize? Also, we can collapse patterns like `let`+`do` into `let-do`.

hellerve (changes requested):
We should also mention that these all operate on `chars`, not `bytes` (as we do in `chars` itself).

PR #1528, opened 2026-04-11
perf: parallelize the test script

hellerve:
Disclaimer: I let Claude take the wheel on this.

This PR adds parallelization options to the test script, speeding up the test feedback loop (it runs in just over 100s on my machine now).

Cheers

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -11,4 +11,4 @@ tasks:
      stack test
   - carp-test: |
      cd Carp
-     scripts/run_carp_tests.sh --no_sdl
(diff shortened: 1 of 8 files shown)


eriksvedang:
Do I need to do anything to get this to take effect? I'm not seeing any speedup at all... (I'm on a mac with M5 pro)

hellerve:
By default it’s still serial. You’ll have to use the flag `-j <n>` (where n is number of parallel workers) to notice a speedup.

eriksvedang:
@hellerve Thanks, that makes sense. The tests run in 30 seconds for me now, that's amazing!

PR #1529, opened 2026-04-11
feat(stdlib): add variadic arithmetic macros and in-place inc/dec

sqrew:
This PR introduces several ergonomic improvements to the arithmetic facilities in the Carp standard library.

### Changes:
- **Variadic Macros:** Added `sum`, `product`, `sub`, and `div`. These allow for more idiomatic Lisp-style arithmetic like `(sum 1 2 3 4)`.
- **Unary Negation:** `(sub x)` now correctly expands to `(neg x)`.
- **In-place Operations:** Added `inc!` and `dec!` macros for cleaner variable updates.
- **Robust Expansion:** Implemented a recursive left-associative expansion (`build-vararg-left`) to ensure correct results for non-commutative operations like subtraction and division.
- **Tests:** Integrated 10 new test cases into `test/macros.carp`.

---


hellerve (commented):
I like this in general!

One thing: if we want to express that integers form a ring, we should handle identity in `sum` and `product`(0 and 1, respectively). That’s not necessary.

hellerve:
Ci seems unhappy. 365m runtime seems, uh, excessive.

core/Macros.carp
```diff
@@ -432,3 +436,33 @@ the filename and module name are the same.")
...
+(defmacro div [:rest forms]
+  (build-vararg-left '/ forms))
+
+(doc inc! "Increments a variable in-place.")
+(defmacro inc! [var]
+  (list 'set! var (list 'inc var)))
+
+(doc dec! "Decrements a variable in-place.")
+(defmacro dec! [var]
+  (list 'set! var (list 'dec var)))
```
hellerve, line 468: I just realized we already have `++` and `--`, which do the same thing. I actually like this naming better. What do the others think? I think one should go in favor of the other in any case.

sqrew: The thing I like about inc!/dec! is that the ! makes it clear it's an in place mutation and that it's a macro. I personally prefer ++ (because it's cooler) and totally spaced that it was even here. I'm totally open to either option honestly.

sqrew:
Turns out the CI issues were name conflicts causing a hang up during macro expansion with common names like sum/sub/mul/div etc, so I've added a -variadic suffix (ie sum-variadic, etc) to each of them. This definitely makes it ugly as hell. Unsure if we really want them added to the language in that form. Thoughts from you guys?

hellerve:
Yeah, not very pretty! But the hang during macro expansion sounds bad, we should definitely fix that. Sorry you had to debug that!

PR #1530, opened 2026-04-12
Two phase eval

hellerve:
This PR splits eval into phases (expand and execute) to reduce the amount of `eval` we need in macros and also fix the REPL such that `(println* "hi")` will now actually print and not just expand. You can still expand using `macro-expand`.

Cheers


eriksvedang (approved):
Awesome!

PR #1533, opened 2026-04-13
fix: Make guessing game example work again (trim input).

eriksvedang:
I tried running the examples and the guessing game didn't work properly, here's a fix.


hellerve:
Should we add it to compile-only tests to make sure this doesn’t happen again?

eriksvedang:
I think it's already added there, this was a runtime error when playing the game (which doesn't happen that often, perhaps :P)

hellerve (approved):

PR #1535, opened 2026-04-15
chore: Bump stack to lts-22.44 (ghc 9.6.7)

eriksvedang:


hellerve:
I believe that Nix build was a fluke and can be retried.

eriksvedang:
It's running into this issue https://github.com/NixOS/nixpkgs/issues/41340 (or something similar)

eriksvedang:
So what seems to happen is that nix (with the help from pkg-config) passes +800 libraries to gcc for linking, which it doesn't like. Not sure why it worked on older nix versions.

eriksvedang:
OK, now it passes but doesn't run any SDL examples on Nix. For me, that's acceptable – any complaints?

hellerve:
That’s fine by me.

hellerve (approved):

PR #1538, opened 2026-04-17
refactor(compiler): resolve original variable names in borrow checker errors

sqrew:
This PR significantly improves the clarity of borrow checker error messages by mapping internal compiler-generated names (like `_16`) back to the original variable names defined by the user.

### Changes:
- **Internal Mapping:** Added a `memStateNames` map to the borrow checker's state (`MemState`) to track the relationship between internal expression IDs and user-defined names.
- **Metadata Capture:** Updated the `manage` and lambda-capture logic in `Memory.hs` to populate this map whenever a new variable is encountered.
- **Error Enhancement:** Modified the `UsingDeadReference` error in `TypeError.hs` to prioritize the original name if available.
- **UX Improvement:** Errors now read `The reference 'r' (depending on the variable 'my-important-string') isn't alive` instead of referring to cryptic internal IDs.

---
**Disclaimer:** This PR was built with the assistance of Gemini (an AI assistant) working in collaboration with @sqrew to improve the Carp compiler's diagnostics.

### DIFF
--- a/src/Memory.hs
+++ b/src/Memory.hs
@@ -23,7 +23,8 @@ data MemState = MemState
   { memStateDeleters :: Set.Set Deleter,
     memStateDeps :: Set.Set Ty,
     memStateLifetimes :: Map.Map String LifetimeMode,
-    memStateParamDeleters :: Set.Set Deleter
+    memStateParamDeleters :: Set.Set Deleter,
+    memStateNames :: Map.Map String String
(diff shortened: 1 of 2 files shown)


hellerve (approved):
Nice! Thanks!

eriksvedang (approved):
Awesome.

PR #1539, opened 2026-04-17
fix: filter out stub binders for multi-candidate implementations

hellerve:
This PR fixes spurious "Too many '<fun>' functions found ... can't figure out dependencies" by removing stub binders from the check.

Cheers


eriksvedang (approved):
Makes sense!

PR #1540, opened 2026-04-19
feat(stdlib): add find-last, take/drop, split-by-predicate, and variadic zip-n to Array

sqrew:
This PR introduces several common functional and search utilities to the `Array` module, bringing it closer to parity with other functional standard libraries.

### Changes:
- **Reverse Searching:** Added `find-last` and `find-last-index` for searching from the end of an array.
- **Slicing Helpers:** Added `take` and `drop` for simple index-based array truncation.
- **Partitioning:** Added `split-by-predicate` to divide an array into matching and non-matching elements (returns a `Pair`).
- **Variadic Zipping:** Added a powerful `zip-n` macro that allows zipping an arbitrary number of arrays with a multi-argument function.
- **Tests:** Integrated 8 new test cases into `test/array.carp`.

---
**Disclaimer:** This PR was built with the assistance of Gemini (an AI assistant) working in collaboration with @sqrew to improve the Carp standard library consistency.

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -116,6 +116,30 @@ If it doesn’t find an index, `Nothing` will be returned.")
             (break))))
       ret))
 
+  (doc find-last "finds the last element in `a` that matches the function `f` and wraps it in a `Just`.
+
+If it doesn’t find an element, `Nothing` will be returned.")
+  (defn find-last [f a]
+    (let-do [res (Maybe.Nothing)]
+      (for [i (Int.dec (length a)) -1 -1]
+        (when (~f (unsafe-nth a i))
+          (do
+            (set! res (Maybe.Just @(unsafe-nth a i)))
+            (break))))
+      res))
+
+  (doc find-last-index "finds the index of the last element in `a` that matches the function `f` and wraps it in a `Just`.
+
+If it doesn’t find an index, `Nothing` will be returned.")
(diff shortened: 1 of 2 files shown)


core/Array.carp
```diff
@@ -595,8 +627,52 @@ Example:
...
+  (doc split-by-predicate
+    "splits an array `a` into two arrays: those that satisfy the predicate `f` and those that do not.
+Returns a `Pair` of arrays.")
+  (defn split-by-predicate [f a]
+    (let-do [matching []
+             not-matching []]
+      (for [i 0 (length a)]
+        (let [x (unsafe-nth a i)]
+          (if (~f x)
+            (set! matching (push-back matching @x))
```
eriksvedang, line 640: Could use `push-back!` here for some speedup, since the function uses (interior) mutation anyways?

test/array.carp
```diff
@@ -435,4 +435,51 @@
                 42
                 (unsafe-nth-value &[42 43] 0)
                 "unsafe-nth-value works as expected")
+
+  (assert-equal test
```
eriksvedang, line 439: This test doesn't really prove that the right element was found, maybe use modulus instead of equals on a list with distinct elements?

eriksvedang (commented):
Nice!

eriksvedang (approved):

PR #1541, opened 2026-04-19
feat(stdlib): add Array.flatten

sqrew:
This PR adds `Array.flatten` to the standard library as a descriptive alias for `Array.concat`.

### Changes:
- **Flattening:** Added `Array.flatten` to collapse nested arrays.
- **Tests:** Integrated a new test case into `test/array.carp`.

---
**Disclaimer:** This PR was built with the assistance of Gemini (an AI assistant) working in collaboration with @sqrew to improve the Carp standard library consistency.

### DIFF
--- a/core/Array.carp
+++ b/core/Array.carp
@@ -363,6 +363,10 @@ The trailing elements of the longer array will be discarded.")
           (set! j (+ j len))))
       result))
 
+  (doc flatten "collapses a nested array of arrays into a single array. (Alias for [`concat`](#concat)).")
+  (defn flatten [xs]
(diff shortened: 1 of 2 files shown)


hellerve:
I’m not sure we want two names for the same thing, might be too much of a pollutant. What do others think?

eriksvedang:
> I’m not sure we want two names for the same thing, might be too much of a pollutant. What do others think?

Yeah I agree, I don't think we should have multiple names for things.

sqrew:
Totally reasonable position. I'll close this PR. Thanks.

PR #1543, opened 2026-04-20
feat(stdlib): add if-let and while-let macros

sqrew:
Hey! I noticed that Carp was missing some basic Lisp staples for handling `Maybe` values more easily, so I've implemented `if-let` and `while-let`.

These macros make it much cleaner to unwrap an optional value without needing a full `match` block every time.

### What's new:
- **if-let:** Perfect for when you only want to do something if a `Maybe` is `Just`.
- **while-let:** A handy way to loop through anything that returns a `Maybe`.
- **Tests:** I've added a few tests to `test/macros.carp` to make sure they're solid.

---

### DIFF
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -98,8 +98,8 @@ Example:
      ""
      "```"
      "(doto-ref @\"hi\""
-     "  (string-set! 0 \o)"
-     "  (string-set! 1 \y))"
+     "  (string-set! 0 \\o)"
+     "  (string-set! 1 \\y))"
      "```")
 (defmacro doto-ref [thing :rest expressions]
   (let [s (gensym)]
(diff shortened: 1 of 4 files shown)


sqrew:
Just realized that perhaps this probably should have gone into controlmacros.carp instead. Oopsies. Thoughts?

sqrew:
I went ahead and moved them to Controlmacros.carp and also added when-let as well since we were here anyways

eriksvedang (approved):

core/Macros.carp
```diff
@@ -217,6 +217,7 @@ other expressions, otherwise it returns the value of the last form in `xs`.
   (cond-internal xs))
 
 (doc refstr "stringifies `x` and takes the reference of that string.")
+
```
hellerve, line 220: Can we get rid of that empty line change to get the diff a bit cleaner?

test/macros.carp
```diff
@@ -401,4 +401,26 @@
...
+  (let-do [counter 0
+           res []]
+    (while-let [x (if (< counter 3) (Maybe.Just counter) (the (Maybe Int) (Maybe.Nothing)))]
+      (do
+        (set! res (Array.push-back res x))
+        (set! counter (inc counter))))
+    (assert-equal test
+                  &[0 1 2]
+                  &res
+                  "while-let works as expected"))
```
hellerve: I think this can be removed since you moved the tests also?

core/ControlMacros.carp
```diff
@@ -140,6 +140,35 @@ Example:
  (defmacro ignore-do [:rest forms]
    (cons 'do (expand (apply ignore* forms))))
 
+(doc if-let "Conditional execution based on a `Maybe` value.
+If `expr` evaluates to `(Just x)`, the `then` branch is executed with `var` bound to `x`.
```
hellerve: The user can’t see `expr` and `var`, so this explanation might be confusing.

core/ControlMacros.carp
```diff
@@ -140,6 +140,35 @@ Example:
...
+Otherwise, the `else` branch is executed.")
+(defmacro if-let [bindings then else]
+  (let [var (car bindings)
+        expr (cadr bindings)]
+    `(match %expr
+       (Maybe.Just %var) %then
+       (Maybe.Nothing) %else)))
+
+(doc while-let "Looping execution based on a `Maybe` value.
+Repeatedly evaluates `expr`. As long as it returns `(Just x)`, the `body` is executed with `var` bound to `x`.")
```
hellerve: Same as above.

core/ControlMacros.carp
```diff
@@ -140,6 +140,35 @@ Example:
...
+(defmacro while-let [bindings body]
+  (let [var (car bindings)
+        expr (cadr bindings)]
+    `(while true
+       (match %expr
+         (Maybe.Just %var) %body
+         (Maybe.Nothing) (break)))))
+
+(doc when-let "Conditional execution based on a `Maybe` value.
+If `expr` evaluates to `(Just x)`, the `body` is executed with `var` bound to `x`.")
```
hellerve: Same as above.

test/macros.carp
```diff
@@ -401,4 +401,26 @@
...
+                &(if-let [x (Maybe.Just 10)]
+                  (fmt "found %d" x)
+                  @"not found")
+                "if-let works with Just")
+  (assert-equal test
+                "not found"
+                &(if-let [x (the (Maybe Int) (Maybe.Nothing))]
+                  (fmt "found %d" x)
+                  @"not found")
+                "if-let works with Nothing")
```
hellerve: I was wrong! This one needs to move over.

hellerve (changes requested):

core/ControlMacros.carp
```diff
@@ -125,6 +125,91 @@ Example:
   (list 'while condition
         (cons 'do forms)))
 
+;; Note: if-let/when-let/while-let are specialized for Maybe and Result types to ensure
+;; type safety. In Carp, a single match block cannot handle both Maybe.Just
+;; and Result.Success constructors simultaneously.
```
hellerve: "a single match block cannot handle both Maybe.Just and Result.Success constructors simultaneously" is tautological, since they're different types, no?

hellerve (changes requested):
Two of my comments are seemingly still unaddressed.

sqrew:
oh god guys git is hard and I made a very ugly commit history by accident. sorry. unsure how to make it prettier.

eriksvedang:
> oh god guys git is hard and I made a very ugly commit history by accident. sorry. unsure how to make it prettier.

that's fine, it'll get squashed

hellerve:
It looks like a bunch of my comments are still relevant here.

PR #1545, opened 2026-04-22
feat(stdlib): add threading, conditional, and monadic macros

sqrew:
This PR introduces a suite of powerful control-flow macros to the Carp standard library, enhancing ergonomics for data threading, conditional logic, and monadic operations.

### Changes:
- **as-> / as->>:** Flexible threading macros that name the value, enabling its use in any argument position.
- **cond-let:** A multibranch conditional for `Maybe` bindings, providing clean unwrapping for multiple options.
- **cond-> / cond->>:** Conditional threading macros that apply transformations only when their corresponding tests are true.
- **maybe-and-then-> / result-and-then->:** Advanced monadic chaining macros for steps that themselves return `Maybe` or `Result`.
- **Tests:** Integrated 11 new test cases into `test/control_macros.carp`.

These macros significantly improve the ergonomics of complex data pipelines and conditional logic.

---

### DIFF
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -60,6 +60,19 @@ Example:
 (defmacro --> [:rest forms]
   (thread-last-internal forms))
 
+(hidden as->-internal)
+(defndynamic as->-internal [name forms]
+  (if (= 0 (length forms))
+    name
+    `(let [%name %(car forms)]
+       %(as->-internal name (cdr forms)))))
+
+(doc as-> "Threading macro with an explicit name.
+Binds `value` to `name` and then threads it through each of the `forms`, re-binding the result to `name` at each step.")
+(defmacro as-> [value name :rest forms]
+  `(let [%name %value]
+     %(as->-internal name forms)))
+
 (hidden comp-internal)
 (defndynamic comp-internal [sym fns]
   (if (= (length fns) 0)
@@ -185,6 +198,26 @@ Example:
   (let [name (gensym)]
     (list 'let [name form]
       (case-internal name branches))))
+
+(hidden cond-let-internal)
+(defndynamic cond-let-internal [xs]
+  (if (= (length xs) 0)
+    (list)
+    (if (= (length xs) 1)
+      (car xs)
+      (let [binding (car xs)
+            var (car binding)
+            expr (cadr binding)]
+        `(match %expr
+           (Maybe.Just %var) %(cadr xs)
+           (Maybe.Nothing) %(cond-let-internal (cddr xs)))))))
+
+(doc cond-let "Multibranch conditional with Maybe bindings.
+Each branch is a `[var expr]` binding and a body.
+If `expr` evaluates to `(Just x)`, `body` is executed with `var` bound to `x`.
+The final argument is an optional else branch.")
+(defmacro cond-let [:rest xs]
+  (cond-let-internal xs))
 
 (defmodule Dynamic
   (doc flip
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -73,6 +73,97 @@ Binds `value` to `name` and then threads it through each of the `forms`, re-bind
   `(let [%name %value]
      %(as->-internal name forms)))
 
+(hidden as->>-internal)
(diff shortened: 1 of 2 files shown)


core/ControlMacros.carp
```diff
@@ -60,6 +60,110 @@ Example:
...
+(defmacro cond->> [val :rest steps]
+  (cond-last-internal val steps))
+
+(doc some-> "Monadic threading macro (first).
+Threads the unwrapped value of `val` (a Maybe) through each `form` in `steps`.
+Wraps the final result in a `Maybe.Just`. If `val` is `Nothing`, returns `Nothing`.
+Each form in `steps` must return a value, not a Maybe.")
+(defmacro some-> [val :rest steps]
+  `(match %val
+     (Maybe.Just x) (Maybe.Just (-> x %@steps))
```
hellerve: Shouldn’t this call `some->` here?

core/ControlMacros.carp
```diff
@@ -60,6 +60,110 @@ Example:
...
+     (Maybe.Nothing) (Maybe.Nothing)))
+
+(doc some->> "Monadic threading macro (last).
+Threads the unwrapped value of `val` (a Maybe) through each `form` in `steps`.
+Wraps the final result in a `Maybe.Just`. If `val` is `Nothing`, returns `Nothing`.
+Each form in `steps` must return a value, not a Maybe.")
+(defmacro some->> [val :rest steps]
+  `(match %val
+     (Maybe.Just x) (Maybe.Just (--> x %@steps))
```
hellerve: Shouldn’t this call `some->>` here?

core/ControlMacros.carp
hellerve: Also can we gensym here to avoid the hygiene leak with `x`?

core/ControlMacros.carp
```diff
@@ -60,6 +60,110 @@ Example:
...
+     (Maybe.Nothing) (Maybe.Nothing)))
+
+(hidden maybe-and-then-internal)
+(defndynamic maybe-and-then-internal [acc steps]
+  (if (empty? steps)
+    acc
+    (maybe-and-then-internal
+      `(match %acc
+         (Maybe.Just x) (-> x %(car steps))
```
hellerve: Can we gensym here too?

core/ControlMacros.carp
```diff
@@ -60,6 +60,110 @@ Example:
...
+(defmacro maybe-and-then-> [val :rest steps]
+  (maybe-and-then-internal val steps))
+
+(hidden result-and-then-internal)
+(defndynamic result-and-then-internal [acc steps]
+  (if (empty? steps)
+    acc
+    (result-and-then-internal
+      `(match %acc
+         (Result.Success x) (-> x %(car steps))
```
hellerve: And here also a gensym? (And the `e` below)

core/ControlMacros.carp
```diff
@@ -60,6 +60,110 @@ Example:
...
+  (if (= 0 (length steps))
+    acc
+    (let [test (car steps)
+          form (cadr steps)
+          new-acc `(if %test (-> %acc %form) %acc)]
+      (cond-first-internal new-acc (cddr steps)))))
+
+(doc cond-> "Conditional threading macro (first).
+Threads `val` through each `form` in `steps` only if the corresponding `test` is true.")
+(defmacro cond-> [val :rest steps]
```
hellerve, line 100: We should check the numbers of forms is even here.

core/ControlMacros.carp
```diff
@@ -186,6 +290,26 @@ Example:
     (list 'let [name form]
       (case-internal name branches))))
 
+(hidden cond-let-internal)
+(defndynamic cond-let-internal [xs]
+  (if (= (length xs) 0)
+    (list)
```
hellerve: This restricts the typing. It should be a `Maybe.Nothing`, right?

hellerve (changes requested):

sqrew:
One thing I noticed: I use -> and ->> whereas the stdlib uses both -> and --> for first and last threading. I wonder if I should alter these to follow the same convention. Anyone have input on this?

PR #1546, opened 2026-04-22
feat(stdlib): add Maybe and Result ergonomic helpers

sqrew:
This PR adds several common functional and ergonomic helpers to the `Maybe` and `Result` modules.

### Changes:
- **Flattening:** Added `flatten` to both modules to handle nested optionality/errors.
- **Unwrapping:** Added `expect` (as a macro to avoid circular `IO` dependencies) for unwrapping with a custom panic message.
- **Inspection:** Added `inspect` (and `inspect-error` for Result) for performing side-effects without consuming the value.
- **Conditional Mapping:** Added `map-or` and `map-or-else` for more concise transformation-with-default patterns.
- **Tests:** Integrated 14 new test cases across `test/maybe.carp` and `test/result.carp`.

---

### DIFF
--- a/test/maybe.carp
+++ b/test/maybe.carp
@@ -165,4 +165,41 @@
                 &(the (Maybe Int) (Nothing))
                 &(filter (the (Maybe Int) (Nothing)) &(fn [x] (Int.even? @x)))
                 "filter works on Nothing")
+
+  (assert-equal test
+                &(Just 10)
+                &(flatten (Just (Just 10)))
+                "flatten works on Just Just")
+  (assert-equal test
+                &(the (Maybe Int) (Nothing))
+                &(flatten (the (Maybe (Maybe Int)) (Just (Nothing))))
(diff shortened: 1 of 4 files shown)


eriksvedang (approved):

test/maybe.carp
```diff
@@ -165,4 +165,41 @@
...
+                &(the (Maybe Int) (Nothing))
+                &(flatten (the (Maybe (Maybe Int)) (Just (Nothing))))
+                "flatten works on Just Nothing")
+
+  (assert-equal test
+                10
+                (expect (Just 10) "should not fail")
+                "expect works on Just")
+
+  (let-do [x 0]
```
hellerve: Let bindings get captured by value, so that won’t work. You’ll have to use a `def`.

hellerve (commented):
This change should turn CI green (used across all call sites)

hellerve (approved):

Issue #1526, opened 2026-04-08
Cabal file version and mentioned in README is still 0.5.5


hellerve:
We released 0.6.0, but the REPL and README still say 0.5.5.

Cheers

M-Tesla:
Hey @hellerve! I've opened a PR with the version fixes and also updated the actions in release.yml to v4 so the CI can build the binaries again. Let me know if everything looks good! :)

eriksvedang:
My bad, thanks for fixing.

PR #1548, opened 2026-05-01
chore: update version to 0.6.0 and fix release CI

M-Tesla:
This PR addresses issue #1526 by updating the version strings from `0.5.5` to `0.6.0` in the README, Cabal file, and REPL.

Additionally, I noticed the v0.6.0 release was missing the pre-built binaries in the Assets. After checking the Actions logs, the `Create Release` workflow failed because it was using a deprecated version of the artifact upload action (`actions/upload-artifact@v2`). 

I've bumped all actions to `v4` and fixed the deprecated `set-output` syntax in `release.yml` so the CI can successfully upload the binaries again. Let me know if any other changes are needed!

### DIFF
--- a/CarpHask.cabal
+++ b/CarpHask.cabal
@@ -1,5 +1,5 @@
 name:                CarpHask
-version:             0.5.5.0
+version:             0.6.0.0
 -- synopsis:
 -- description:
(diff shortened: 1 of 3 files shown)


hellerve (approved):
Thank you!

PR #1549, opened 2026-05-06
fix(emit): emit lambda env as void* to match call-site cast

hellerve:
This PR avoids UB by making `_env` a `void*` pointer.

Cheers

### DIFF
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -12,8 +12,8 @@ module Emit
   )
 where
 
+import Control.Monad (unless, when, zipWithM_)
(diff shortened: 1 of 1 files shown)


eriksvedang:
Fix looks nice.

@hellerve How do I produce the bug?

hellerve:
> Fix looks nice.
> 
> @hellerve How do I produce the bug?

It only happens on Linux with UBsan on, but then it complains about the pointer shape on closures: https://github.com/carpentry-org/persistent/actions/runs/25436609273/job/74616167898

PR #1550, opened 2026-05-09
fix: ensure closure move semantics are clean and route through the captured vars

hellerve:
**Disclaimer:** Claude helped me diagnose this issue and write a reproducible.

This PR fixes an issue where complex closure use cases (passed in structs, handed around and deleted much later) actually leaked.

Reproducible:

```clojure
(defn wrap [f]
  (fn [x] (f x)))

(defn main []
  (let-do [acc 0]
    (for [i 0 10000000]
      (let [s (Int.str i)
            inner (fn [n] (Int.+ (String.length &s) n))
            outer (wrap inner)]
        (set! acc (Int.+ acc (outer i)))))
    (IO.println &(fmt "done acc=%d" acc))))
```

Running with `time -l` (`time -v` on Linux), I see peak RSS on 10m iterations: 162.6 MB without the fix, 1.4 MB with the fix.

Cheers

### DIFF
--- a/src/Concretize.hs
+++ b/src/Concretize.hs
@@ -26,8 +26,8 @@ where
 import AssignTypes
 import Constraints
 import Control.Applicative
-import Control.Monad.State
 import Control.Monad (foldM, unless)
+import Control.Monad.State
 import Data.Either (fromRight)
 import Data.List (foldl')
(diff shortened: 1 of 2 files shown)


eriksvedang:
Oh damn, sneaky error!

PR #1552, opened 2026-05-11
Extend definterface with grouped protocol support

sqrew:
This PR extends definterface to support grouped, protocol-like behavior while preserving existing legacy interface semantics.

It introduces multiple implementation styles (atomic type-based, function-based bindings, and module-qualified forms) under a unified interface system, without introducing a separate defprotocol abstraction.

## Changes
- definterface supports multi-member, protocol-like interfaces
- Types can implement interfaces either as a whole or via explicit function bindings
- Legacy implements behavior is preserved
- Generic and polymorphic usage is supported
- Expanded test coverage for positive, negative, and edge-case behavior

## Scope
This is not a finalized or fully formalized protocol system. It does not introduce defprotocol and does not define exhaustive dispatch or resolution rules.

Some semantics (notably implementation resolution and retroactive behavior details) are intentionally left open for future iteration.

## Intent
Extend definterface into a flexible foundation for protocol-style abstraction while preserving backward compatibility and allowing incremental refinement based on real usage.


sqrew:
I forgot to include the disclaimer that I did use LLM assistance in this PR so please review accordingly. Also sorry about the 30 commits (lol) good luck!!

src/Interfaces.hs
```diff
@@ -140,8 +165,54 @@ registerInInterface ctx implementation interface =
...
 
+-- | Verify that a type fully implements all members of a protocol.
+verifyProtocolImplementation :: Context -> Ty -> XObj -> Either [InterfaceError] (Context, [(SymPath, SymPath)])
+verifyProtocolImplementation ctx ty _protocol@(XObj (Lst [XObj (Protocol memberPaths _) _ _, _]) _ _) =
+  let typeEnv = contextTypeEnv ctx
+      checkMember (errors, paths) mPath@(SymPath _ mName) =
+        case Env.getBinder typeEnv mName of
+          Right (Binder _ (XObj (Lst (XObj (Interface (sig : _) iPaths) _ _ : _)) _ _)) ->
+            let dummy = XObj (Lst []) Nothing Nothing
+                specializedSig = case solve [Constraint sig ty dummy dummy dummy OrdInterfaceImpl] of
```
hellerve: `sig` is the member's function type (e.g. `(Fn [a a] a)`) and `ty` is the implementing type (e.g. `BadNum`). These can never unify (Fn vs Struct).

src/Interfaces.hs
```diff
@@ -140,8 +165,54 @@ registerInInterface ctx implementation interface =
...
+                  Right mappings -> replaceTyVars mappings sig
+                  Left _ -> sig -- If unification fails, specializedSig remains generic and getFirstSatisfyingImplementation will fail.
+             in case getFirstSatisfyingImplementation ctx iPaths specializedSig of
```
hellerve, line 206: `getFirstSatisfyingImplementation` with the fully generic `sig` will succeed against any registered implementation, because the impl will unify with the typevar’d version (contrary to what the comment above says).

`(implements BadNum NumLike)` passes verification even though add is `(Fn [a a] Int)` instead of (`Fn [a a] a)`.

I believe the fix here is that to specialize correctly you need to substitute `a := ty` directly into the codomain, and not call the unifier on the whole signature.

src/Interfaces.hs
```diff
@@ -140,8 +165,54 @@ registerInInterface ctx implementation interface =
...
+      (allErrors, foundPaths) = foldl' checkMember ([], []) memberPaths
+   in if null allErrors
+        then Right (ctx, foundPaths)
+        else Left allErrors
+verifyProtocolImplementation _ _ protocol = Left [NonInterface (getPath protocol)]
+
+-- | Enforce the orphan instance rule for protocols.
+-- TODO: Protocols should only be implemented in the same module as the type or the protocol.
+checkOrphanRule :: Context -> SymPath -> Ty -> Either InterfaceError ()
+checkOrphanRule _ctx _protocol _ty = Right ()
```
hellerve, line 224: Is this part of this PR?

src/Obj.hs
```diff
@@ -171,7 +171,8 @@ data Obj
   | The
   | Ref
   | Deref
-  | Interface Ty [SymPath]
+  | Interface [Ty] [SymPath]
```
hellerve, line 174: Why did we widen this? We still only have one `Ty` interfaces. The others are protocols.

src/Polymorphism.hs
```diff
@@ -61,6 +61,10 @@ allImplementations typeEnv env functionName functionType =
           -- getPoly might return some functions we already found. Use set ops
           -- to remove duplicates.
           found -> (unionBy (\x y -> (snd x) == (snd y)) found getPoly)
+      -- Fallback to normal multi-sym style lookup if possible,
+      -- as protocol members are also registered as individual interfaces.
+      Right (Binder _ (XObj (Lst (XObj (Protocol _ _instances) _ _ : _)) _ _)) ->
```
hellerve, line 66: This does the same as the `_` catchall. We don’t need it.

src/Concretize.hs
```diff
@@ -458,6 +446,9 @@ visitInterfaceSym visited allowAmbig tenv env xobj@(InterfaceSymPat name) =
               [] -> pure (Right xobj) -- No exact match of types
               [y] -> updateSym y
               ps -> pure (Left (SeveralExactMatches xobj name actualType (map (\(t, _, p) -> (t, p)) ps)))
+    go (Binder _ (ListPat (ProtocolPat _ _instances))) =
+      -- Fallback to normal multi-sym style lookup if possible.
+      pure (Left (CannotConcretize xobj))
```
hellerve, line 451: This does the same as the `_` fallback. We don’t need it.

src/Types.hs
```diff
@@ -329,6 +335,8 @@ replaceTyVars mappings t =
         _ -> StructTy (replaceTyVars mappings name) (fmap (replaceTyVars mappings) tyArgs)
     (PointerTy x) -> PointerTy (replaceTyVars mappings x)
     (RefTy x lt) -> RefTy (replaceTyVars mappings x) (replaceTyVars mappings lt)
+    (ProtocolTy n is) -> ProtocolTy n (map (replaceTyVars mappings) is)
+    Universe -> Universe
```
hellerve, line 340: That seems fine, but did we need it as part of this PR?

hellerve (changes requested):
Good direction. I’ve found a few issues on soundness and simple code fixes, but there are some bigger, general comments:

1. I’d prefer if we write the documentation by hand. The additions to the docs are repetitive and don’t read well, it changes unrelated parts of the docs, and it has some actual correctness issues (like mentioning in the overview that derive can derive protocols, then clarifying that it cannot in the body of the derive doc).
2. There are quite a few unrelated changes in this PR (I stopped commenting on them at some point). To make this PR easier to review, let’s stick to the meat of it.

And more generally I think overloading definterface muddies the model.

The same form now produces two different binder kinds depending on whether the second argument is a type or an array. The protocol form also auto-creates individual interfaces for every member, so you can never have a protocol without polluting the namespace with member interfaces. This means that if two protocols with the same member name (like `P1` and `P2` in the test suite) silently share the same interface binder (`f` in the test case) with incompatible signatures the second definterface would either error or overwrite. In the test it works because `P1` and `P2` are compatible.

A separate `defprotocol` form would have been cleaner IMO.

sqrew:
@hellerve thanks for the review and the detailed comments — they’re helpful and fair.

I agree that a dedicated defprotocol form would likely be a cleaner long-term model in several respects. The choice to extend definterface instead was influenced partly by the discussion in #1292, which seemed to lean toward evolving the existing interface system rather than introducing a parallel abstraction immediately.

There are definitely tradeoffs with the current approach: overloading definterface increases cognitive overhead, protocol members currently share the interface namespace, and the implementation introduces additional compiler complexity internally. I definitely don’t disagree with those concerns.

That said, my primary goal in this work was to preserve backwards compatibility, extend Carp’s expressive capabilities incrementally, and establish a foundation that can evolve further over time.

My current implementation is intentionally conservative in that sense. I tried to reuse existing interface machinery rather than attempting to introduce an entirely separate dispatch model. While that leads to some awkwardness architecturally, it also keeps the feature relatively approachable to iterate on within the current state of the language.

Overall, I still believe grouped behavioral contracts are the correct direction for Carp and a meaningful capability increase for the language. Whether the final shape ends up being an extended definterface model or a dedicated defprotocol form is probably something that will become clearer through iteration and discussion.

PR #1553, opened 2026-05-11
perf: make parser a bit faster

hellerve:
This PR makes the parser around ~2x faster by dispatching without having to try full parser every time.

Cheers


eriksvedang (approved):
Neat. That makes a lot of sense!

PR #1555, opened 2026-05-12
Introduce minimal protocol system for structured generic programming

sqrew:
## Overview

This PR introduces a minimal  /  system for Carp, providing a structured abstraction for generic programming on top of the existing evaluation and metadata infrastructure.
It enables coherent grouping of related operations under a single abstraction and supports type-directed dispatch across multiple methods.

## Motivation
The existing interface system (, , ) is well-suited for single-function bindings and low-level abstraction.
However, it does not naturally model multi-method polymorphism where:
- multiple operations belong to a single abstraction (e.g. , )
- implementations must be registered per type and per protocol
- completeness and consistency across a set of methods is required
Attempting to extend the interface system to support these requirements leads to structural overlap between:
- single-function binding semantics
- grouped multi-method dispatch semantics

## Design Decision
Rather than extending , this PR introduces a new abstraction layer:
### defprotocol
Defines a named collection of method signatures as a single unit.
### impl-for
Registers implementations for a given type and protocol, evaluating and binding all required methods in a single structured step.

### 1. Separation of concerns
Interfaces model function-level binding. Protocols model grouped behavioral contracts across types.
### 2. Avoiding abstraction layering
Extending interfaces into multi-method dispatch effectively re-implements protocol behavior while retaining interface semantics underneath, resulting in overlapping systems.
### 3. Alignment with existing evaluator design
Carp’s evaluation and metadata system already supports constructing higher-level abstractions without modifying core primitives. A thin protocol layer composes cleanly with this model.

## Properties
- minimal surface area
- no changes to core evaluation engine
- built entirely on existing binder + metadata mechanisms
- clean separation between primitive binding and structured polymorphism
- extensible foundation for future generic programming features

## Relationship to existing features
This does not replace existing interface primitives.
- definterface / implements / impl remain available as low-level mechanisms
- defprotocol / impl-for are the primary user-facing abstraction for generic programming

## Result
This introduces a minimal but expressive foundation for:
- structured polymorphism
- type-directed dispatch
- multi-method abstraction units
- future generic programming features without additional core changes

### DIFF
--- a/test/protocol.carp
+++ b/test/protocol.carp
@@ -4,13 +4,13 @@
 ;; Basic protocol definitions
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(defprotocol MyShow
+(defprotocol (MyShow a) 
   [(show (Fn [(Ref a)] String))])
 
-(defprotocol MyEq
+(defprotocol (MyEq a) 
   [(eq (Fn [(Ref a) (Ref a)] Bool))])
 
-(defprotocol MyOrd
+(defprotocol (MyOrd a) 
   [(lt (Fn [(Ref a) (Ref a)] Bool))
    (gt (Fn [(Ref a) (Ref a)] Bool))])
 
--- a/test/protocol.carp
+++ b/test/protocol.carp
@@ -10,6 +10,10 @@
 (defprotocol MyEq
   [(eq (Fn [(Ref a) (Ref a)] Bool))])
 
+(defprotocol MyOrd
+  [(lt (Fn [(Ref a) (Ref a)] Bool))
+   (gt (Fn [(Ref a) (Ref a)] Bool))])
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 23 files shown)


test/protocol.carp
```diff
@@ -0,0 +1,256 @@
...
+                                "Generic MyBox gt works (via lt interdependence)")]
+        (assert-true &state
+                     (gt &b2 &b1)
+                     "Generic MyBox gt works for greater value"))))
+
+  ;; MyPair Show tests
+  (let [pair (MyPair.init 1 @"hi")]
+    (assert-equal protocol-test
+                  &@"(Pair 1 hi)"
+                  &(MyPair.show &pair)
```
hellerve: we should use the generic `show` here to test it, right?

test/protocol.carp
```diff
@@ -0,0 +1,256 @@
...
+                  "MyPair MyShow works for (MyPair Int String)"))
+
+  ;; Nested MyBox Ord tests
+  (let [nested1 (MyBox.init (MyBox.init 10))
+        nested2 (MyBox.init (MyBox.init 20))]
+    (assert-true protocol-test
+                 (MyBox.lt &nested1 &nested2)
```
hellerve: same as above. we should use the generic `lt` here to test it, right?

hellerve (commented):
i’ve only been able to give this a cursory glance so far. i think it’s the right direction, but i’m not a big fan of the `impl-for` name (maybe just `impl` or `defimpl`?), and i haven’t checked the invariants of the actual implementation yet.

PR #1558, opened 2026-05-31
fix: complete the Double math suite with missing C wrappers and registrations

sqrew:
This PR addresses several dangling definitions in the Double module where functions were registered in the Lisp layer but missing their corresponding C wrappers in carp_double.h.

Changes:
- Added Double_tan and Double_round to carp_double.h.
- Added C wrappers for all to-type conversion functions in carp_double.h.
- Registered Double.round in Double.carp (matching the Float module).
- Implemented standard interface registrations (to-int, to-float, to-long, to-uint64, to-bytes) in Double.carp.

### DIFF
--- a/core/Double.carp
+++ b/core/Double.carp
@@ -103,6 +103,8 @@
   (register tan (Fn [Double] Double))
   (doc tanh "returns the hyperbolic tangent of `a`.")
   (register tanh (Fn [Double] Double))
+  (doc round "returns the nearest integer to `a`.")
+  (register round (Fn [Double] Int))
 
   (implements abs Double.abs)
   (implements acos Double.acos)
@@ -126,6 +128,7 @@
   (implements sqrt Double.sqrt)
   (implements tan Double.tan)
   (implements tanh Double.tanh)
+  (implements to-int Double.to-int)
   (implements sign Double.sign)
 
   (doc sign "Returns `-1.0` if `x` is negative, `0.0` if zero, or `1.0` if positive.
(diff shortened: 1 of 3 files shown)


core/Double.carp
```diff
@@ -126,6 +128,7 @@
   (implements sqrt Double.sqrt)
   (implements tan Double.tan)
   (implements tanh Double.tanh)
+  (implements to-int Double.to-int)
```
hellerve, line 131: Already exists on line 59.

core/carp_double.h
```diff
@@ -1,4 +1,13 @@
+#ifndef CARP_DOUBLE_H
+#define CARP_DOUBLE_H
+
+#include <float.h>
+#include <math.h>
+#include <stdbool.h>
+#include <string.h>
+
 const double CARP_DBL_MAX = DBL_MAX;
+const double CARP_DBL_MIN = DBL_MIN;
```
hellerve, line 10: This is not `register`ed, so currently dead code?

core/carp_double.h
```diff
@@ -1,4 +1,13 @@
+#ifndef CARP_DOUBLE_H
+#define CARP_DOUBLE_H
```
hellerve, line 2: We don’t have this guard anywhere else.

core/carp_double.h
```diff
@@ -1,4 +1,13 @@
...
+
+#include <float.h>
+#include <math.h>
+#include <stdbool.h>
+#include <string.h>
```
hellerve, line 7: We rely on include order, so this is odd.

hellerve (changes requested):
A few more things:
- Add tests for `Double.round`.
- Reduce churn in `carp_double.h`. Your description says that functions were previously not defined in C, but this is not true. You just reordered them all over.

PR #1567, opened 2026-06-13
fix(eval): don't leak the implicit 'use Dynamic' into the global env

hellerve:
Ths PR fixes an issue where `evalDynamicVM`, `evalExpandVM` and `annotateWithinContextVM` did an implicit `(use Dynamic)` on the global env so unqualified dynamic calls resolve during evaluation, but the modified env escaped through the returned context and stuck around permanently. Static symbol qualification then saw the `Dynamic` module used, and it just got super messy.

The implicit use path is now scrubbed from result contexts, and qualification in `annotateWithinContextVM` runs on the regular context.

Cheers

### DIFF
--- a/src/Env.hs
+++ b/src/Env.hs
@@ -40,6 +40,7 @@ module Env
     deleteBinding,
     addListOfBindings,
     addUsePath,
+    removeUsePath,
     -------------------------
     -- finds
     findPoly,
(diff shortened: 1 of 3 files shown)


eriksvedang (approved):
Pesky bug!

PR #1570, opened 2026-06-18
Let types mark as recursive heap-indirection via a flag (generalizes Box)

hellerve:
This PR is both a bit of a RFC and an implementation. It generalizes `Box`’s way of allowing recursive definitions with a user-defined flag. It’s up to the user to prove that this is safe.

I want this to make `Rc` a bit more ergonomic. Right now, `Rc` needs to be a C type with a `void*` pointer inside, and a macro that monomorphises. If I could use the normal generic type machinery, it would be much more ergonomic, I just can’t prove it to the compiler to be correct.

The basics of how `Rc` works is:

```clojure
(deftype (Rc a) [cell (Ptr (RcCell a))])
(recursive Rc)
```

which is like a generic pointer, but I want it type-safe! Previously, I had to monomorphise by generating special cases through macros. By marking it as recursive, I can let the type layer handle it

To do this we:
- need to be able to mark a type as recursive, and then use that as we did with `isBoxName` before,
- fix emit ordering (it gets a shallow score) such that it comes right behind foward declarations,
- add proper deletors for that type.

So far so easy. Why is this so much code then? Well, it turns out a generic type looping through another generic type (like, say `(Rc (Lst a))` would make the com piler loop forever during concretization, so we need a set of paths that we already visited that we keep track of. Threading that through everything is the rest of the diff. It’s plumbing, and it sucks that it touches so much, but I couldn’t come up with a better solution.

Cheers


eriksvedang:
Has merge conflicts now.

eriksvedang (commented):
I think the PR should include an example or test that excercises this new feature.

eriksvedang (approved):
Cool feature! Should the built-in box switch to using this instead?

hellerve:
Good idea. I can work on this, but would it make sense as a follow-up PR?

hellerve:
Okay, as always, I couldn’t help it.

```clojure
(deftype (Box a) [ptr (Ptr a)])
(defmodule Box
    (deftemplate halloc (Fn [a] (Ptr a)) "$a* $NAME($a v)"
      "$DECL { $a* p = CARP_MALLOC(sizeof($a)); *p = v; return p; }")

    (deftemplate hfree (Fn [(Ptr a)] ()) "void $NAME($a* p)" "$DECL { CARP_FREE(p); }")

    (defn box [v] (Box.init (Box.halloc v)))
    (defn peek [b] (Pointer.to-ref @(Box.ptr b)))
    (defn unbox [b]
      (let-do [p @(Box.ptr &b)
               v (Pointer.to-value p)]
      (Box.hfree p)
      (Unsafe.leak b) v))

    (defn copy [b] (Box.box @(Box.peek b)))
    (implements copy Box.copy)

    (defn delete [b]
      (let-do [p @(Box.ptr &b)]
        (ignore (Pointer.to-value p))
        (Box.hfree p)))
    (implements delete Box.delete))
(recursive Box)
```
  
There’s no real way to overload `init` in a good way, though, so that one would have to stay special. That’s because the built-in `Box.init` has the sig `(Fn [a] (Box a))`, but regularly it would have the sig `(Fn [(Ptr a)] (Box a))`.

eriksvedang:
Lets ship this as-is then.

PR #1572, opened 2026-07-20
Fix duplicate deleters in manage function

iwami4438:
- Filter out existing deleters with the same name before insertion.
- Added a comment suggesting exporting deleterVar in Info.hs later.
- ormolu fixed the style.
- Ref: https://github.com/carp-lang/Carp/issues/1457


src/Memory.hs
```diff
@@ -519,7 +519,15 @@ manage typeEnv globalEnv xobj =
...
       Just deleter -> do
         m <- get
-        let newDeleters = Set.insert deleter (memStateDeleters m)
+        let var = varOfXObj xobj
+            deleterV = \case
+              -- It's probably better to export deleterVar in Info.hs's.
+              ProperDeleter _ _ v -> v
+              RefDeleter v -> v
+              PrimDeleter v -> v
+              FakeDeleter v -> v
```
hellerve: I’d agree to instead export `deleterVar` and use that.

hellerve (commented):
Thanks for looking into this!

I think this is an alright fix (I would remove the duplicated check here), but I think we could go for something more principled, and remove the root cause of the issue.

As far as I remember, the main issue was that the parser doesn’t assign identifiers (they are set in `setNewIdentifiers` in expansion). This is fine for all regular forms, but primitives never get there, so `kind` and `type` call qualification and annotation on uninitialized identifiers.

What I propose we do instead is to renumber before annotating in the two primitives that do this, mirroring what every other caller to `annotate` does. This is still brittle, because it also needs any future caller to do this, but I think it’s nicer than filtering it out later.

I’ll happily let others chime in, though.

iwami4438:
Thank you for the suggestion. I agree that addressing the root cause is better.

​I will update the implementation to renumber before annotating in the primitives as you proposed.

hellerve (approved):
Looks good to me now! If we want to go the extra mile, we can add some regression tests, but other than that, this is clean!

iwami4438:
Okay, I'll add the test code.

hellerve (approved):

Issue #1542, opened 2026-04-20
Reassigning unit values emits broken C


hellerve:
We emit broken C when we re-assign a unit variable (should be a no-op):

```
鲤 (defn main [] (let [x ()] (set! x ())))
鲤 :bx
out/main.c:11335:9: error: use of undeclared identifier 'x'
 11335 |         x = ;  // () = ()
       |         ^
out/main.c:11335:13: error: expected expression
 11335 |         x = ;  // () = ()
       |             ^
2 errors generated.
carp: callCommand: clang  -o out/Untitled -I /Users/veitheller/Documents/Code/Github/carp/Carp//core/  -fPIC -g -std=c99 -D_DEFAULT_SOURCE -Wall -Werror -Wno-unknown-warning-option -Wno-unused-variable -Wno-unused-but-set-variable -Wno-self-assign -lm out/main.c (exit 1): failed
```

Cheers

PR #1573, opened 2026-07-31
fix: broken C emitted when re-assigning a unit-typed variable

iwami4438:
- `set!` always emitted an assignment, producing invalid C (`x = ;`) for unit-typed (()) variables, which have no C declaration
- Now skipped as a no-op, matching declaration behavior (Emit.hs 624~)
- Added a regression test
- Ref: https://github.com/carp-lang/Carp/issues/1542


hellerve (approved):
looks good, thanks! in the future it might be better to separate the reformat from the fix. i had to dig for a while to understand the magic is in `src/Emit.hs:625`.

iwami4438:
Thank you for the review, and sorry for the extra digging it caused.
From now on, I'll use `git commit --no-verify` to keep the reformat separate from the logic change.

hellerve:
i wonder why my commit hooks don’t seem to fire the same way. we should revisit the ormolu setup in general i guess! i’ll open an issue.

iwami4438:
I'm using the version I installed this way. `stack install ormolu`
However, it does feel a little intrusive.
For reference, here's the version I'm using:
```sh
ormolu -v
ormolu 0.7.2.0
using ghc-lib-parser 9.6.7.20250325

stack --version
Version 3.11.1, Git revision 2352d78a8ac5b42d021c8064b8f64ac1c8b8b3d5 x86_64 hpack-0.39.6
```

PR #1575, opened 2026-08-05
Fix format leaks memory

### LINKED ISSUE #1432: `format` leaks memory
```
> (format "%d%d" 1)
1-1900850528
> (format "%d %d" 1)
1 279724704
```

iwami4438:
- Changed to use the internal function `format*` instead of `format` (name change only).
- Updated calls to the existing `format` interface to use `format*` (name change only).
- Configured the code to call `format*` when there is a single argument, and `fmt` otherwise.
- Fixes #1432


hellerve (changes requested):
That does not fix the issue, it just shifts it from `format` to `format*`. See, for instance `(format* "%d%d" 1)`. `format` is intentionally the single argument version, if we want to make it safer, we should make all implementors check the format string.

iwami4438:
Thank you for the review. I overlooked the fact that `format*` could be called directly, bypassing the check, sorry.

I'll revert this and intercept each type's implementation instead, e.g.:

```clojure
(defmodule Int
  (private format*)
  (hidden format*)
  (register format* (Fn [&String Int] String) "Int_format")

  (defn format [s x]
    (let [n (count-specifiers s)]
      (if (= n 1)
        (format* s x)
        (let [msg (String.append "error in format string: expected exactly 1 specifier, found "
                                   &(Int.str n))]
          (do
            (IO.errorln &msg)
            msg)))))
  (implements format Int.format))
```

Small question: I intend to define a helper function called `count-specifiers`. `count-specifiers` needs to be called from every type module, so it ends up effectively global (can't be `private`). Is that okay? My concern is simply that `count-specifiers` could be called from anywhere, such as the REPL.

hellerve:
So the issue with this is that it’s a runtime check (which we try to avoid, which is why we do it at compile time in `fmt`). Maybe we can try to do it in a macro there, too?

I can sketch out what I mean in case it’s unclear.

iwami4438:
I don't yet have a concrete idea of ​​the implementation, so it would be helpful if you could provide an overview (or a sketch).

My concerns are as follows:

- Since `format` is defined via `definterface`, the appropriate implementation is dispatched based on argument types. Because macros are expanded prior to type checking, I cannot envision a mechanism where the macro itself handles this dispatch logic.
- I tried simply shadowing (overwriting) `format` with a macro that calls itself:
```clojure
(defmacro format [s x] (format s x))
(format "%d" 1) ;; Infinite loop
```
Once `format` is redefined as a macro, it seems there would be no way to access the original interface using that name (or perhaps it simply keeps overwriting itself?).

Specifically, the behavior of a `defmacro` inside a `defmodule` (e.g., `(defmodule Int (defmacro format ...))`) is unclear. Would it participate in the interface dispatch mechanism in some way, or would it simply become a namespaced macro accessible only as `Int.format`? In the latter case, I do not see how the compiler could determine which module's macro to use before the argument types are resolved.

hellerve:
i just played around with it and have a working patch. A hint:

```clojure
(hidden format-directives)
(doc format-directives "counts the format directives in `s`, ignoring escaped `%%`.")
(defndynamic format-directives [s]
   ; do cool stuff
)

(doc format-unchecked "formats `x` into `s` without checking the format string. Prefer `format`, which checks it at compile time.")
(defn format-unchecked [s x] (format s x))

(doc format "formats `x` into the format string `s`. `s` must contain exactly one format directive, and is required to be a string literal. Use `fmt` to format more than one value.")
(defmacro format [s x]
  (let [n (format-directives s)]
    (if (= n 1)
     `(format-unchecked %s %x)
      (macro-error (str "error in format string: expected exactly one directive, but found " n
          " in \"" s "\". use fmt to format more than one value")))))
```

my version also required load order to be changed such that the format file is loaded after `String.carp`.

iwami4438:
Does anyone know how to catch `macro-error`? It seems there is no such method; is it safe to assume that not catching `macro-error` is acceptable in the current situation?

eriksvedang:
> Does anyone know how to catch `macro-error`? It seems there is no such method; is it safe to assume that not catching `macro-error` is acceptable in the current situation?

Yeah it's not possible to catch. I think that's acceptable.

iwami4438:
OK. 
This time, we'll remove the test that catches `macro-error`.

iwami4438:
I removed the test that checked for `macro-error`, which was causing the CI to fail. Please review.

core/FormatChecked.carp
```diff
@@ -0,0 +1,29 @@
...
+    (cond
+      (= idx -1) 0 ;; no %
+      (= idx (dec len)) -1 ;; found % at the end -> error
+      (= \% (String.char-at s (inc idx))) ;; next char = % -> this is an escaped %
+        (format-directives (String.suffix s (+ idx 2))) ;; other %?
+      (let [rest (format-directives (String.suffix s (inc idx)))]
+        (cond
+          (= rest -1) -1 ;; ex: "a%%%" -> error
+          (= rest 0) 1   ;; found %    -> ok
+          2)))))         ;; too many % -> error
```
hellerve: as it stands right now, this doesn’t actually count the directives. it will return `2` for anything over 1 directive.

iwami4438: I'm sorry, that was a mistake.

iwami4438:
I've made the corrections, so please review it.

core/FormatChecked.carp
```diff
@@ -0,0 +1,30 @@
...
+      (= idx (dec len)) -1 ;; found % at the end -> error
+      (= \% (String.char-at s (inc idx))) ;; next char = % -> this is an escaped %
+        (format-directives (String.suffix s (+ idx 2))) ;; other %?
+      (let [rest (format-directives (String.suffix s (inc idx)))]
+        (if (= rest -1)
+          -1              ;; too many % -> error
+          (+ rest 1)))))) ;; count %
+
+(doc format-unchecked "formats `x` into `s` without checking the format string. Prefer `format`, which checks it at compile time.")
+(defn format-unchecked [s x] (format s x))
```
eriksvedang, line 17: I assume this is referring to `format` in the `Format` module? The code would be much more readable if such usages were prefixed with the module name (`Format.format`).

iwami4438: That’s correct. I originally had this code in `core/Format.carp`, but I moved it to a separate file because the load order was causing some errors. Should I make that change?

eriksvedang: Having it in a separate file sounds good. But if you can use qualified names, that'd be much appreciated.

iwami4438: OK. I will change the notation to `Format.format`.

iwami4438: Sorry to reply again. Upon re-checking, I found that `format` is not actually included in the `Format` module; due to file splitting, it exists only in `core/Format.carp`. Therefore, I believe no specific changes are necessary—what do you think?

eriksvedang: Ah, right it's an interface. Which makes me a bit concerned about redefining it as a macro like you do here... that seems like it should break things. Or at least make it hard to add more implementations later (in user code). Have you experimented with reloading everything in the repl (:r) and adding new format implementations?

iwami4438: I've only been looking at the happy path—sorry about that. I'll try using `:r` or implementing it.

iwami4438: I gave it a try. There don't seem to be any particular issues.
I'll just paste the REPL output here.
```clojure
> (deftype Point [x Int, y Int])
> (defmodule Point
(  (defn format [s p]
((   (IntRef.format s (Point.x &p)))
(  (implements format Point.format))
> (format "%d" (Point.init 1 2))
1
=> ()
> :r
> (format "%d" (Point.init 1 2))
1
=> ()
```

iwami4438: I forgot to redefine it. That looks fine as well. (This is a continuation of the REPL session from earlier.)
```clj
> :r
> (defmodule Point
((defn format [s p]
(((IntRef.format s (Point.x &p)))
((implements format Point.format))
> (format "%d" (Point.init 1 2))
1
=> ()
```

eriksvedang: Thanks. I haven't had time to try out the branch myself, but I'll get to it.

hellerve:
i’ve been thinking about this, and the crux is that this is still not fixing the issue at its core, because i can still call `Int.format` directly and run into undefined behavior, which is not cool. i think we should admit defeat here, and do the following:

1. rename the interface to `unsafe-format`. it’s unsafe, that’s fine, everyone should call compile-time-checked `fmt` anyway when possible.
2. harden the c primitives with a `NDEBUG`-guarded runtime check. i was against runtime checks before, because i thought we can solve this statically with macros. i was wrong, and i’m sorry. we could add a macro similar to `CHK_INDEX` for all `format` implementations in c, and tuck them into an `ifdef`.

iwami4438:
Okay.
I understand your proposal. However, due to personal circumstances, it would be difficult for me to handle this matter during September and October. Therefore, if there is someone else who would like to take this on, I would appreciate it if you could entrust the task to them.

hellerve:
Absolutely. Sorry about the back and forth. I’ll take it on in the coming days and build on your work to avoid it getting lost and you getting the credit where it’s due. Thanks for your contribution so far!

PR #1579, opened 2026-08-16
improve(core): upgrade trace macro and add custom assert messages

sqrew:
Improves Debug.trace to display the expression text along with its value (using prn formatting), adds custom error message support to global assert via varargs, and removes the redundant/buggy Dynamic.Debug.trace macro.

### DIFF
--- a/test/produces-output/trace_assert_check.carp
+++ b/test/produces-output/trace_assert_check.carp
@@ -1,0 +1,9 @@
+(Project.config "file-path-print-length" "short")
+(defn main []
+  (let [x (Debug.trace (+ 10 20))
+        y (Debug.trace @"hello")]
+    (do
+      (assert (= x 30))
+      (assert (= x 30) "x must be 30")
+      (IO.println "All test assertions passed successfully!")
+      ())))
--- a/test/trace_assert_check.carp
+++ b/test/trace_assert_check.carp
@@ -1,0 +1,11 @@
+;; This file verifies the improved 'trace' and 'assert' macros in core/Debug.carp
+
+(defn main []
+  (let [x (Debug.trace (+ 10 20))
+        y (Debug.trace @"hello")]
+    (do
+      (assert (= x 30))
+      (assert (= x 30) "x must be 30")
+      ;; Print success message
+      (IO.println "All test assertions passed successfully!")
+      ())))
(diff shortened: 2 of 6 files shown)


hellerve (approved):

test/trace_assert_check.carp
```diff
@@ -0,0 +1,11 @@
+;; This file verifies the improved 'trace' and 'assert' macros in core/Debug.carp
+
+(defn main []
+  (let [x (Debug.trace (+ 10 20))
```
eriksvedang, line 4: Is the output from these trace calls checked somewhere?

sqrew: No, you're probably right that it should be though. I didn't think of that!!

eriksvedang: Should be doable if you add it to the tests that check their output.

sqrew: will do

sqrew: Hey @eriksvedang! I looked into adding the output check to the produces-output/ suite, but I ran into a bit of a snag with how trace resolves paths.

  Because the macro uses (file) under the hood, it evaluates to the absolute path of the file at compile time. If I add it to produces-output, the .expected file would need my hardcoded local path (e.g., /home/.../Carp/test/...) and
  would instantly fail on GitHub Actions when the CI runner checks it out to a different directory.

  I could write a custom shell script to sanitize $PWD from the test output, but since the rest of the test suite cleanly uses execute.sh, I figured it's better to avoid polluting the test harnesses with one-off hacks. Because of this,
  the test currently just executes the macro to ensure it compiles and evaluates properly without crashing. Let me know if you'd still prefer me to add a path-sanitizing hack to the test scripts though!

eriksvedang: I see. Does setting `file-path-print-length` to `short` work in this context?

sqrew: that is exactly the correct way to have solved this. I didn't think of that at all. Nice catch

test/produces-output/trace_assert_check.carp
```diff
@@ -0,0 +1,9 @@
+(Project.config "file-path-print-length" "short")
+(defn main []
+  (let [x (Debug.trace (+ 10 20))
+        y (Debug.trace @"hello")]
+    (do
+      (assert (= x 30))
+      (assert (= x 30) "x must be 30")
+      (IO.println "All test assertions passed successfully!")
```
hellerve, line 8: this doesn’t currently test the output of assert. shouldn’t it be the other way around?

PR #1581, opened 2026-08-26
Add docexample and print-docexample to compiler and stdlib

sqrew:
Introduces 'docexample' metadata support for documenting code examples. Defines the 'docexample' macro in Macros.carp and updates the compiler's 'info' command to display it.


eriksvedang (commented):
Nice. The example should probably be printed on a new line, to not mess up the formatting too badly?

eriksvedang (approved):

PR #1582, opened 2026-08-27
Treat flagged recursive types as generic to prevent infinite instantiation recursion

sqrew:
### Summary
This PR prevents infinite recursion during type instantiation and template generation (e.g. `copy`, `delete`, and `str`/`prn` instances) for recursive sumtypes and deftypes.

### Problem
When a type includes a recursive reference to itself (such as a recursive `Box` type), the compiler's concrete template generator would recursively attempt to instantiate copies and deletes of the inner recursive member. This would result in an infinite loop during compilation, eventually crashing the compiler.

### Solution
In `src/Deftype.hs` and `src/Sumtypes.hs`, we update the instantiation functions (`binderForStrOrPrn`, `binderForDelete`, `binderForCopy`) to check if the type is recursive using the `flagged` helper:
- If a type is recursive/flagged, we treat it as a **generic type** during template instantiation.
- This instructs the compiler to generate a parameterized template instead of a concrete instance, successfully breaking the cycle and preventing the infinite copy/instantiation loops.

### DIFF
--- a/src/Deftype.hs
+++ b/src/Deftype.hs
@@ -239,7 +239,7 @@ binderForStrOrPrn strOrPrn candidate =
   let binderP = SymPath (TC.getFullPath candidate) strOrPrn
       binderT = (FuncTy [RefTy (TC.toType candidate) (VarTy "q")] StringTy StaticLifetimeTy)
       doc = "converts a `" ++ TC.getName candidate ++ "` to a string."
-   in if isTypeGeneric (TC.toType candidate)
+   in if isTypeGeneric (TC.toType candidate) || flagged (TC.getTypeEnv candidate) (TC.getValueEnv candidate) (TC.toType candidate)
         then Right $ (defineTypeParameterizedTemplate (TG.generateGenericTypeTemplate candidate strGenerator) binderP binderT doc, [])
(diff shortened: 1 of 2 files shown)


sqrew:
### Summary
This PR prevents infinite recursion during type instantiation and template generation (e.g. `copy`, `delete`, and `str`/`prn` instances) for recursive sumtypes and deftypes.

### Problem
When a type includes a recursive reference to itself (such as a recursive `Box` type), the compiler's concrete template generator would recursively attempt to instantiate copies and deletes of the inner recursive member. This would result in an infinite loop during compilation, eventually crashing the compiler.

### Solution
In `src/Deftype.hs` and `src/Sumtypes.hs`, we update the instantiation functions (`binderForStrOrPrn`, `binderForDelete`, `binderForCopy`) to check if the type is recursive using the `flagged` helper:
- If a type is recursive/flagged, we treat it as a **generic type** during template instantiation.
- This instructs the compiler to generate a parameterized template instead of a concrete instance, successfully breaking the cycle and preventing the infinite copy/instantiation loops.

hellerve:
I have trouble believing this. Do you have a concrete red/green example/regression test?

PR #1584, opened 2026-09-01
Add result-do macro for flat Result pipelines

sqrew:
This PR adds the `result-do` macro to flatten nested `Result` operations (similar to Haskell's `do` notation or Rust's `?` operator). Includes an arity check for the bindings array, and corresponding unit tests.

### DIFF
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -194,6 +194,33 @@ Example:
     (list 'let [name form]
       (case-internal name branches))))
 
+(hidden result-do-internal)
+(defndynamic result-do-internal [bindings body]
+  (if (= (length bindings) 0)
+    body
+    (let [name (car bindings)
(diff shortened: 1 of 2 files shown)


hellerve (changes requested):
i wouldn’t make the control macros depend on `Result`. It should go into the `Result` file and maybe even into its namespace.

sqrew:
i will make those changes

sqrew:
I've moved the macro into the `Result` module as `Result.do`. Thanks for the review!

core/ControlMacros.carp
```diff
@@ -194,6 +194,7 @@ Example:
     (list 'let [name form]
       (case-internal name branches))))
 
+
```
hellerve, line 197: Can we remove the whitespace churn?

PR #1585, opened 2026-09-01
Add maybe-do macro for flat Maybe pipelines

sqrew:
This PR adds the `maybe-do` macro to flatten nested `Maybe` operations. Includes an arity check for the bindings array, and corresponding unit tests.

### DIFF
--- a/core/ControlMacros.carp
+++ b/core/ControlMacros.carp
@@ -194,6 +194,32 @@ Example:
     (list 'let [name form]
       (case-internal name branches))))
 
+(hidden maybe-do-internal)
+(defndynamic maybe-do-internal [bindings body]
+  (if (= (length bindings) 0)
+    body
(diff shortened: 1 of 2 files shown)


hellerve (changes requested):
i wouldn’t make the control macros depend on `Maybe`. It should go into the `Maybe` file and maybe even into its namespace.

sqrew:
i will make those changes

sqrew:
I've moved the macro into the `Maybe` module as `Maybe.do`. Thanks for the review!

core/ControlMacros.carp
```diff
@@ -194,6 +194,8 @@ Example:
     (list 'let [name form]
       (case-internal name branches))))
 
+
+
```
hellerve, line 198: Can we remove the whitespace churn?

PR #1586, opened 2026-09-03
feat(core): Add source-location macro for compile-time AST location extraction

sqrew:
This PR introduces a `source-location` macro to `core/Dynamic.carp`. 

It evaluates the builtin `file` and `line` metadata against a given AST node, returning a formatted string (e.g. `"src/main.carp:42"`). 

This is highly useful for building custom runtime loggers, assertions, and profilers without having to rely entirely on the REPL `info` command.

### DIFF
--- a/core/Dynamic.carp
+++ b/core/Dynamic.carp
@@ -123,3 +123,7 @@ integers [`imod`](#imod).")
 
 (defndynamic add-c [relpath]
   (Project.config "cmod" (relative-to (current-file) relpath)))
+
+(doc source-location "Returns a string representing the file and line number of the given AST node. Useful for building custom logging and profiling macros.")
+(defmacro source-location [x]
+  (Dynamic.String.concat [(eval (list 'file x)) ":" (str (eval (list 'line x)))]))


hellerve (commented):
can you make an example invocation for this? i cannot seem to get this to work as expected.

sqrew:
@hellerve Good point! To demonstrate its utility, I've updated the PR to:
1. Include `column` in the `source-location` macro for even more precise tracking.
2. Refactored the core `trace` and `assert` macros in `core/Debug.carp` to use `source-location`. 

For example, `assert` can now be simplified to just rely on `source-location` instead of manually piecing together file, line, and column variables:
```clojure
(defmacro assert [expr]
  `(unless (= true %expr)
     (do
       (println* (fmt "Assertion '%s' failed at %s" %(str expr) %(eval `(source-location %expr))))
       (System.abort))))
```

I've also added a compile-time assertion test in `test/macros.carp`. Let me know what you think of this approach!

sqrew:
@hellerve Just a heads up: I realized that if you merge #1579 first (which touches the same `trace` and `assert` macros), this PR will get a merge conflict. Whichever one you merge first, just let me know and I'll happily rebase the other one to combine them cleanly!

core/Dynamic.carp
```diff
@@ -123,3 +123,7 @@ integers [`imod`](#imod).")
 
 (defndynamic add-c [relpath]
   (Project.config "cmod" (relative-to (current-file) relpath)))
+
+(doc source-location "Returns a string representing the file, line, and column number of the given AST node. Useful for building custom logging and profiling macros.")
+(defmacro source-location [x]
+  (Dynamic.String.concat [(eval (list 'file x)) ":" (str (eval (list 'line x))) ":" (str (eval (list 'column x)))]))
```
hellerve, line 129:
```suggestion
(defndynamic source-location-of [x]
  (Dynamic.String.concat [(eval (list 'file x)) ":" (str (eval (list 'line x))) ":" (str (eval (list 'column x)))]))

(doc source-location "Returns a string representing the file, line, and column number of the given AST node. Useful for building custom logging and profiling macros.")
(defmacro source-location [x] (source-location-of x))
```

hellerve (commented):
if we change the macro to a macro+dynamic, we can get rid of the `eval`.

PR #1591, opened 2026-09-14
feat(docs): render docexample metadata in HTML generator

sqrew:
This PR updates `src/RenderDocs.hs` to extract and render the `docexample` metadata into the HTML documentation, formatted as clojure code blocks. This completes the `docexample` feature from #1581 by ensuring it is visible in the HTML output, not just the REPL.


eriksvedang (commented):
Nice! Code looks good, but I think it would be good to clearly mark the snippet as an example in the html. Right now it could be interpreted as the source code, or anything really.

Issue #1576, opened 2026-08-10
we leak memory on `break` inside `if` or `match`


hellerve:
i encountered an ouch-y in the carpentry:
```clojure
(defn leaky []
    (while true
      (if false
        ()
        (do
          (IO.print &(fmt "%d\n" 1))
          (break)))))

(defn main [] (for [j 0 100] (leaky)))
```

run & observe: 

```
$ carp -b repro.carp && leaks -atExit -- ./out/Untitled | grep 'leaks for'
Process 63306: 200 leaks for 3200 total leaked bytes.
```

we’re forgetting to insert the right deleters in `if` or `match` cases. i’ve not yet investigated/triaged this further.

cheers

PR #1592, opened 2026-09-22
fix: delete branch-local values before break inside if/match

seaslugr:
Fixes issue #1576

if/match branches where scheduling their local delete calls to run *after* the branch body, but a break inside that branch jumps out before reaching them... so those deletes are dead and the memory leaks.

Fixed!


hellerve (changes requested):
thanks for your contribution! this is a great start, unfortunately it’s a bit too aggressive, because it inserts deleters for values that aren’t created yet.

minimal poc:

```clojure
(defn f []
  (while true
    (when true
      (do (break)
          (IO.println &(str 1))))))

(defn main [] (f))
```

compiles in carp, dies in c, because we emit something like:

```c
  if (true) {
      String_delete(_14); 
      break;
      String _14 = Int_str(1);
      // rest of program...
  }
```

